PR #101, opened 2023-05-26
fix: hardhat config to correctly pick up a non-ledger address

kupermind:
- Hardhat config to correctly pick up a non-ledger address.


hardhat.config.js
```diff
@@ -50,18 +53,13 @@ module.exports = {
     networks: {
         mainnet: {
             url: "https://eth-mainnet.g.alchemy.com/v2/" + ALCHEMY_API_KEY_MAINNET,
+            accounts,
```
kupermind, line 56: Need to provide accounts, otherwise the default ones are taken

mariapiamo (approved):

PR #103, opened 2023-07-06
fix: correctly calculate incentive values for the event

kupermind:
- Correctly calculate incentive values for the event when the next epoch is going to be a year changing epoch.


77ph (approved):

mariapiamo (approved):

kupermind:
Was refactored during staking implementation and is already in main.

PR #104, opened 2023-07-24
refactor: updating product behavior

kupermind:
- Bonding product now does not expire;
- Bonding product becomes inactive and gets deleted only when the supply becomes zero;
- If the product is closed, trying to close it by the contract owner does not revert.


contracts/Depository.sol
```diff
@@ -65,10 +65,11 @@ contract Depository is IErrorsTokenomics {
...
     event CreateBond(address indexed token, uint256 indexed productId, address indexed owner, uint256 bondId,
-        uint256 amountOLAS, uint256 tokenAmount, uint256 expiry);
-    event RedeemBond(uint256 indexed productId, address indexed owner, uint256 bondId);
-    event CreateProduct(address indexed token, uint256 indexed productId, uint256 supply, uint256 priceLP, uint256 expiry);
-    event CloseProduct(address indexed token, uint256 indexed productId);
+        uint256 amountOLAS, uint256 tokenAmount, uint256 maturity, uint256 startTime);
+    event RedeemBond(uint256 indexed productId, address indexed owner, uint256 bondId, uint256 redeemTime);
+    event CreateProduct(address indexed token, uint256 indexed productId, uint256 supply, uint256 priceLP,
+        uint256 vesting, uint256 startTime);
+    event CloseProduct(address indexed token, uint256 indexed productId, uint256 closeTime);
```
kupermind, line 72: Adding start and redeem / close time to events.

contracts/Depository.sol
```diff
@@ -208,10 +209,11 @@ contract Depository is IErrorsTokenomics {
...
 
-        // Check for the expiration time overflow
-        uint256 expiry = block.timestamp + vesting;
-        if (expiry > type(uint32).max) {
-            revert Overflow(expiry, type(uint32).max);
+        // Check for the maturity time overflow for the current timestamp considering that the product will be live
+        // for at least MIN_VESTING time
+        uint256 maturity = block.timestamp + MIN_VESTING + vesting;
+        if (maturity > type(uint32).max) {
+            revert Overflow(maturity, type(uint32).max);
```
kupermind, line 216: Check that the vesting does not overflow when converted to the maturity time once the bond is created.

contracts/Depository.sol
```diff
@@ -246,36 +249,37 @@ contract Depository is IErrorsTokenomics {
         for (uint256 i = 0; i < productIds.length; ++i) {
             uint256 productId = productIds[i];
             // Check if the product is still open
-            if (mapBondProducts[productId].expiry == 0) {
-                revert ProductClosed(productId);
+            if (mapBondProducts[productId].vesting == 0) {
+                continue;
```
kupermind, line 253: Do not revert if the product is already closed or never existed.

contracts/Depository.sol
```diff
@@ -246,36 +249,35 @@ contract Depository is IErrorsTokenomics {
         for (uint256 i = 0; i < productIds.length; ++i) {
             uint256 productId = productIds[i];
             // Check if the product is still open
-            if (mapBondProducts[productId].expiry == 0) {
-                revert ProductClosed(productId);
-            }
+            if (mapBondProducts[productId].vesting > 0) {
```
kupermind, line 252: Do not revert if the product is already closed or never existed.

contracts/Depository.sol
```diff
@@ -246,36 +249,35 @@ contract Depository is IErrorsTokenomics {
...
+                }
+                address token = mapBondProducts[productId].token;
+                delete mapBondProducts[productId];
 
-            uint256 supply = mapBondProducts[productId].supply;
-            // Refund unused OLAS supply from the program if it was not used by the product completely
-            if (supply > 0) {
-                ITokenomics(tokenomics).refundFromBondProgram(supply);
+                numClosedProducts++;
+                emit CloseProduct(token, productId, block.timestamp);
```
kupermind, line 262: Calculate the number of closed products.

contracts/Depository.sol
```diff
@@ -288,13 +290,19 @@ contract Depository is IErrorsTokenomics {
...
+        uint256 vesting = product.vesting;
+        if (vesting == 0) {
+            revert ProductExpired(token, productId, vesting, block.timestamp);
         }
 
+        maturity = vesting + block.timestamp;
+        // Check for the time limits
+        if (maturity > type(uint32).max) {
+            revert Overflow(maturity, type(uint32).max);
+        }
```
kupermind, line 304: Check for the maturity limits

contracts/Depository.sol
```diff
@@ -311,13 +319,19 @@ contract Depository is IErrorsTokenomics {
...
         ITreasury(treasury).depositTokenForOLAS(msg.sender, tokenAmount, token, payout);
 
-        emit CreateBond(token, productId, msg.sender, bondId, payout, tokenAmount, expiry);
+        // Close the product if the supply becomes zero
+        if (supply == 0) {
+            delete mapBondProducts[productId];
+            emit CloseProduct(token, productId, block.timestamp);
+        }
+
+        emit CreateBond(token, productId, msg.sender, bondId, payout, tokenAmount, maturity, block.timestamp);
```
kupermind, line 334: If the supply is completely taken out, close the product.

contracts/Depository.sol
```diff
@@ -344,58 +358,39 @@ contract Depository is IErrorsTokenomics {
...
-                    positions[i] = true;
-                    ++numSelectedProducts;
-                }
-            }
-        } else {
-            for (uint256 i = 0; i < numProducts; i++) {
-                if (mapBondProducts[i].token != address(0) && mapBondProducts[i].expiry <= block.timestamp) {
-                    positions[i] = true;
-                    ++numSelectedProducts;
-                }
```
kupermind, line 398: There are no inactive products anymore, as they are either active when the supply is not zero, or closed ones, and all the product data is cleared.

contracts/Depository.sol
```diff
@@ -208,10 +211,15 @@ contract Depository is IErrorsTokenomics {
...
-            revert Overflow(expiry, type(uint32).max);
+        // Check the vesting maximum limit value
+        if (vesting > MAX_VESTING) {
+            revert Overflow(vesting, MAX_VESTING);
+        }
+
+        // Check for the maturity time overflow for the current timestamp
+        uint256 maturity = block.timestamp + vesting;
+        if (maturity > type(uint32).max) {
+            revert Overflow(maturity, type(uint32).max);
```
kupermind, line 222: Check for the max vesting and maturity limits.

contracts/Depository.sol
```diff
@@ -226,56 +234,55 @@ contract Depository is IErrorsTokenomics {
...
-            }
-            address token = mapBondProducts[productId].token;
-            delete mapBondProducts[productId];
+                address token = mapBondProducts[productId].token;
+                delete mapBondProducts[productId];
 
-            emit CloseProduct(token, productId);
+                numClosedProducts++;
+                emit CloseProduct(token, productId, block.timestamp);
+            }
```
kupermind, line 268: Removed revert for trying to close an already-closed product. Also reduced the check for the supply. If the supply is zero, the product is already closed, otherwise the remaining amount is sent back to tokenomics.

contracts/Depository.sol
```diff
@@ -288,13 +295,19 @@ contract Depository is IErrorsTokenomics {
...
-        // Check for the product expiry
-        // Note that if the token or productId are invalid, the expiry will be zero by default and revert the function
-        expiry = product.expiry;
-        if (expiry < block.timestamp) {
-            revert ProductExpired(token, productId, product.expiry, block.timestamp);
+        // Check for the product vesting time
+        // Note that if the token or productId are invalid, the vesting will be zero by default and result in revert
+        uint256 vesting = product.vesting;
+        if (vesting == 0) {
+            revert ProductExpired(token, productId, vesting, block.timestamp);
```
kupermind, line 302: The product does not expire unless its supply becomes zero, and then it's instantly closed.

contracts/Depository.sol
```diff
@@ -65,13 +65,16 @@ contract Depository is IErrorsTokenomics {
     event TreasuryUpdated(address indexed treasury);
     event BondCalculatorUpdated(address indexed _bondCalculator);
     event CreateBond(address indexed token, uint256 indexed productId, address indexed owner, uint256 bondId,
-        uint256 amountOLAS, uint256 tokenAmount, uint256 expiry);
-    event RedeemBond(uint256 indexed productId, address indexed owner, uint256 bondId);
-    event CreateProduct(address indexed token, uint256 indexed productId, uint256 supply, uint256 priceLP, uint256 expiry);
-    event CloseProduct(address indexed token, uint256 indexed productId);
+        uint256 amountOLAS, uint256 tokenAmount, uint256 maturity, uint256 startTime);
+    event RedeemBond(uint256 indexed productId, address indexed owner, uint256 bondId, uint256 redeemTime);
```
DavidMinarsch, line 69: Why do we need redeemTime here? that will presumably be the timestamp of the transaction which the log is part of?

contracts/Depository.sol
```diff
@@ -65,13 +65,16 @@ contract Depository is IErrorsTokenomics {
...
+    event CreateProduct(address indexed token, uint256 indexed productId, uint256 supply, uint256 priceLP,
+        uint256 vesting, uint256 startTime);
```
DavidMinarsch, line 71: Same as above

contracts/Depository.sol
DavidMinarsch: Same as above

contracts/Depository.sol
```diff
@@ -65,13 +65,16 @@ contract Depository is IErrorsTokenomics {
...
+    event CloseProduct(address indexed token, uint256 indexed productId, uint256 closeTime);
 
     // Minimum bond vesting value
     uint256 public constant MIN_VESTING = 1 days;
+    // Maximum bond vesting value
+    uint256 public constant MAX_VESTING = 4 * 365 days;
```
DavidMinarsch, line 77: Why?

contracts/Depository.sol
DavidMinarsch: Same as above

contracts/Depository.sol
```diff
@@ -208,10 +211,15 @@ contract Depository is IErrorsTokenomics {
...
         }
 
-        // Check for the expiration time overflow
-        uint256 expiry = block.timestamp + vesting;
-        if (expiry > type(uint32).max) {
-            revert Overflow(expiry, type(uint32).max);
+        // Check the vesting maximum limit value
+        if (vesting > MAX_VESTING) {
+            revert Overflow(vesting, MAX_VESTING);
+        }
```
DavidMinarsch, line 217: Seems unnecessary?

contracts/Depository.sol
```diff
@@ -226,56 +234,55 @@ contract Depository is IErrorsTokenomics {
...
 
     /// @dev Closes bonding products.
     /// @notice This will terminate programs regardless of their vesting time.
     /// @param productIds Set of product Ids.
+    /// @return numClosedProducts Number of closed products.
     /// #if_succeeds {:msg "productCounter not touched"} productCounter == old(productCounter);
-    /// #if_succeeds {:msg "success closed"} forall (uint k in productIds) mapBondProducts[productIds[k]].expiry == 0 && mapBondProducts[productIds[k]].supply == 0;
-    function close(uint256[] memory productIds) external {
+    /// #if_succeeds {:msg "success closed"} forall (uint k in productIds) mapBondProducts[productIds[k]].vesting == 0 && mapBondProducts[productIds[k]].supply == 0;
+    function close(uint256[] memory productIds) external returns (uint256 numClosedProducts) {
```
DavidMinarsch, line 249: Hm why not return the ids of the closed products?

contracts/Depository.sol
```diff
@@ -226,56 +234,55 @@ contract Depository is IErrorsTokenomics {
...
         productCounter = uint32(productId + 1);
-        emit CreateProduct(token, productId, supply, priceLP, expiry);
+        emit CreateProduct(token, productId, supply, priceLP, vesting, block.timestamp);
     }
 
     /// @dev Closes bonding products.
     /// @notice This will terminate programs regardless of their vesting time.
     /// @param productIds Set of product Ids.
+    /// @return numClosedProducts Number of closed products.
     /// #if_succeeds {:msg "productCounter not touched"} productCounter == old(productCounter);
```
DavidMinarsch, line 247: Add a fuzzing test to show that product being closed still allows calling this

contracts/Depository.sol
```diff
@@ -344,58 +358,39 @@ contract Depository is IErrorsTokenomics {
...
-            // Close the program if it was not yet closed
-            if (mapBondProducts[productId].expiry > 0) {
-                uint256 supply = mapBondProducts[productId].supply;
-                // Refund unused OLAS supply from the program if not used completely
-                if (supply > 0) {
-                    ITokenomics(tokenomics).refundFromBondProgram(supply);
-                }
-                address token = mapBondProducts[productId].token;
-                delete mapBondProducts[productId];
-
```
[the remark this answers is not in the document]
DavidMinarsch: Please double check throughout that it is not possible for funds to be trapped as a result of this scenario:
1) (multiple) deposit, 2) close, 3) cannot redeem. Any redemption must always be honored irrespective of governance closing products.

contracts/Depository.sol
DavidMinarsch: Hm ... inactive products equivalent are now those with supply zero... might be worth keeping the old format and changing the meaning of active/inactive?

contracts/Depository.sol
DavidMinarsch: I am not yet seeing the benefit of the change to the events.

contracts/Depository.sol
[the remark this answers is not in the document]
77ph: Such a scenario is not possible. redeem not used mapBondProduct. close means no more deposit() after close().

contracts/Depository.sol
77ph: I don't see a scenario to distinguish "inactive" and "non-exists".
It seems to me that the new approach makes existence state "inactive" impossible - if we leave the garbage collection.
Which product will be considered "inactive" but will remain in storage?

contracts/Depository.sol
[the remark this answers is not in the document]
kupermind: Bonds are independent of products

contracts/Depository.sol
kupermind: We didn't have any limit for vesting before, we believe it must be set. Also, better to check for the overflow values with it.

contracts/Depository.sol
kupermind: See above

contracts/Depository.sol
kupermind: @DavidMinarsch this checks for the possible time overflow in a distant future, whereas `MAX_VESTING` limits the vesting period itself.

contracts/Depository.sol
kupermind: We can do this, however one can be read IDs from events. This implementation is just easier.

contracts/Depository.sol
kupermind: Once the supply hits zero, the product is closed. This way things are more straightforward in implementation and there are no additional branches. For historical reasons one can read from events.

contracts/Depository.sol
kupermind: Done in foundry.

contracts/Depository.sol
kupermind: Let's remove as it feels unnecessary

77ph:
in general (if we want to keep the old designations, they only make sense for compatibility).
we can defined:
active product: supply > 0
inactive product: supply = 0 and  numProduct < productCounter . other fields also contain zero.
non-exist product: supply = 0 and  numProduct >= productCounter . other fields also contain zero.
it doesn't make much sense, because no information about the removed (closed)/"inactive" product can be retrieved from the storage.
historical information will still have to be obtained by event

contracts/Depository.sol
```diff
@@ -65,13 +65,16 @@ contract Depository is IErrorsTokenomics {
...
     event BondCalculatorUpdated(address indexed _bondCalculator);
     event CreateBond(address indexed token, uint256 indexed productId, address indexed owner, uint256 bondId,
-        uint256 amountOLAS, uint256 tokenAmount, uint256 expiry);
+        uint256 amountOLAS, uint256 tokenAmount, uint256 maturity);
     event RedeemBond(uint256 indexed productId, address indexed owner, uint256 bondId);
-    event CreateProduct(address indexed token, uint256 indexed productId, uint256 supply, uint256 priceLP, uint256 expiry);
-    event CloseProduct(address indexed token, uint256 indexed productId);
+    event CreateProduct(address indexed token, uint256 indexed productId, uint256 supply, uint256 priceLP,
+        uint256 vesting);
+    event CloseProduct(address indexed token, uint256 indexed productId, uint256 supply);
```
kupermind, line 72: Adding the closing product remaining supply value

contracts/Depository.sol
DavidMinarsch: The thing is, returning the `numClosedProducts` is totally pointless afaict. It tells the "on-chain" caller nothing (as they don't have access to the events). The "off-chain" caller has access to the event in principle but must construct another call to get them.

contracts/Depository.sol
DavidMinarsch: I think the natural limit is max of uint32? So I see that as a relevant bound. Not sure why 4 years has any relevance.

kupermind: Ok, let's revert this change.

contracts/Depository.sol
kupermind: I will restructure.

contracts/Depository.sol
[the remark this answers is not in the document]
DavidMinarsch: Please add a test to confirm that this holds true (i.e. is not possible)!

contracts/Depository.sol
DavidMinarsch: Just pointing out that a product that once got filled (zero remaining supply) is different from one that never existed. I think it's a design smell that we don't have different names for these different products.

contracts/Depository.sol
DavidMinarsch: Yep I understand.

contracts/Depository.sol
kupermind: Let's revert the classification, just to get the product id numbers that are already inactive.

77ph (approved):

contracts/Depository.sol
DavidMinarsch: But here if the product ID is invalid this will be a misleading revert?

DavidMinarsch: I'm confused about the whole condition. Why would vesting ever be zero?

.solhint.json
```diff
@@ -2,7 +2,7 @@
   "extends": "solhint:recommended",
   "plugins": [],
   "rules": {
-    "compiler-version": ["warn", "^0.8.20"],
+    "compiler-version": ["warn", "^0.8.18"],
```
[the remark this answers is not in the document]
kupermind: The version applies from the lowest one in solidity files, this way we are not going to have warnings for compiler with a lower version than 0.8.20

contracts/Depository.sol
```diff
@@ -226,56 +229,71 @@ contract Depository is IErrorsTokenomics {
...
-            emit CloseProduct(token, productId);
+        // Get indexes of closed products
+        closedProductIds = new uint256[](numClosedProducts);
+        uint256 numPos;
+        for (uint256 i = 0; i < numProducts; ++i) {
+            if (positions[i] ) {
+                closedProductIds[numPos] = productIds[i];
+                ++numPos;
+            }
         }
```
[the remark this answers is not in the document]
kupermind: We need to allocate the resulting array of a correct size, that's why we calculate `numClosedProducts` first. But yes, this can be optimized further such that Ids are recorded in the same loop, and then just written to the array with the correct size. Will raise a PR.

contracts/Depository.sol
kupermind: The vesting is zero when the product is inactive or never existed. The revert is not confusing in a sense that the product can't be used. We are just utilizing the revert that was defined from the previous version as this error is part of the contract anyway. We can redefine another one and call it `ProductInactiveOrNonExistent(productId)`, would that work better?

kupermind: We also have a `ProductClosed(uint256 productId)` error, I think it is good enough.

PR #170, opened 2024-07-14
AIP-1 bonding

kupermind:
- Initial implementation and testing of dynamic IDF and bonding refactor.

### DIFF
--- a/contracts/BondCalculator.sol
+++ b/contracts/BondCalculator.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.25;
 
 import {mulDiv} from "@prb/math/src/Common.sol";
+import {GenericBondCalculator} from "./GenericBondCalculator.sol";
 import {IVotingEscrow} from "./interfaces/IVotingEscrow.sol";
 
 interface ITokenomics {
@@ -22,6 +23,9 @@ error Overflow(uint256 provided, uint256 max);
 
 /// @dev Provided zero address.
 error ZeroAddress();
+
+/// @dev Provided zero value.
+error ZeroValue();
 
 // Struct for discount factor params
 // The size of the struct is 96 + 64 + 64 = 224 (1 slot)
@@ -56,22 +60,18 @@ struct Product {
     uint32 vesting;
 }
 
-/// @title GenericBondSwap - Smart contract for generic bond calculation mechanisms in exchange for OLAS tokens.
-/// @dev The bond calculation mechanism is based on the UniswapV2Pair contract.
-/// @author AL
+/// @title BondCalculator - Smart contract for bond calculation payout in exchange for OLAS tokens based on dynamic IDF.
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
-contract GenericBondCalculator {
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract BondCalculator is GenericBondCalculator {
     event OwnerUpdated(address indexed owner);
     event DiscountParamsUpdated(DiscountParams newDiscountParams);
 
     // Maximum sum of discount factor weights
     uint256 public constant MAX_SUM_WEIGHTS = 10_000;
-    // OLAS contract address
-    address public immutable olas;
     // veOLAS contract address
     address public immutable ve;
-    // Tokenomics contract address
-    address public immutable tokenomics;
 
     // Contract owner
     address public owner;
@@ -79,20 +79,26 @@ contract GenericBondCalculator {
     DiscountParams public discountParams;
 
 
-    /// @dev Generic Bond Calcolator constructor
+    /// @dev Bond Calculator constructor.
     /// @param _olas OLAS contract address.
     /// @param _tokenomics Tokenomics contract address.
-    constructor(address _olas, address _ve, address _tokenomics, DiscountParams memory _discountParams) {
-        // Check for at least one zero contract address
-        if (_olas == address(0) || _ve == address(0) || _tokenomics == address(0)) {
+    /// @param _ve veOLAS contract address.
+    /// @param _discountParams Discount factor parameters.
+    constructor(address _olas, address _tokenomics, address _ve, DiscountParams memory _discountParams)
+        GenericBondCalculator(_olas, _tokenomics)
+    {
+        // Check for zero address
+        if (_ve == address(0)) {
             revert ZeroAddress();
         }
 
-        olas = _olas;
         ve = _ve;
-        tokenomics = _tokenomics;
         owner = msg.sender;
-        
+
+        // Check for zero values
+        if (_discountParams.targetNewUnits == 0 || _discountParams.targetVotingPower == 0) {
+            revert ZeroValue();
+        }
         // Check the sum of factors that cannot exceed the value of 10_000 (100% with a 0.01% step)
         uint256 sumWeights;
         for (uint256 i = 0; i < _discountParams.weightFactors.length; ++i) {
@@ -129,6 +135,10 @@ contract GenericBondCalculator {
             revert OwnerOnly(msg.sender, owner);
         }
 
+        // Check for zero values
+        if (newDiscountParams.targetNewUnits == 0 || newDiscountParams.targetVotingPower == 0) {
+            revert ZeroValue();
+        }
         // Check the sum of factors that cannot exceed the value of 10_000 (100% with a 0.01% step)
         uint256 sumWeights;
         for (uint256 i = 0; i < newDiscountParams.weightFactors.length; ++i) {
@@ -143,26 +153,16 @@ contract GenericBondCalculator {
         emit DiscountParamsUpdated(newDiscountParams);
     }
 
-    /// @dev Calculates the amount of OLAS tokens based on the bonding calculator mechanism.
-    /// @notice Currently there is only one implementation of a bond calculation mechanism based on the UniswapV2 LP.
-    /// @notice IDF has a 10^18 multiplier and priceLP has the same as well, so the result must be divided by 10^36.
+    /// @dev Calculates the amount of OLAS tokens based on the bonding calculator mechanism accounting for dynamic IDF.
     /// @param tokenAmount LP token amount.
     /// @param priceLP LP token price.
-    /// @param bondVestingTime Bond vesting time.
-    /// @param productMaxVestingTime Product max vesting time.
-    /// @param productSupply Current product supply.
-    /// @param productPayout Current product payout.
+    /// @param data Custom data that is used to calculate the IDF.
     /// @return amountOLAS Resulting amount of OLAS tokens.
-    /// #if_succeeds {:msg "LP price limit"} priceLP * tokenAmount <= type(uint192).max;
     function calculatePayoutOLAS(
-        address account,
         uint256 tokenAmount,
         uint256 priceLP,
-        uint256 bondVestingTime,
-        uint256 productMaxVestingTime,
-        uint256 productSupply,
-        uint256 productPayout
-    ) external view returns (uint256 amountOLAS) {
+        bytes memory data
+    ) external view override returns (uint256 amountOLAS) {
         // The result is divided by additional 1e18, since it was multiplied by in the current LP price calculation
         // The resulting amountDF can not overflow by the following calculations: idf = 64 bits;
         // priceLP = 2 * r0/L * 10^18 = 2*r0*10^18/sqrt(r0*r1) ~= 61 + 96 - sqrt(96 * 112) ~= 53 bits (if LP is balanced)
@@ -178,7 +178,7 @@ contract GenericBondCalculator {
         }
 
         // Calculate the dynamic inverse discount factor
-        uint256 idf = calculateIDF(account, bondVestingTime, productMaxVestingTime, productSupply, productPayout);
+        uint256 idf = calculateIDF(data);
 
         // Amount with the discount factor is IDF * priceLP * tokenAmount / 1e36
         // At this point of time IDF is bound by the max of uint64, and totalTokenValue is no bigger than the max of uint192
@@ -186,26 +186,25 @@ contract GenericBondCalculator {
     }
 
     /// @dev Calculated inverse discount factor based on bonding and account parameters.
-    /// @param account Bonding account address.
-    /// @param bondVestingTime Bonding desired vesting time.
-    /// @param productMaxVestingTime Product max vesting time.
-    /// @param productSupply Current product supply.
-    /// @param productPayout Current product payout.
+    /// @param data Custom data that is used to calculate the IDF:
+    ///        - account Account address.
+    ///        - bondVestingTime Bond vesting time.
+    ///        - productMaxVestingTime Product max vesting time.
+    ///        - productSupply Current product supply.
+    ///        - productPayout Current product payout.
     /// @return idf Inverse discount factor in 18 decimals format.
-    function calculateIDF(
-        address account,
-        uint256 bondVestingTime,
-        uint256 productMaxVestingTime,
-        uint256 productSupply,
-        uint256 productPayout
-    ) public view returns (uint256 idf) {
+    function calculateIDF(bytes memory data) public view virtual returns (uint256 idf) {
+        // Decode the required data
+        (address account, uint256 bondVestingTime, uint256 productMaxVestingTime, uint256 productSupply,
+            uint256 productPayout) = abi.decode(data, (address, uint256, uint256, uint256, uint256));
+
         // Get the copy of the discount params
         DiscountParams memory localParams = discountParams;
         uint256 discountBooster;
 
         // First discount booster: booster = k1 * NumNewUnits(previous epoch) / TargetNewUnits(previous epoch)
         // Check the number of new units coming from tokenomics vs the target number of new units
-        if (localParams.targetNewUnits > 0) {
+        if (localParams.weightFactors[0] > 0) {
             uint256 numNewUnits = ITokenomics(tokenomics).getLastEpochNumNewUnits();
 
             // If the number of new units exceeds the target, bound by the target number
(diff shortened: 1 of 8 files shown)


contracts/Depository.sol
```diff
@@ -40,50 +61,56 @@ struct Bond {
...
-    // Token to accept as a payment
-    address token;
     // Supply of remaining OLAS tokens
     // After 10 years, the OLAS inflation rate is 2% per year. It would take 220+ years to reach 2^96 - 1
     uint96 supply;
+    // Token to accept as a payment
+    address token;
+    // Current OLAS payout
+    // This value is bound by the initial total supply
+    uint96 payout;
```
kupermind, line 77: Adding as part of dynamic IDF requirement

contracts/Depository.sol
```diff
@@ -25,10 +48,8 @@ import {ITreasury} from "./interfaces/ITreasury.sol";
 * In conclusion, this contract is only safe to use until 2106.
 */
 
-// The size of the struct is 160 + 96 + 32 * 2 = 256 + 64 (2 slots)
+// The size of the struct is 96 + 32 * 2 = 160 (1 slot)
 struct Bond {
-    // Account address
-    address account;
```
kupermind, line 31: Account is not needed as bonds become ERC-721 tokens

contracts/Depository.sol
```diff
@@ -285,10 +330,10 @@ contract Depository is IErrorsTokenomics {
     /// #if_succeeds {:msg "token is valid"} mapBondProducts[productId].token != address(0);
     /// #if_succeeds {:msg "input supply is non-zero"} old(mapBondProducts[productId].supply) > 0 && mapBondProducts[productId].supply <= type(uint96).max;
     /// #if_succeeds {:msg "vesting is non-zero"} mapBondProducts[productId].vesting > 0 && mapBondProducts[productId].vesting + block.timestamp <= type(uint32).max;
-    /// #if_succeeds {:msg "bond Id"} bondCounter == old(bondCounter) + 1 && bondCounter <= type(uint32).max;
+    /// #if_succeeds {:msg "bond Id"} totalSupply == old(totalSupply) + 1 && totalSupply <= type(uint32).max;
     /// #if_succeeds {:msg "payout"} old(mapBondProducts[productId].supply) == mapBondProducts[productId].supply + payout;
     /// #if_succeeds {:msg "OLAS balances"} IToken(mapBondProducts[productId].token).balanceOf(treasury) == old(IToken(mapBondProducts[productId].token).balanceOf(treasury)) + tokenAmount;
-    function deposit(uint256 productId, uint256 tokenAmount) external
+    function deposit(uint256 productId, uint256 tokenAmount, uint256 bondVestingTime) external
```
kupermind, line 336: As part of the AIP-1 for dynamic IDF, the bonder can choose the bond vesting time, just not to be below the minimum allowed one.

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,232 @@
...
+        owner = msg.sender;
+
+        // Check for zero values
+        if (_discountParams.targetNewUnits == 0 || _discountParams.targetVotingPower == 0) {
+            revert ZeroValue();
+        }
+        // Check the sum of factors that cannot exceed the value of 10_000 (100% with a 0.01% step)
+        uint256 sumWeights;
+        for (uint256 i = 0; i < _discountParams.weightFactors.length; ++i) {
+            sumWeights += _discountParams.weightFactors[i];
```
77ph, line 104: _discountParams.weightFactors[i] must > min and sumWeights > 0

contracts/GenericBondCalculator.sol
```diff
@@ -36,16 +35,23 @@ contract GenericBondCalculator {
...
+    /// @param data Custom data to calculate the IDF.
     /// @return amountOLAS Resulting amount of OLAS tokens.
-    /// #if_succeeds {:msg "LP price limit"} priceLP * tokenAmount <= type(uint192).max;
-    function calculatePayoutOLAS(uint256 tokenAmount, uint256 priceLP) external view
-        returns (uint256 amountOLAS)
-    {
+    function calculatePayoutOLAS(
+        uint256 tokenAmount,
+        uint256 priceLP,
+        bytes memory data
```
77ph, line 53: I'm not sure that we should use implicit pack of arguments as a raw sequence of bytes (memory data) and then unpack it (abi.decode) instead of explicitly enumerating the arguments.

contracts/Depository.sol
```diff
@@ -315,9 +368,11 @@ contract Depository is IErrorsTokenomics {
         // Get the LP token address
         address token = product.token;
 
-        // Calculate the payout in OLAS tokens based on the LP pair with the discount factor (DF) calculation
+        // Calculate the payout in OLAS tokens based on the LP pair with the inverse discount factor (IDF) calculation
         // Note that payout cannot be zero since the price LP is non-zero, otherwise the product would not be created
-        payout = IGenericBondCalculator(bondCalculator).calculatePayoutOLAS(tokenAmount, product.priceLP);
+        payout = IBondCalculator(bondCalculator).calculatePayoutOLAS(tokenAmount, product.priceLP,
+            // Encode parameters required for the IDF calculation
+            abi.encode(msg.sender, bondVestingTime, productMaxVestingTime, supply, product.payout));
```
77ph, line 375: abi.encode(msg.sender, bondVestingTime, productMaxVestingTime, supply, product.payout)
This design seems like overengineering to me.
better just
IBondCalculator(bondCalculator).calculatePayoutOLAS(tokenAmount, product.priceLP,
            // Encode parameters required for the IDF calculation
            msg.sender, bondVestingTime, productMaxVestingTime, supply, product.payout);

contracts/Depository.sol
```diff
@@ -326,16 +381,34 @@ contract Depository is IErrorsTokenomics {
...
+            payout += supply;
+            supply = 0;
+        }
         product.supply = uint96(supply);
+        product.payout += uint96(payout);
+
+        // Create and mint a new bond
+        bondId = totalSupply;
+        // Safe mint is needed since contracts can create bonds as well
+        _safeMint(msg.sender, bondId);
```
77ph, line 395: audit notes: re-check reentrancy in internal audit

77ph (dismissed):

contracts/BondCalculator.sol
kupermind: weights can all be zero, leaving as is.

contracts/Depository.sol
kupermind: service Ids indexing must start from zero due to possible mapping conflicts. There is no such thing for bonds.

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+/// @title BondCalculator - Smart contract for bond calculation payout in exchange for OLAS tokens based on dynamic IDF.
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+/// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
+contract BondCalculator {
+    event OwnerUpdated(address indexed owner);
+    event DiscountParamsUpdated(DiscountParams newDiscountParams);
+
+    // Maximum sum of discount factor weights
+    uint256 public constant MAX_SUM_WEIGHTS = 10_000;
```
kupermind, line 72: Sum of weights for `k` fractions - using the Vote Weighting terminology let them be from 0 to 100% with a 0.01% step.

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+    // DAO set voting power limit for the bonding account
+    // This value is bound by the veOLAS total voting power
+    uint96 targetVotingPower;
+    // DAO set number of new units per epoch limit
+    // This number is bound by the total number of possible components and agents
+    uint64 targetNewUnits;
+    // DAO set weight factors
+    // The sum of factors cannot exceed the value of 10_000 (100% with a 0.01% step)
+    uint16[4] weightFactors;
+}
```
kupermind, line 42: Corresponds to 2 variables and 4 weights set by the DAO.

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+    /// @dev Calculates the amount of OLAS tokens based on the bonding calculator mechanism accounting for dynamic IDF.
+    /// @param account Account address.
+    /// @param tokenAmount LP token amount.
+    /// @param priceLP LP token price.
+    /// @param bondVestingTime Bond vesting time.
+    /// @param productMaxVestingTime Product max vesting time.
+    /// @param productSupply Current product supply.
+    /// @param productPayout Current product payout.
+    /// @return amountOLAS Resulting amount of OLAS tokens.
+    function calculatePayoutOLAS(
```
kupermind, line 240: This is left as is from the `GenericBondCalculator`, which is now obsolete.

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+        if (totalTokenValue > type(uint192).max) {
+            revert Overflow(totalTokenValue, type(uint192).max);
+        }
+
+        // Calculate the dynamic inverse discount factor
+        uint256 idf = calculateIDF(account, bondVestingTime, productMaxVestingTime, productSupply, productPayout);
+
+        // Amount with the discount factor is IDF * priceLP * tokenAmount / 1e36
+        // At this point of time IDF is bound by the max of uint64, and totalTokenValue is no bigger than the max of uint192
+        amountOLAS = (idf * totalTokenValue) / 1e36;
```
kupermind, line 268: This is the only new difference - tokenomics-calculated IDF is changed with the local calculated dynamic IDF according to the proposed approach.

contracts/Depository.sol
```diff
@@ -305,8 +358,16 @@ contract Depository is IErrorsTokenomics {
...
         }
 
+        uint256 productMaxVestingTime = product.vesting;
+        // Calculate vesting limits
+        if (bondVestingTime < MIN_VESTING) {
+            revert LowerThan(bondVestingTime, MIN_VESTING);
+        }
+        if (bondVestingTime > productMaxVestingTime) {
+            revert Overflow(bondVestingTime, productMaxVestingTime);
+        }
```
kupermind, line 368: Bond vesting time is now chosen by the user from `MIN_VESTING` to max product defined vesting time.

contracts/Depository.sol
```diff
@@ -315,9 +376,10 @@ contract Depository is IErrorsTokenomics {
         // Get the LP token address
         address token = product.token;
 
-        // Calculate the payout in OLAS tokens based on the LP pair with the discount factor (DF) calculation
+        // Calculate the payout in OLAS tokens based on the LP pair with the inverse discount factor (IDF) calculation
         // Note that payout cannot be zero since the price LP is non-zero, otherwise the product would not be created
-        payout = IGenericBondCalculator(bondCalculator).calculatePayoutOLAS(tokenAmount, product.priceLP);
+        payout = IBondCalculator(bondCalculator).calculatePayoutOLAS(msg.sender, tokenAmount, product.priceLP,
```
kupermind, line 381: Calls new `BondCalculator` contract based on proposed dynamic IDF.

contracts/Depository.sol
```diff
@@ -326,16 +388,34 @@ contract Depository is IErrorsTokenomics {
 
         // Decrease the supply for the amount of payout
         supply -= payout;
+        // Adjust payout and set supply to zero if supply drops below the min defined value
+        if (supply < minOLASLeftoverAmount) {
+            payout += supply;
+            supply = 0;
+        }
```
kupermind, line 395: If the product supply drops below a specified value, it's remainder is given to the bonder, and the product is instantly closed.

contracts/Depository.sol
```diff
@@ -40,54 +69,60 @@ struct Bond {
...
+    // Max bond vesting time
+    // 2^32 - 1 is enough to count 136 years starting from the year of 1970. This counter is safe until the year of 2106
+    uint32 vesting;
 }
 
 /// @title Bond Depository - Smart contract for OLAS Bond Depository
 /// @author AL
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
-contract Depository is IErrorsTokenomics {
+contract Depository is ERC721, IErrorsTokenomics {
```
kupermind, line 94: Bonds become NFT-s, yay!

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+error ZeroAddress();
+
+/// @dev Provided zero value.
+error ZeroValue();
+
+// Struct for discount factor params
+// The size of the struct is 96 + 64 + 64 = 224 (1 slot)
+struct DiscountParams {
+    // DAO set voting power limit for the bonding account
+    // This value is bound by the veOLAS total voting power
```
DavidMinarsch, line 34: Description is not very clear. what does the value affect?

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+    uint96 targetVotingPower;
+    // DAO set number of new units per epoch limit
```
DavidMinarsch, line 36: New units for what?

contracts/GenericBondCalculator.sol
```diff
@@ -1,93 +0,0 @@
-// SPDX-License-Identifier: MIT
```
DavidMinarsch, line 1: Why is all this logic no longer needed?

contracts/Tokenomics.sol
```diff
@@ -664,15 +661,6 @@ contract Tokenomics is TokenomicsConstants {
...
         }
 
-        // Check the epsilonRate value for idf to fit in its size
-        // 2^64 - 1 < 18.5e18, idf is equal at most 1 + epsilonRate < 18e18, which fits in the variable size
-        // epsilonRate is the part of the IDF calculation and thus its change will be accounted for in the next epoch
-        if (_epsilonRate > 0 && _epsilonRate <= 17e18) {
-            epsilonRate = uint64(_epsilonRate);
-        } else {
-            _epsilonRate = epsilonRate;
-        }
```
DavidMinarsch, line 674: why was this dropped?

contracts/Tokenomics.sol
```diff
@@ -962,13 +950,6 @@ contract Tokenomics is TokenomicsConstants {
                     if (!mapNewUnits[unitType][serviceUnitIds[j]]) {
                         mapNewUnits[unitType][serviceUnitIds[j]] = true;
                         mapEpochTokenomics[curEpoch].unitPoints[unitType].numNewUnits++;
-                        // Check if the owner has introduced component / agent for the first time
-                        // This is done together with the new unit check, otherwise it could be just a new unit owner
-                        address unitOwner = IToken(registries[unitType]).ownerOf(serviceUnitIds[j]);
-                        if (!mapNewOwners[unitOwner]) {
-                            mapNewOwners[unitOwner] = true;
-                            mapEpochTokenomics[curEpoch].epochPoint.numNewOwners++;
-                        }
```
DavidMinarsch, line 971: why removed?

contracts/Tokenomics.sol
```diff
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.25;
```
DavidMinarsch, line 2: Not sure we should change Tokenomics at all. Feels like we could get away with keeping the existing implementation ...

contracts/Depository.sol
```diff
@@ -100,21 +135,39 @@ contract Depository is IErrorsTokenomics {
...
+        // Check for base URI zero value
+        if (bytes(_baseURI).length == 0) {
+            revert ZeroValue();
+        }
+
         olas = _olas;
         tokenomics = _tokenomics;
         treasury = _treasury;
         bondCalculator = _bondCalculator;
+        baseURI = _baseURI;
```
DavidMinarsch, line 169: Doesn't make sense here imo. We can have a purely on-chain NFT. There are these SVG libraries that are pretty cool. We don't want to have to serve custom images here, just unnecessary.

contracts/Depository.sol
DavidMinarsch: What's the dynamic IDF again - how does it differ from previous one?

contracts/BondCalculator.sol
```diff
@@ -0,0 +1,301 @@
...
+            revert ZeroAddress();
+        }
+
+        owner = newOwner;
+        emit OwnerUpdated(newOwner);
+    }
+
+    /// @dev Changed inverse discount factor parameters.
+    /// @param newDiscountParams Struct of new discount parameters.
+    function changeDiscountParams(DiscountParams memory newDiscountParams) external {
```
DavidMinarsch, line 136: So this is global, not product specific?

contracts/BondCalculator.sol
DavidMinarsch: ok

contracts/Depository.sol
kupermind: Dynamic IDF based on 4 parameters described in the doc: number of new units per epoch, vesting time (bound by the product vesting time), bond liquidity coefficient (the more one buys from the beginning of the product, the bigger IDF they get), veOLAS amount of the bonder.

In the previous approach the IDF was a pure tokenomics value based on the number of new units added in the epoch, and new owners providing units during the epoch as well.

contracts/BondCalculator.sol
kupermind: Added description in #182

contracts/BondCalculator.sol
kupermind: Added a description in https://github.com/valory-xyz/autonolas-tokenomics/pull/179

contracts/GenericBondCalculator.sol
kupermind: Generic bond calculator used the old IDF approach, that is no longer the case. The new approach requires a new set of functions.

contracts/Tokenomics.sol
kupermind: `epsilonRate` was used to set the default addition to the discount factor set by the DAO. In a new IDF approach, its value varies from 1 to 2 only (whereas in the old approach it could go up to 17), and thus no defaults are needed.

contracts/Tokenomics.sol
kupermind: This is no longer used anywhere in the calculations.

contracts/Tokenomics.sol
kupermind: Dev incentives revision force us to change tokenomics anyway.

contracts/BondCalculator.sol
kupermind: Yes, as otherwise this would need to be voted on for every product.

contracts/Depository.sol
kupermind: Actually we don't need any link at all. All we need to comply with the standard is that the function `tokenURI()` does not revert. So it can be as simple as just returning an NFT id as a string.

contracts/BondCalculator.sol
DavidMinarsch: Technically each product is voted on. So wouldn't it be more useful to be able to set this as part of the product? Just thinking out loud.

contracts/Tokenomics.sol
DavidMinarsch: I see. However, ETH is still being used as part of the donation and distributed to all components/agents. Is it distributed according to the same proportion as the boosts?

contracts/BondCalculator.sol
kupermind: Considering how products are launched, those would be pretty much the same. Also, if something happens on a protocol level, my understanding is that we want all the bonds to follow new parameters even for unfinished products, since otherwise the old params could become wrong / unjustified.

contracts/Tokenomics.sol
kupermind: The ETH distribution is going to be left as is. Meaning that proportions will be different.

PR #179, opened 2024-07-25
Introducing tokenomics changes to reflect on new top-ups approach

kupermind:
- Introducing tokenomics changes to reflect on new top-ups approach.

PR #178, opened 2024-07-22
chore: adding scripts to update owners on L2

kupermind:
- Adding scripts to update owners on L2;
- Adding to the static audit to make sure the contracts are set up correctly.

PR #187, opened 2024-09-13
Fixed typo in variable name and optimized array length handling

saniksin:
1. Fixed typo in variable 
`// Tkenomics contract address > // Tokenomics contract address`

2. Added checks for non-empty arrays in some places and stored their length in a variable.
 - contracts/DonatorBlacklist.sol > setDonatorsStatuses
 - contracts/Tokenomics.sol > getOwnerIncentives

3. In the TokenomicsConstants.sol contract, it might be worth moving `inflationAmounts` and `supplyCaps` to constant arrays.

### DIFF
--- a/contracts/Depository.sol
+++ b/contracts/Depository.sol
@@ -87,7 +87,7 @@ contract Depository is IErrorsTokenomics {
 
     // OLAS token address
     address public immutable olas;
-    // Tkenomics contract address
+    // Tokenomics contract address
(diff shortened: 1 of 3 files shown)


DavidMinarsch:
Thanks @saniksin - these are deployed contracts and therefore it does not make sense to alter their implementation now. We welcome contributions to on-going work like #179 and #178 for instance

PR #200, opened 2025-01-07
feat: accounting for inflation update in tokenomics

kupermind:
- Accounting for inflation update in tokenomics.

### DIFF
--- a/contracts/Tokenomics.sol
+++ b/contracts/Tokenomics.sol
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
         return true;
     }
 
+    /// @dev Updates inflation per second if .
+    function updateInflationPerSecond() external {
+        // Check for the contract ownership
+        if (msg.sender != owner) {
+            revert OwnerOnly(msg.sender, owner);
+        }
+
+        // Get current epoch length
+        uint256 curEpochLen = epochLen;
+        // Find if the year changes within the epoch
+        uint256 numYears = (block.timestamp + curEpochLen - timeLaunch) / ONE_YEAR;
+        // Revert if the year changes within the next epoch as it requires more complicated set of calculations
+        if (numYears > currentYear) {
+            revert Overflow(numYears, currentYear);
+        }
+
+        // Ensure effectiveBond is bigger than maxBond
+        uint256 curEffectiveBond = effectiveBond;
+        uint256 curMaxBond = maxBond;
+        if (curMaxBond > curEffectiveBond) {
+            revert Overflow(curMaxBond, curEffectiveBond);
+        }
+
+        // Adjust effective bond by reducing it with the on-going epoch maxBond value
+        curEffectiveBond -= curMaxBond;
+
+        // Recalculate inflation per second based on the updated current year inflation
+        uint256 curInflationPerSecond = getInflationForYear(currentYear) / ONE_YEAR;
+
+        // maxBond has to be recalculated
+        curMaxBond = (curEpochLen * curInflationPerSecond * mapEpochTokenomics[epochCounter].epochPoint.maxBondFraction) / 100;
+        // Adjust effective bond with a new maxBond value
+        curEffectiveBond += curMaxBond;
+
+        // Update state variables
+        maxBond = uint96(curMaxBond);
+        effectiveBond = uint96(curEffectiveBond);
+        inflationPerSecond = uint96(curInflationPerSecond);
+    }
+
     /// @dev Gets component / agent owner incentives and clears the balances.
     /// @notice `account` must be the owner of components / agents Ids, otherwise the function will revert.
     /// @notice If not all `unitIds` belonging to `account` were provided, they will be untouched and keep accumulating.
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -251,7 +251,7 @@ module.exports = {
                 settings: {
                     optimizer: {
                         enabled: true,
-                        runs: 1500,
+                        runs: 750,
                     },
                     evmVersion: "cancun"
                 },


contracts/Tokenomics.sol
```diff
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
...
+
+        // maxBond has to be recalculated
+        curMaxBond = (curEpochLen * curInflationPerSecond * mapEpochTokenomics[epochCounter].epochPoint.maxBondFraction) / 100;
+        // Adjust effective bond with a new maxBond value
+        curEffectiveBond += curMaxBond;
+
+        // Update state variables
+        maxBond = uint96(curMaxBond);
+        effectiveBond = uint96(curEffectiveBond);
+        inflationPerSecond = uint96(curInflationPerSecond);
```
kupermind, line 1336: These are critical values that need to be updated.

hardhat.config.js
```diff
@@ -251,7 +251,7 @@ module.exports = {
                 settings: {
                     optimizer: {
                         enabled: true,
-                        runs: 1500,
+                        runs: 750,
```
kupermind, line 254: Need to reduce optimization iterations in order to fit into deployment size

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
...
+        }
+
+        // Get current epoch length
+        uint256 curEpochLen = epochLen;
+        // Find if the year changes within the epoch
+        uint256 numYears = (block.timestamp + curEpochLen - timeLaunch) / ONE_YEAR;
+        // Revert if the year changes within the next epoch as it requires more complicated set of calculations
+        if (numYears > currentYear) {
+            revert Overflow(numYears, currentYear);
+        }
```
DavidMinarsch, line 1313: hm - can you be more explicit what happens then?

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
...
+
+        // Ensure effectiveBond is bigger than maxBond
+        uint256 curEffectiveBond = effectiveBond;
+        uint256 curMaxBond = maxBond;
```
DavidMinarsch, line 1317: what's maxBond?

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
...
+        if (curMaxBond > curEffectiveBond) {
+            revert Overflow(curMaxBond, curEffectiveBond);
+        }
+
+        // Adjust effective bond by reducing it with the on-going epoch maxBond value
+        curEffectiveBond -= curMaxBond;
+
+        // Recalculate inflation per second based on the updated current year inflation
```
DavidMinarsch, line 1325: If the curve hasn't changed, the value will be the same as before right? We should have a test

contracts/Tokenomics.sol
DavidMinarsch: So we only need to do this for bonding?

kupermind: Correct, bonding is the only one that is issued as a credit for epoch, the rest is end-of-epoch-computed.

contracts/Tokenomics.sol
kupermind: When the year changes, the complex re-calculation of inflation before the exact second year end and the rest of the epoch happens, so it's better to avoid to perform another change during that time.

contracts/Tokenomics.sol
kupermind: max amount that can go for bonding per epoch.

contracts/Tokenomics.sol
kupermind: Correct

contracts/Tokenomics.sol
DavidMinarsch: OK

DavidMinarsch: Please add as much detail in the contract comment for future

contracts/Tokenomics.sol
DavidMinarsch: Let's add this as a comment to the contract and also write a test for it

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
...
+
+        // Ensure effectiveBond is bigger than maxBond
+        uint256 curEffectiveBond = effectiveBond;
+        uint256 curMaxBond = maxBond;
+        if (curMaxBond > curEffectiveBond) {
+            revert Overflow(curMaxBond, curEffectiveBond);
+        }
+
+        // Adjust effective bond by reducing it with the on-going epoch maxBond value
+        curEffectiveBond -= curMaxBond;
```
DavidMinarsch, line 1323: don't follow

kupermind: maxBond is given as a credit for epoch, so we need to subtract what is already given in the current epoch based on previous inflation, and add it again later based on a new inflation.

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1296,46 @@ contract Tokenomics is TokenomicsConstants {
...
+
+        // Recalculate inflation per second based on the updated current year inflation
+        uint256 curInflationPerSecond = getInflationForYear(currentYear) / ONE_YEAR;
```
77ph, line 1326: @kupermind @mariapiamo 
Is there some unaccounted/side effect between the current implementation, when it changes strictly once a year and in the constants there is a mutually agreed table of inflation per year with the fact that now inflation can change an arbitrary number of times during the year.

77ph, line 1329: Needing new implementation getInflationForYear()

kupermind: We just need to give an extra buffer accounting for the current inflation and for the time when it's going to change to another one.

mariapiamo (approved):

mariapiamo:
we just need to fix gitleaks

mariapiamo (dismissed):

77ph (approved):

contracts/TokenomicsConstants.sol
```diff
@@ -74,20 +75,95 @@ abstract contract TokenomicsConstants {
...
+
+            // Inflation amount is the difference between last two caps (inflation for the current year)
+            inflationAmount = (supplyCap * maxMintCapFraction) / 100;
+        }
+    }
+
+    /// @dev Gets actual inflation cap for a specific year.
+    /// @param numYears Number of years passed from the launch date.
+    /// @return supplyCap Supply cap.
+    function getActualSupplyCapForYear(uint256 numYears) public pure returns (uint256 supplyCap) {
```
kupermind, line 109: The difference between original and actual supply caps functions is that in the original ones we preserve the theoretical inflation numbers to match the historical tokenomics points, whereas in actual ones we reflect the real inflation happening over first 3 years.

contracts/TokenomicsConstants.sol
```diff
@@ -74,20 +75,95 @@ abstract contract TokenomicsConstants {
...
+                702_281_516e18,
+                731_564_805e18,
+                761_726_593e18
+            ];
+            supplyCap = supplyCaps[numYears];
+        } else {
+            // Number of years after ten years have passed (including ongoing ones)
+            numYears -= 9;
+            // Max cap for the first 10 years
+            supplyCap = 761_726_593e18;
```
DavidMinarsch, line 129: can we not use the constant definition once rather than multiple times?

contracts/TokenomicsConstants.sol
DavidMinarsch: what are "tokenomics points"?

where are the old ones used?

DavidMinarsch (dismissed):

contracts/Tokenomics.sol
```diff
@@ -277,6 +283,8 @@ contract Tokenomics is TokenomicsConstants {
     event EpochSettled(uint256 indexed epochCounter, uint256 treasuryRewards, uint256 accountRewards,
         uint256 accountTopUps, uint256 effectiveBond, uint256 returnedStakingIncentive, uint256 totalStakingIncentive);
     event TokenomicsImplementationUpdated(address indexed implementation);
+    event InflationPerSecondFractionsUpdated(uint256 inflationPerSecond, uint256 maxBond, uint256 maxBondFraction,
```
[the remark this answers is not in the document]
kupermind: Yes, it's global

contracts/TokenomicsConstants.sol
kupermind: Old ones are used for the historical reasons. Tokenomics points are produced every checkpoint (end of epoch) with all the essential into written into them.

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1305,84 @@ contract Tokenomics is TokenomicsConstants {
...
+        uint256 curEpochCounter = epochCounter;
+        uint256 endTime = mapEpochTokenomics[curEpochCounter - 1].epochPoint.endTime;
+        // Find if this epoch is the year change epoch
+        uint256 lastEpochYear = (endTime - timeLaunch) / ONE_YEAR;
+        if (numYears > lastEpochYear) {
+            revert Overflow(numYears, lastEpochYear);
+        }
+
+        // Reset staking inflation
+        // Assume there are no more than 100 epochs to clear retained staking amount
```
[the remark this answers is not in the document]
kupermind: No, this is unrealistic as 100 epochs means 9 years, and the function will run out of gas (for now).

contracts/Tokenomics.sol
```diff
@@ -1296,6 +1305,84 @@ contract Tokenomics is TokenomicsConstants {
...
+            // Break when all the retains are executed
+            if (!success) {
+                break;
+            }
+        }
+        mapEpochStakingPoints[curEpochCounter].stakingIncentive = 0;
+
+        // Update OLAS inflation fractions
+        uint256 sumFractions = maxBondFraction + topUpComponentFraction + topUpAgentFraction + stakingFraction;
+        if (sumFractions > 100) {
```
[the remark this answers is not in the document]
kupermind: It's allowed by our protocol, same as in general behavior.

contracts/TokenomicsConstants.sol
kupermind: Yes sure, will optimize

mariapiamo (approved):

77ph (approved):

PR #217, opened 2025-06-23
refactor: account for withheld amount update flag

kupermind:
-  Account for withheld amount update flag.

### DIFF
--- a/test/StakingBridging.js
+++ b/test/StakingBridging.js
@@ -416,11 +416,11 @@ describe("StakingBridging", async () => {
             // Process data maintenance by the owner
             const payload = ethers.utils.defaultAbiCoder.encode(["address[]", "uint256[]", "bytes32"],
                 [[stakingTarget], [stakingIncentive * 2], batchHash]);
-            await arbitrumTargetDispenserL2.processDataMaintenance(payload, true);
+            await arbitrumTargetDispenserL2.processDataMaintenance(payload, false);
 
             // Try to do it not from the owner
             await expect(
-                arbitrumTargetDispenserL2.connect(signers[1]).processDataMaintenance("0x", true)
+                arbitrumTargetDispenserL2.connect(signers[1]).processDataMaintenance("0x", false)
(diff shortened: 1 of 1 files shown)


contracts/staking/DefaultTargetDispenserL2.sol
```diff
@@ -140,7 +140,7 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
 
     /// @dev Processes the data received from L1.
     /// @param data Bytes message data sent from L1.
-    function _processData(bytes memory data) internal {
+    function _processData(bytes memory data) internal returns (uint256 totalAmount) {
```
kupermind, line 143: Since proposed deposited amounts could be further cut off by the StakingFactory verification and result in more `withheldAmount`-s, let's compute the total amount that is deposited.

contracts/staking/DefaultTargetDispenserL2.sol
```diff
@@ -198,6 +198,9 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
                 emit AmountWithheld(target, targetWithheldAmount);
             }
 
+            // Update total to-be-deposited amount
+            totalAmount += amount;
```
kupermind, line 202: Add up to the total amount to-be-deposited

contracts/staking/DefaultTargetDispenserL2.sol
```diff
@@ -329,14 +332,31 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
...
+            // Check for overflow
+            if (totalAmount > localWithheldAmount) {
+                revert Overflow(totalAmount, localWithheldAmount);
+            }
+
+            // Update withheld amount
+            localWithheldAmount -= totalAmount;
+            withheldAmount = localWithheldAmount;
+
+            emit WithheldAmountUpdated(localWithheldAmount);
```
kupermind, line 358: Calculate total amount being deposited from the balance of this contract. If the balance is used that is recorded in `withheldAmount` value, then it's the exact amount we need to subtract from `withheldAmount`.

PR #261, opened 2026-03-03
Fix bbb

kupermind:
- Fix BBB contracts.


contracts/utils/BuyBackBurner.sol
```diff
@@ -104,38 +109,25 @@ abstract contract BuyBackBurner {
...
     mapping(address => address) public mapV2Oracles;
-    // Map of V3 pool address => whitelisted status
-    mapping(address => bool) public mapV3Pools;
 
     /// @dev BuyBackBurner constructor.
-    /// @param _liquidityManager LiquidityManager address.
     /// @param _bridge2Burner Bridge2Burner address.
     /// @param _treasury Treasury address.
-    /// @param _swapRouter Concentrated liquidity swap router address.
-    constructor(address _liquidityManager, address _bridge2Burner, address _treasury, address _swapRouter) {
```
kupermind, line 126: Remove V3 related functionality

contracts/utils/BuyBackBurner.sol
```diff
@@ -476,6 +317,11 @@ abstract contract BuyBackBurner {
     /// @dev Transfers specified token to treasury.
     /// @param token Token address.
     function transfer(address token) external {
+        // Check that token is not set for swapping into OLAS
+        if (mapV2Oracles[token] != address(0)) {
+            revert UnauthorizedToken(token);
+        }
```
kupermind, line 323: If second token is in added oracles - it will be swapped for OLAS, otherwise transferred.

contracts/utils/BuyBackBurner.sol
```diff
@@ -296,75 +227,28 @@ abstract contract BuyBackBurner {
...
+            if (secondTokens[i] == olas) {
+                revert UnauthorizedToken(secondTokens[i]);
+            }
 
-    /// @dev Checks pool prices via Uniswap V3 built-in oracle.
-    /// @notice This is a legacy function for compatibility with one of apps, it accounts for UniswapV3 only.
-    /// @param token0 Token0 address.
-    /// @param token1 Token1 address.
-    /// @param feeTier Fee tier.
-    function checkPoolPrices(address token0, address token1, address uniV3PositionManager, uint24 feeTier)
```
kupermind, line 352: This function is for V3 pools only and it needs to be corrected according to latest external audit findings.

77ph (approved):

contracts/utils/BuyBackBurner.sol
```diff
@@ -189,49 +163,6 @@ abstract contract BuyBackBurner {
...
         olasAmount = _performSwap(secondToken, secondTokenAmount, poolOracle);
-
-        // Get current pool price
-        uint256 tradePrice = IOracle(poolOracle).getPrice();
-
-        // Validate against slippage thresholds
-        uint256 lowerBound = (previousPrice * (100 - maxSlippage)) / 100;
-        uint256 upperBound = (previousPrice * (100 + maxSlippage)) / 100;
-
-        require(tradePrice >= lowerBound && tradePrice <= upperBound, "After swap slippage limit is breached");
```
mariapiamo, line 200: I think we still need to have post trade protection, while using maxSlippage as BPS

kupermind: The audit findings showed that it's counter beneficial, and having a post check is a vulnerability.

mariapiamo (approved):

PR #263, opened 2026-03-09
refactor: getTWAP() instead of validatePrice() and compare by price

kupermind:
- Fix TOCTOU vulnerability in BuyBackBurner swap slippage: Replace amountOutMin = 0 with TWAP-derived minimum output amounts. Previously, validatePrice() was a pre-swap gate
  check that didn't enforce anything atomically during the swap — the router accepted any output regardless. Now _buyOLAS() computes amountOutMin from getTWAP() and passes it
  directly to the router/vault.
  - Fix hardcoded minAmountsOut = 1 in LiquidityManager remove-liquidity: Both LiquidityManagerETH and LiquidityManagerOptimism used hardcoded minimum output amounts of 1 when
  removing V2 liquidity, making them vulnerable to sandwich attacks. Now they compute manipulation-resistant fair reserves using the constant product invariant (k = r0 * r1)
  combined with TWAP price and sqrt, then derive proper minAmountsOut with slippage bounds.
  - Add getTWAP() to oracles: Both UniswapPriceOracle and BalancerPriceOracle now expose a getTWAP() function returning the time-weighted average price in 1e18 format (OLAS per
  secondToken).
  - Remove validatePrice() from both oracles along with maxSlippageBps and MAX_BPS — no longer needed since slippage is now enforced atomically by callers.
  - Add Bridge2BurnerPolygon and Bridge2BurnerArbitrum contracts for bridging OLAS to L1 for burning on those chains.
  - Add fork tests for BuyBackBurnerBalancer on Polygon and Arbitrum, and update existing fork tests for oracle warmup and correct price direction.
  - Update deployment scripts to match new oracle constructor signatures (removed maxSlippageBps parameter), add maxStalenessSeconds to Balancer globals JSONs, and add
  Bridge2Burner deployment scripts.


contracts/oracles/UniswapPriceOracle.sol
```diff
@@ -190,25 +169,18 @@ contract UniswapPriceOracle {
 
         // TWAP history check
         if (elapsed < minTwapWindow || elapsed == 0) {
-            // Window too small: not enough history for TWAP
-            return false;
+            revert ZeroValue();
```
mariapiamo, line 196: Should we also check observations are enough?

contracts/oracles/BalancerPriceOracle.sol
```diff
@@ -239,14 +215,10 @@ contract BalancerPriceOracle {
...
+            revert ZeroValue();
         }
 
-        // Calculate price difference
-        uint256 diff = (spot > twap) ? (spot - twap) : (twap - spot);
-
-        // Compare as BPS: diff / twap <= slippageBps / MAX_BPS
-        // => diff * MAX_BPS <= twap * slippageBps
-        return diff * MAX_BPS <= twap * slippageBps;
+        return twap;
```
mariapiamo, line 221: Similar to uniswap, should we explicitly enforce a minimum number of observations?

contracts/oracles/UniswapPriceOracle.sol
mariapiamo: to be discussed with @77ph

PR #272, opened 2026-04-13
Audit response — C4R 2026-01 + internal-15 + internal-16

kupermind:
## Summary

Umbrella PR consolidating the V3-restoration arc and all subsequent C4R 2026-01 / internal-15 / internal-16 audit fixes that landed on top of it. Originally split across PRs #273, #275–#283 (all merged into this branch); collapsed here for a single review surface against `main`.

The arc starts with restoring the Uniswap V3 / Slipstream CL swap path (#272) so the price-guard fixes have a real attack surface to land on, then runs through three rounds of audit response (C4R 2026-01, internal-15, internal-16) plus deploy-pipeline ergonomics.

### #272 — Restore V3 swap path in BuyBackBurner

V3 / Slipstream CL swap code that was stripped from BBB in `ca9203a` is restored as **prep work** for the C4R `LiquidityManagerCore` price-guard fixes — those fixes need a V3 attack surface to land on. V2 audit fixes are preserved.

- 4-arg constructor `(liquidityManager, bridge2Burner, treasury, swapRouter)`
- `mapV3Pools` whitelist + `setV3PoolStatuses` (later reshaped — see #280)
- V3 overloads of `_performSwap`, `_buyOLAS`, `buyBack`
- `getV3Pool` virtual + Uniswap (`fee`) / Slipstream (`tickSpacing`) overrides
- `IRouterV3` / `ICLFactory` / `IFactory` / `ISwapRouter` interfaces
- State-variable order: `mapV2Oracles, mapV3Pools, mapTokenMaxSlippages` — proxy storage compatibility preserved (deprecated `maxSlippage` slot 4 untouched)

### #273 — `LiquidityManagerCore` V3 price-guard fixes (C4R #17/#18/#19)

- **#17 + #18 (HIGH)** — `checkPoolAndGetCenterPrice` had two issues with the same root cause: `centerSqrtPriceX96` (intended to be slot0 instantaneous) was overwritten by the oracle TWAP value before the deviation check ran. The check then compared TWAP vs TWAP and was unable to detect manipulation. Fixed by capturing slot0 once and treating TWAP as a separate variable.
- **#19 (MEDIUM)** — `changeRanges` silently routed value to treasury on failure. Now reverts.

### #275 — V3 `_performSwap` honours `mapTokenMaxSlippages` (H-02 / C4A M-11)

[...]

PR #278, opened 2026-04-22
feat: V3 path optional in BuyBackBurner (per-chain V3 enable/disable)

kupermind:
## Summary

Surfaced by the deployment audit on top of #272: the `BuyBackBurner` constructor required all four addresses non-zero, which blocked V2-only chains (gnosis, polygon, arbitrum) from deploying without first standing up a `LiquidityManager`. This PR makes `_liquidityManager` and `_swapRouter` optional and adds a `V3PathDisabled` runtime guard on every V3-touching surface.

- **Constructor**: relaxes zero check on `_liquidityManager` and `_swapRouter`; keeps it for `_bridge2Burner` and `_treasury`.
- **New error** `V3PathDisabled()` reverted by V3 surfaces when either V3 immutable is unset.
- **Two internal guards**: `_requireV3Enabled()` (both V3 immutables) and `_requireLiquidityManager()` (LM only — for `checkPoolPrices`, which doesn't read `swapRouter`).
- **Guarded surfaces**: `buyBack(V3 4-arg)`, `_buyOLAS(V3)`, `setV3PoolStatuses`, `checkPoolPrices`.
- **Ungated** (verified V3-independent): V2 `buyBack`, `setV2Oracles`, `setMaxSlippages` (V2 + V3 share `mapTokenMaxSlippages`), `changeOwner`, `transferToken`, `updateOraclePrice`, `changeImplementation`.
- **Deploy scripts**: `deploy_01_buy_back_burner_balancer.sh` and `deploy_02_buy_back_burner_uniswap.sh` normalise empty / null globals values for `liquidityManagerAddress` and `swapRouterV3Address` to `0x0…0` so `forge create` accepts them.
- **Backward compat**: existing on-chain proxies have non-zero LM / swapRouter (deployed under the old strict check), so V3 is enabled and the new contract is bytecode-equivalent for them. To enable V3 on a chain initially deployed with zeros, deploy a new impl with both addresses populated and call `changeImplementation`.

## Tests

- `test/BuyBackBurnerV3Disabled.t.sol` — **19 new unit tests** (all pass):
  - Constructor accepts zero LM / swapRouter / both; still reverts on zero bridge2Burner / treasury (Uniswap and Balancer variants).
  - `buyBack(V3)` reverts `V3PathDisabled` on each of LM-zero / swapRouter-zero / both-zero.
  - `setV3PoolStatuses` reverts `V3PathDisabled` on LM-zero / swapRouter-zero; owner check fires before the V3 guard.
  - `checkPoolPrices` reverts `V3PathDisabled` only when LM is zero; passes through with `swapRouter == 0`.
  - V2 admin surfaces (`setV2Oracles`, `setMaxSlippages`, `changeOwner`, `changeImplementation`) succeed on V3-disabled deployment.

## Test plan

- [x] `forge build` — clean
- [x] `forge test --mc BuyBackBurnerV3Disabled -vv` — 19/19
- [x] `forge test --mc 'BuyBackBurnerV3Swap|BuyBackBurnerV2OracleRefresh|LiquidityManagerCorePriceGuard|LowFindingsAudit15|NeighborhoodScannerPrecision'` — 31/31 (regression)
- [x] `forge test -f https://mainnet.gateway.tenderly.co --mc BuyBackBurnerUniswapETH -vv` — 21/21 (regression)
- [x] `forge test -f https://arbitrum.gateway.tenderly.co --mc BuyBackBurnerBalancerArbitrum -vv` — 19/19 (regression)

## Docs

- `docs/Vulnerabilities_list_tokenomics.md` — entry #21 added (rationale, guarded surfaces, upgrade path).
- `audits/internal15/FINAL_REVIEW.md` — section 7 records this post-cycle change for the audit trail.
- `scripts/deployment/pol/README.md` — note that POL is per-chain optional; V2-only chains can skip the entire `pol/` directory.

Targets `restore-v3-bbb` (PR #272), so this lands as part of the same merge into `main` once #272 is reviewed.


contracts/utils/BuyBackBurner.sol
```diff
@@ -161,6 +166,23 @@ abstract contract BuyBackBurner {
...
+    function _requireV3Enabled() internal view {
+        if (liquidityManager == address(0) || swapRouter == address(0)) {
+            revert V3PathDisabled();
+        }
+    }
+
+    /// @dev Reverts with V3PathDisabled when liquidityManager is zero. Used by surfaces that
+    ///      need the LiquidityManager but do not touch swapRouter (e.g. `checkPoolPrices`,
+    ///      a read-only diagnostic helper).
+    function _requireLiquidityManager() internal view {
```
kupermind: It's used only once, not worth a function

mariapiamo (commented):
-

mariapiamo (commented):
Potential deployment footgun in the new BBB deploy scripts:

Both scripts normalize missing `liquidityManagerProxyAddress` / `swapRouterV3Address` to `0x000...000`:
- scripts/deployment/utils/deploy_01_buy_back_burner_balancer.sh
- scripts/deployment/utils/deploy_02_buy_back_burner_uniswap.sh

Because `BuyBackBurner` now accepts zero LM/router in the constructor, a misconfigured V3-intended chain can silently deploy in V3-disabled mode and only fail later at runtime with `V3PathDisabled`.

Suggestion (fail-closed default):
1) Error out if either field is missing by default.
2) Allow zero-values only with explicit opt-in (e.g. `ALLOW_V3_DISABLED=true`) or a strict chain allowlist (e.g. gnosis).
3) Reject partial config (one zero, one non-zero).
4) During migration, optionally fall back to legacy `liquidityManagerAddress` with a clear warning.

kupermind:
Thanks @mariapiamo — you're right that silent zero-default on a V3-intended chain is a real footgun. Addressed in `fee94be` with a minimal policy rather than env-var ceremony:

- `liquidityManagerProxyAddress` and `swapRouterV3Address` must each be set explicitly in `globals_<chain>.json` — either to a real address (V3 enabled) or to `0x0000000000000000000000000000000000000000` (V3 disabled). Empty/null is refused with a clear message pointing at the offending field.
- Gnosis globals already carry explicit zero addresses for both fields, so it continues to deploy V3-disabled as before.
- Every other chain currently has empty strings for these fields and will now fail the check until the operator makes the V3-enabled / V3-disabled choice explicit per chain. No accidental silent V3-disabled on a chain that was meant to be V3-enabled.
- No env flags, no mixed-config branching, no operator ceremony. The globals file is the single source of truth.

I tried a more elaborate policy first (`2e0bad2`: `V3_DISABLED=true` opt-in + mixed-config rejection) but that added friction and state to the deploy flow for little benefit over "just require the fields". Dropped in favour of the simple form above.

Let me know if this resolves the concern or if you'd like a tighter check somewhere.

mariapiamo (approved):

mariapiamo (commented):
[P2] V3 buyback tokens remain sweepable to treasury

transfer() still only protects tokens that have a V2 oracle in mapV2Oracles. A V3-only buyback input token is authorized through mapV3Pools, so it can have no V2 oracle and still be intended for BBB swaps; any caller can call transfer(secondToken) first and move the accumulated input token to treasury instead of letting BBB swap it into OLAS. PR #278 does not add V3 second-token tracking or a guard here

PR #286, opened 2026-05-12
deploy(pol): tune LM cardinality + slippage; drop dead v3PoolStatuses

kupermind:
## Summary

Per-chain tuning of the POL `LiquidityManager` deploy params in `scripts/deployment/pol/globals_<chain>_mainnet.json`, plus removal of an obsolete field and a new fork test that empirically backs the cardinality choice.

| Field | ETH | Base | Optimism |
|---|---|---|---|
| `observationCardinality` | 60 *(unchanged)* | **60 → 120** | **60 → 120** |
| `liquidityManagerMaxSlippage` | **500 → 1000** | **500 → 1000** | **500 → 1000** |
| `v3PoolStatuses` | *removed* | *removed* | *removed* |

### `observationCardinality` — sized by block time

Fed once to `IUniswapV3.increaseObservationCardinalityNext` per fresh pool inside `convertToV3` (`contracts/pol/LiquidityManagerCore.sol:733`). Sized so the buffer covers a reasonable worst-case TWAP window without making `convertToV3` OOG:

- ETH (12s blocks): `60 × 12s = 720s` nominal worst-case coverage.
- Base / Optimism (2s blocks): `120 × 2s = 240s` nominal worst-case coverage.

Matching `SECONDS_AGO = 1800s` directly (e.g. `cardinality ≈ 1024` on 2s-block L2s) is rejected on gas. Empirical measurement on an ETH fork (new test, see below):

| `observationCardinality` | `increaseObservationCardinalityNext` gas |
|---|---|
| 60 | 1,322,701 (~1.32M) |
| 120 | 2,656,965 (~2.66M) |
| 1024 | **22,760,133 (~22.76M)** — overruns realistic L1 tx budget once V3 mint (~500k–1M) and LM accounting are layered on |

60 / 120 are safe in practice because:
1. A V3 observation is written at most once per block → for sparsely-traded OLAS pools, 60–120 slots cover hours of real time.
2. `checkPoolAndGetCenterPrice` falls back to slot0 for `cardinality <= 1` with the deviation guard still in force (`LiquidityManagerCore.sol:1145-1153`; covered by `internal15` M-01 / `internal16` note).
3. `increaseObservationCardinalityNext` is permissionless — any specific pool can be bumped post-deploy via `cast` without an LM redeploy. The constructor immutable is just the baseline.

### `liquidityManagerMaxSlippage` 500 → 1000 (5% → 10%)

Seeded via `LiquidityManagerCore.initialize(uint16)` from the proxy constructor, used in `_optimizeTicksAndMintPosition`, `_increaseLiquidity`, `_decreaseLiquidity` to compute `amount{0,1}Min` for the V3 position manager.

Aligned with the contract's `MAX_ALLOWED_DEVIATION` (10%, `LiquidityManagerCore.sol:122`), the slot0-vs-TWAP deviation guard. With both at 10%, the position-manager `amount0Min` / `amount1Min` guard does not reject mints that the upstream deviation guard has already accepted. Updatable post-deploy via `changeMaxSlippage(uint16)`.

### `v3PoolStatuses` removed

Dead config left over from an earlier `setV3PoolStatuses` signature. The current on-chain setter is `BuyBackBurner.setV3Pools(address[] secondTokens, address[] pools)` (`contracts/utils/BuyBackBurner.sol:393`) — two arrays only — and `script_03_buy_back_burner_wire_v3.sh` already never read this field.

### Docs

- `docs/bbb_update_options.md`: new `## LM deployment parameter rationale` section.
- `scripts/deployment/pol/README.md`: expanded `observationCardinality` / `liquidityManagerMaxSlippage` rows; dropped `v3PoolStatuses` from the wire-step comment and the table.
- `CLAUDE.md`: listed the new fork test.

## Test plan

- [x] `forge build` clean.
- [x] New fork test passes against a public ETH RPC:
  ```
  forge test -f https://ethereum-rpc.publicnode.com --mc LiquidityManagerObservationCardinalityGasETH -vv
  ```
  Three tests, three measured gas points matching the table above.
- [x] `gitleaks detect` on the changed JSON files: no new findings.
- [ ] Sanity-run existing LM fork tests (`LiquidityManagerETH`, `LiquidityManagerBase`) on a reviewer's chosen RPCs to confirm the deploy-param edits don't ripple into them (they shouldn't — these test files hardcode their own cardinality / slippage constants).


77ph (commented):
# Review — PR #286 «deploy(pol): tune LM cardinality + slippage; drop dead v3PoolStatuses»

**Reviewed against**: `origin/main` → `pr-286` (`c19caa9`) — 1 commit, 7 files, +168/-13. **Context**: per-chain tuning of POL `LiquidityManager` deploy parameters (`scripts/deployment/pol/globals_<chain>_mainnet.json`) + removal of obsolete `v3PoolStatuses` field + new fork test backing the cardinality choice empirically. **Methodology**: applied audit playbook v2.61 (Rule 43 «state-mutation propagation», Pattern 14.14 «external dependency on critical path») + verification that the per-chain numbers are mutually consistent + sanity-check that the test methodology actually proves what the PR description claims.

**Verdict**: ✅ **Approve**. Numbers are coherent, removed field is genuinely dead, new test methodology is sound. 4 INFO observations — none blocking.

## Numerical sanity check

### `observationCardinality`: ETH `60` (unchanged), Base/Optimism `60 → 120`

Block-time math from PR body verified:

| Chain | Block time | Cardinality | Nominal worst-case coverage |
|---|---|---|---|
| ETH | ~12s | 60 | `60 × 12 = 720s` |
| Base | ~2s | 120 | `120 × 2 = 240s` |
| Optimism | ~2s | 120 | `120 × 2 = 240s` |

**`SECONDS_AGO = 1800s`** is set in both `contracts/pol/LiquidityManagerCore.sol:124` and `contracts/utils/BuyBackBurner.sol:113`. Nominal coverage on all three chains is **shorter than `SECONDS_AGO`** — meaning a pool that sees activity in EVERY block during a 1800s window will have `oldestObservation < block.timestamp - SECONDS_AGO` and `observe([1800])` would revert with `OLD`.

The PR's safety argument rests on the «sparsely-traded OLAS pools» assumption: in practice an observation is written **at most once per block**, but only if there's a swap that block. For a pool with one swap per minute, 60 observations cover ~60 minutes = 3600s > 1800s. The assumption holds for the current OLAS pool activity profile but is **load-dependent**, not load-independent — see INFO-1 below.

### `liquidityManagerMaxSlippage`: `500 → 1000` (5% → 10%) on all 3 chains

Cross-check against `LiquidityManagerCore.sol:122`:
```solidity
uint256 public constant MAX_ALLOWED_DEVIATION = 1e17;  // 10%
```

PR's design intent is **intentional alignment** of slippage and deviation guard at 10%, so that the V3 NPM `amount0Min`/`amount1Min` floor (the slippage check) doesn't reject mints that the upstream slot0-vs-TWAP deviation check has already accepted. This is a defensible design choice: the two guards operate at different layers but on the same percentage tolerance.

Edge-case I considered: between `checkPoolAndGetCenterPrice` returning a snapshot and the V3 mint, can slot0 shift further within the same transaction (intra-tx reentrancy, composed sub-calls) such that combined slot0-vs-TWAP movement exceeds 10% but each individual check passes? Atomic checks within a single call keep this bounded; sandwich attacks across multiple blocks don't apply here (single-tx scope). **OK.**

### `v3PoolStatuses` removal

Verified against on-chain setter signature:
```solidity
// contracts/utils/BuyBackBurner.sol:393
function setV3Pools(address[] memory secondTokens, address[] memory pools) external virtual {
```

Two arrays, no statuses. The deployment script `script_03_buy_back_burner_wire_v3.sh` already never reads `v3PoolStatuses` (PR author claim). Field was dead config — removal is correct cleanup.

## New fork test methodology

`test/LiquidityManagerObservationCardinalityGasETH.t.sol` empirically establishes the gas cost of `increaseObservationCardinalityNext(N)`:

| Cardinality | Gas | Note |
|---|---|---|
| 60 | ~1.32M | Production ETH choice |
| 120 | ~2.66M | Production Base/Optimism choice |
| 1024 | ~22.76M | Reference: would cover full 1800s SECONDS_AGO on 2s-block L2s |

Test methodology:
- ✅ Fresh ERC20 mock tokens to guarantee `(cardinality, cardinalityNext) = (1, 1)` start state — isolates the bump cost.
- ✅ `gasleft()` measurement around the single `increaseObservationCardinalityNext` call.
- ✅ Three independent pools at distinct fee tiers (100, 500, 3000) so cardinality bumps don't share storage warmth.
- ✅ Post-condition assertion `cardinalityNext == target` verifies the call took effect, not just gas-consumed.

The measured `~22.76M` for cardinality=1024 is the load-bearing claim: **it would not leave room for the actual V3 mint (~500k-1M gas) on top, in a 30M ETH block cap with single-tx ceilings ~25M**. Test proves this empirically rather than by appeal to a constant. Good.

## Findings

### INFO-1 (Operational) — Cardinality coverage is load-dependent; need a monitoring trigger

The `60 × block_time` (ETH) or `120 × block_time` (Base/Optimism) nominal coverage of 720s/240s is **shorter than `SECONDS_AGO=1800s`**. The safety argument depends on the OLAS pool being sparsely-traded enough that `cardinality` slots span >1800s of real time.

**Failure mode if the assumption breaks**: pool gets enough trading to fill the buffer within <1800s (e.g., post-launch interest surge, MEV bot activity, deep DEX-aggregator routing through OLAS). Then:

1. `LiquidityManagerCore.sol:1137` first-line defense — if `oldestTimestamp + SECONDS_AGO < block.timestamp` → return slot0 with no deviation check. Triggers for **dormant** pools only.
2. Otherwise `getTwapFromOracle` calls `observe([1800])` → reverts with `OLD` error.
3. `LiquidityManagerCore.sol:1145-1153` second-line: `if (!success || returnData.length == 0)` AND cardinality > 1 → `revert ObservationFailed(pool)`. **This is the critical path.**

Net behavior under heavy activity: `convertToV3` / `buyBack` reverts. Service degradation, **not fund loss**. Mitigation is permissionless `increaseObservationCardinalityNext` — anyone can bump cardinality on the affected pool from `cast` post-deploy.

**Suggested follow-up** (not blocking):
- Add a monitoring metric `(oldestObservationAge / SECONDS_AGO)` per active V3 pool. Alert when ratio drops below e.g. 1.5×. This gives lead time to bump cardinality before the pool actually crosses into `ObservationFailed` territory.
- Document in `docs/bbb_update_options.md` the operational procedure: detect via Tenderly or RPC poll → run `cast send <pool> "increaseObservationCardinalityNext(uint16)" <new_value>` from the Treasury multisig (or anyone with gas budget — it's permissionless).
- Optional: add a fork test that fills the cardinality=60 buffer with high-frequency mock swaps and asserts `checkPoolAndGetCenterPrice` reverts with `ObservationFailed` — proves the failure mode is correctly attributed to cardinality wrap, not to some other bug.

This is **Pattern 14.14 sister** (external dependency on critical path) — the V3 pool's observation buffer is an external resource on the protocol's hot path. Mitigation lever (`increaseObservationCardinalityNext`) is permissionless, which is the right design — but it requires operational awareness, not just code correctness.

### INFO-2 — Slippage/deviation alignment claim deserves an explicit invariant comment

PR text says «with both at 10%, the V3 `amount0Min`/`amount1Min` guard does not reject mints that the upstream slot0-vs-TWAP deviation check has already accepted». This is correct but **not obvious from the code alone**. Anyone reading `LiquidityManagerCore.sol:122` and `liquidityManagerMaxSlippage = 1000` separately won't see why the two numbers must match.

Cosmetic suggestion (separate PR or end of this one): add a comment near `MAX_ALLOWED_DEVIATION`:
```solidity
/// @dev Pinned at 10%. Deploy parameter `liquidityManagerMaxSlippage` should be set
///      to the same percentage so the V3 NPM slippage guard doesn't reject mints
///      that this deviation pre-check has already accepted.
uint256 public constant MAX_ALLOWED_DEVIATION = 1e17;
```

And in the README/docs row for `liquidityManagerMaxSlippage`, link explicitly to `MAX_ALLOWED_DEVIATION` (you already added the line number — good).

### INFO-3 — `v3PoolStatuses` removal is correct; check that no off-repo consumer reads it

Verified within the repo: no consumer of `globals_*.json` reads `v3PoolStatuses` (`grep` shows zero hits in `scripts/deployment/`). External operator tooling (off-repo runbooks, custom scripts, Tanya frontend, etc.) might still reference this field if they were written against an earlier signature. Worth a one-line announcement in the next operator changelog: «`v3PoolStatuses` removed from `globals_*.json` — no action needed because the field was already unused at runtime, but external tooling should not expect to find it».

### INFO-4 — Test scope is gas only, not the operational revert path

`LiquidityManagerObservationCardinalityGasETH` measures `increaseObservationCardinalityNext` gas — that's what the PR claims to test, and it does. But it does **not** exercise the operational risk that motivates this PR's deploy-param choice: a heavily-traded pool wrapping the cardinality buffer within `SECONDS_AGO` and `checkPoolAndGetCenterPrice` reverting.

A complementary test (not necessarily this PR) could:
1. Deploy a pool with `cardinality = 60`.
2. Fill it with 60 mock swaps spaced 1s apart (so oldestObservation is 60s old).
3. Call `LiquidityManagerCore.checkPoolAndGetCenterPrice(pool)`.
4. Expect `revert ObservationFailed(pool)`.

This would empirically validate INFO-1's mitigation reasoning (that the failure mode is `revert`, not silent return) and proves the operator-runbook bump path is actually needed. Not blocking; a useful invariant test for the next test-suite expansion.

## Scope of fix vs prior reviews

| Item | Severity | Status |
|---|---|---|
| Cardinality tuning (Base/Optimism 60→120) | Deploy param | ✅ Coherent with block-time math |
| Slippage tuning (all 3 chains 500→1000) | Deploy param | ✅ Aligned with `MAX_ALLOWED_DEVIATION=10%` |
| `v3PoolStatuses` removal | Cleanup | ✅ Field was dead; `setV3Pools` signature has 2 args only |
| New fork test (gas measurement) | Test addition | ✅ Methodology sound; proves cardinality=1024 OOG claim empirically |

No regression against prior `internal15` / `internal16` audit findings (M-01 slot0-fallback for fresh pools is preserved; `LiquidityManagerCore.sol:1145-1153` deviation guard unchanged).

## What I deliberately did NOT find

- No new external network calls. No new secrets handling. No new on-chain access surface.
- No security primitive change (no crypto/TEE/AA surface) → Rule 42 not triggered.
- No new state-mutation patterns that Rule 43 would flag — the changes are deploy-param JSON edits + docs + a gas test.
- Pattern 14.13 (governance parameter change blocks based on past state): N/A. `liquidityManagerMaxSlippage` change applies to new mint operations; updatable post-deploy via `changeMaxSlippage()` per PR docs row.
- Pattern 1.X (oracle/price manipulation): the slot0 fallback for dormant pools (`LiquidityManagerCore.sol:1137`) is a **pre-existing** path; this PR doesn't introduce new oracle attack surface. The slot0 fallback path was reviewed in `internal15` M-01 / `internal16`.

## Cross-cycle dispatch from playbook v2.61

| Pattern | Applicability | Outcome |
|---|---|---|
| **14.14** (external upgradeable dep on access-control path — TOB-KILGND-1) | Pool observation buffer = external resource on hot path | ✅ permissionless mitigation lever exists; INFO-1 flags operational monitoring need |
| **11.29** (validate-then-settle gap — TOB-AAVE-7 / Rhea anchors) | Deviation check vs V3 mint slippage | ✅ both at 10%, no gap |
| **3.106** (pre-condition validates input not post-mutation state) | N/A — no mutation in this PR | ✓ pass-by-N/A |
| **5.x** (cryptographic patterns) | N/A — no crypto surface | ✓ pass-by-N/A |

## Summary

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 0 |
| **Info** | **4** (INFO-1 operational monitoring trigger; INFO-2 slippage/deviation alignment comment; INFO-3 off-repo consumer announcement; INFO-4 functional revert-path test) |

PR #286 is a focused, well-scoped deploy-param tuning with empirical backing for the cardinality choice. Numbers are coherent across chains, the `MAX_ALLOWED_DEVIATION` / `liquidityManagerMaxSlippage` alignment is intentional design, and the `v3PoolStatuses` removal is correct dead-code cleanup. The new fork test proves the gas-cost claim that motivates the deploy-param values.

**Recommend approve and merge.** INFO-1 (monitoring trigger for cardinality coverage) is the most worthwhile follow-up — a low-cost addition to the operator runbook that would prevent service-degradation incidents under future activity surges.

77ph (dismissed):

mariapiamo (commented):
The parameter choices look sensible as gas-conscious deployment defaults, but I think the `observationCardinality` rationale should qualify the safety claim.

`60` on ETH and `120` on Base/Optimism do not guarantee `SECONDS_AGO = 1800` TWAP availability under high pool activity. They only set the pool’s observation capacity baseline. The actual retained time window depends on how often the pool writes observations.

For example, on Base/Optimism:
- `120` slots can cover much more than 1800s if the OLAS pool is sparse.
- But if the pool writes once per 2s block, `120 * 2s = 240s`, below the 1800s TWAP window.
- Full worst-case 1800s coverage on a 2s chain would be about `900` slots, which the gas test shows is expensive.

So I would phrase these values as “reasonable low-gas baselines” rather than generally “safe”. The deployment/ops note should also mention that active pools may need a post-deploy `increaseObservationCardinalityNext(<higher N>)` bump, followed by enough pool activity/time to fill the new slots.

The rest of the PR direction makes sense to me:
- `liquidityManagerMaxSlippage = 1000` aligns with `MAX_ALLOWED_DEVIATION = 10%`.
- Removing `v3PoolStatuses` looks correct because the current setter is `setV3Pools(secondTokens, pools)`.
- The fork gas test is useful evidence for rejecting very high default cardinality.

77ph:
Follow-up to the cardinality discussion: I dug into production values used by live Uniswap V3 pools on ETH mainnet (12s blocks) to anchor the choice in industry precedent rather than abstract math.

**Reference: `observationCardinality` in production V3 pools** (source: [Alex Roan gas test gist](https://gist.github.com/alexroan/71b38d387ed2a86bf3abdf3acd0f8415), measured on mainnet):

| Pool | observationCardinality |
|---|---|
| ETH/USDC 0.05% | 720 |
| ETH/USDC 0.3% | 1440 |
| DAI/ETH | 300 |
| UNI/ETH | 300 |
| LINK/ETH | 144 |
| OHM/USDC, stETH/ETH | 1 (default — never bumped) |

The Uniswap V3 docs don't recommend a specific number; they only describe the trade-off. The production wisdom that emerges from these data points:
- High-volume pairs (ETH/USDC): 720–1440
- Mid-volume pairs (LINK/ETH, DAI/ETH, UNI/ETH): 144–300
- Low-volume / never-bumped pools: 1 (default)

`120` does not appear as a canonical baseline in any of the production references I checked. The closest live precedent for a mid-volume pair like OLAS/WETH is LINK/ETH at **144** (which gives `144 × 12s = 1728s ≈ 96% of SECONDS_AGO=1800s` even under the worst case of one observation per block).

### Proposed compromise

Anchor the deploy values to mid-volume production precedent and verify they fit within the new 16M tx gas limit:

| Chain | Current | Proposed | Anchor | Bump gas (~22.2k/slot) | Total `convertToV3` with mint + LM logic |
|---|---|---|---|---|---|
| ETH (12s) | 60 | **144** | LINK/ETH | ~3.2M | ~4.7M — fits 16M comfortably |
| Base (2s) | 120 | **300** | DAI/ETH, UNI/ETH | ~6.7M | ~8.2M — fits 16M with margin |
| Optimism (2s) | 120 | **300** | same as Base | ~6.7M | ~8.2M — same |

Coverage under various activity scenarios:

| Scenario | ETH @ 144 | Base/Op @ 300 |
|---|---|---|
| Worst case (1 obs per block) | 1728s ≈ window | 600s = 33% of window |
| Realistic surge (1 obs / 20s) | 2880s ✓ | 6000s ✓ |
| Typical OLAS pool (1 obs / minute or rarer) | 144 min ✓ | 5+ hours ✓ |

For Base/Optimism, full worst-case (1 obs per block for the entire 1800s) coverage would require 900 slots, which costs ~20M gas just for the bump and exceeds the 16M tx ceiling once mint + LM accounting are added. So sustained every-block activity on a 2s chain remains a residual operational scenario regardless of deploy value — but at 300 vs 120 the safety envelope is 5× wider against realistic surges. Below this surge regime, both the deviation-guard fallback and `increaseObservationCardinalityNext` remain available as defense.

The `liquidityManagerMaxSlippage = 1000` and `v3PoolStatuses` removal don't change.

77ph (approved):
Commit `d1d682e` cleanly addresses both INFO-1 and INFO-2:

**INFO-1 (cardinality)** — values re-anchored to mid-volume V3 production precedent (ETH 120, Base/Op 300), gas re-measured empirically (300 = 6.66M, comfortably under 16M L2 ceiling), and the docs now carry an executable operational runbook for the cardinality-wrap failure mode (detection signal, permissionless `increaseObservationCardinalityNext` mitigation, recommended `300 → 600 → 1200` ramp). The honest framing in `bbb_update_options.md` ("mid-volume V3 pool precedent" rather than abstract "safe") matches the reality of the trade-off.

**INFO-2 (deviation/slippage co-existence)** — both `MAX_ALLOWED_DEVIATION` and `maxSlippage` now carry paired NatSpec explaining they operate in different units (price space vs amount space), at different points in the call lifecycle (pre-flight pool-sanity vs post-flight execution gate), and with different governance scopes (tamper-evident immutable vs operator-tunable). This is stronger than my original "add a comment" suggestion — it clarifies the two guards are not the same check at the same percentage, but distinct checks whose coherence is achieved by aligning the percentage as a heuristic, not a substitution.

INFO-3 and INFO-4 are out of this PR's scope (off-repo tooling announcement, functional revert-path test) and remain reasonable forward suggestions.

LGTM on the deploy params, NatSpec, docs, and runbook.

77ph (approved):

mariapiamo:
Oracle config check, first pass.

I compared the PR #286 oracle globals against:

- https://olas.network/olas-token#token-details
- https://github.com/valory-xyz/autonolas-tokenomics/blob/main/docs/lp_token_bridging.md

Reusable local check output:

- `oracle_param_source_check_pr286.json`

Results:

- OLAS token addresses match for ETH, Gnosis, Polygon, Base, Optimism, and Arbitrum.
- ETH Uniswap pair matches `lp_token_bridging.md`.
- Balancer pool IDs match `lp_token_bridging.md` for Gnosis, Polygon, Base, Optimism, and Arbitrum.
- `TWAP/update/staleness = 900/900/86400` is not validated by either of those sources. Those appear to be deployment policy params from globals.
- Balancer vault `0xBA1222...C8` is also not validated by those two sources; `lp_token_bridging.md` validates pool IDs / LP addresses.

Base nuance:

`lp_token_bridging.md` has two Base Balancer pools:

- OLAS-USDC: `0x5332584890d6e415a6dc910254d6430b8aab7e69000200000000000000000103`
- WETH-OLAS: `0x2da6e67c45af2aaa539294d9fa27ea50ce4e2c5f0002000000000000000001a3`

The PR oracle params use the Base WETH-OLAS pool, and that matches the second Base row in `lp_token_bridging.md`. Which is also the one with larger TVL.

However, the OLAS token details page's Base "Balancer" get-OLAS link points to the OLAS-USDC pool, not the WETH-OLAS pool. So the Base oracle config is correct if the intended oracle pool is the bridging-doc WETH-OLAS pool, but not if the intended source is the token-details page's first Base Balancer link.

mariapiamo:
Bridge2Burner config check, first pass.

I compared the PR #286 Bridge2Burner globals/scripts against:

- https://olas.network/olas-token#token-details
- https://github.com/valory-xyz/autonolas-tokenomics/blob/main/docs/lp_token_bridging.md
- https://docs.gnosischain.com/bridges/About%20Token%20Bridges/omnibridge
- https://specs.optimism.io/protocol/predeploys.html
- https://arbiscan.io/address/0x5288c571Fd7aD117beA99bF60FE0846C4E84F933
- https://docs.celo.org/cel2/notices/l2-migration

Reusable local check output:

- `bridge2burner_config_check_pr286.json`

Results:

- OLAS constructor token addresses match the token-details page for all checked chains.
- Gnosis looks consistent: `Bridge2BurnerGnosis`, OLAS `0xcE11...9d9f`, Omnibridge `0xf6A78083...CE2e268d`.
- Optimism/Base use `Bridge2BurnerOptimism` with OP Stack `L2StandardBridge` `0x4200000000000000000000000000000000000010`; this matches OP Stack docs.
- Arbitrum uses `Bridge2BurnerArbitrum` with L2 Gateway Router `0x5288c571...84F933` and L1 OLAS `0x0001...45CB0`; this matches Arbiscan/token details.
- Polygon is special: `Bridge2BurnerPolygon` does not call a bridge relayer. It transfers OLAS to `bridgeMediatorAddress = 0x9338...55fD` because the contract comments say Polygon PoS withdrawal has no explicit L1 recipient.
- Celo is the main unresolved item. Config uses OP Stack `L2StandardBridge` `0x4200...0010`, and Celo is now OP Stack L2, but OLAS token details and `lp_token_bridging.md` still point to Wormhole/Portal. PR docs also say Celo Bridge2Burner is forward-looking while whOLAS/native OLAS-CELO reconfiguration settles.

Interpretation:

- I would mark Gnosis, Optimism, Base, and Arbitrum as source-consistent.
- Polygon appears intentional from the contract comments, but it is not validated by the two public Olas references checked here. Checked that 0x9338b5153AE39BB89f50468E608eD9d764B755fD is the correct one reported in https://github.com/valory-xyz/autonolas-governance/blob/main/docs/configuration.json. 
- Celo should be explicitly flagged as forward-looking / not proven for current OLAS token bridge compatibility by the checked sources.

mariapiamo:
Reviewed the BBB implementation params and the corresponding proxy params.

Summary:
- Static script/contract shape looks correct for the BBB deployments.
- `BuyBackBurnerUniswap` and `BuyBackBurnerBalancer` both use the expected constructor shape:
  `constructor(address _liquidityManager, address _bridge2Burner, address _treasury, address _swapRouter)`.
- V3-enabled chains use the LM proxy output from `pol/deploy_03_liquidity_manager_proxy.sh` plus the nonzero V3 swap router:
  ETH, Base, Optimism.
- V3-disabled chains correctly use zero LM / zero V3 router:
  Gnosis, Polygon, Arbitrum.
- Bridge / treasury params match the intended per-chain sources:
  ETH uses `burnerAddress` + `timelockAddress`; L2s use `bridge2BurnerAddress` + `bridgeMediatorAddress`.

On-chain proxy checks:
- ETH proxy owner and current implementation match globals.
- Gnosis proxy owner and current implementation match globals.
- Optimism proxy owner and current implementation match globals.
- Polygon proxy owner and current implementation match globals.
- Arbitrum proxy owner and current implementation match globals.
- Base current implementation matches globals, but the owner is the legacy agents.fun address:
  `0x6F7a4938AB3bbF69480E7C109Af778ee78099Be7`.

Only operational caveat:
- Base `script_01_buy_back_burner_change_implementation.sh base_mainnet` will not succeed with the current standard derivation path `m/44'/60'/2'/0/0`.
- Either transfer Base BBB proxy ownership to the standard Autonolas deployer first, or run that single Base implementation-change tx with derivation path `m/44'/60'/9'/0/0`, then revert the local globals change.

Conclusion: BBB params and proxy targets look correct; Base owner handling is the one deployment-step caveat to account for.

mariapiamo:
Reviewed the LiquidityManager deployment config on current PR head `3c7e7ef`.

Scope checked:
- `scripts/deployment/pol/globals_{eth,base,optimism}_mainnet.json`
- LM deploy scripts consuming those globals
- corresponding `utils`, `oracles`, `staking`, proposal, and audit-owner configs
- live read-only checks for configured V3 position managers

Summary:
- LM constructor/config values are consistent with repo config and contract expectations.
- `liquidityManagerMaxSlippage = 1000` is valid on all three LM chains.
- One deployability issue: Optimism `networkURL` currently points to `https://0xrpc.io/op`, which returned HTTP 404 in a live `cast chain-id` check. `https://1rpc.io/op`, already used in staking Optimism globals, returned chain id `10`.

Treasury / governing executor:
- ETH correctly uses L1 timelock `0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE`.
  `deploy_02_liquidity_manager_eth.sh` reads `timelockAddress`, and this matches main/staking/utils configs plus audit owner records.
- Base correctly uses L2 governing bridge mediator `0xE49CB081e8d96920C38aA7AB90cb0294ab4Bc8EA`.
  This matches staking Base globals, utils globals, `proposal_19_migrate_l2_dispenser_base.js`, and `ownable_owners.csv`.
- Optimism correctly uses L2 governing bridge mediator `0x87c511c8aE3fAF0063b3F3CF9C6ab96c4AA5C60c`.
  This matches staking Optimism globals, utils globals, `proposal_18_migrate_l2_dispenser_optimism.js`, and `ownable_owners.csv`.

Position managers:
- ETH `0xC36442b4a4522E871399CD717aBDD847Ab11FE88`: valid Uniswap V3 `NonfungiblePositionManager`; live `factory()` returned `0x1F98431c8aD98523631AE4a59f267346ea31F984`.
- Base `0x827922686190790b37229fd06084350E74485b72`: valid Aerodrome Slipstream position manager; live `factory()` returned `0x5e7BB104d84c7CB9B682AaC2F3d509f5F406809A`.
- Optimism `0xf7f8ccce99Ca2896eC75D3A399D152dB96808399`: valid Velodrome Slipstream position manager; live `factory()` returned `0xe13Dd1fbA721Aa81a1826D9523AC9BC7d260c879`.

Remaining fields:
- OLAS addresses match repo main/staking configs.
- ETH `routerV2Address` matches Uniswap V2 Router02.
- Base/Optimism `balancerVaultAddress` matches Balancer Vault.
- Oracle addresses match `scripts/deployment/oracles/globals_*`; these are expected to be overwritten by the oracle deploy steps.
- `observationCardinality`: ETH `120`, Base/Optimism `300`, matching the PR docs.
- deploy-output fields are correctly empty before running deploy steps: `neighborhoodScannerAddress`, `liquidityManagerAddress`, `liquidityManagerProxyAddress`, and L2 `bridge2BurnerAddress`.

Slippage:
- `liquidityManagerMaxSlippage = 1000` means 10% in bps.
- It satisfies `0 < value <= MAX_BPS` (`MAX_BPS = 10_000`).
- It also matches `MAX_ALLOWED_DEVIATION = 1e17`, i.e. 10%, as documented in `LiquidityManagerCore`.
- Separate BBB V3 slippage arrays are still pending: `v3SecondTokens`, `v3Pools`, and `v3MaxSlippages` are empty. That is fine before V3 wiring, but `script_03_buy_back_burner_wire_v3.sh` must not be run until those arrays are populated with matching lengths and non-zero slippage values.

mariapiamo (approved):

PR #287, opened 2026-05-18
chore: deployment of oracle and BBB related contracts

DavidMinarsch:
## Summary

Deploys the new oracle / BuyBackBurner / LiquidityManager / Bridge2Burner stack to all 6 chains (mainnet, gnosis, polygon, base, optimism, arbitrum), extends the audit script to verify the post-deploy wiring end-to-end, and adds ownership-transfer scripts for the new proxies.

## What's in here

**Deployment (per-chain JSON address bumps)**
- `scripts/deployment/{oracles,pol,utils}/globals_<chain>_mainnet.json` — addresses for every newly deployed contract, including BBB impl bumps and the matching proxy state.
- `scripts/deployment/globals_mainnet.json` — `tokenomicsAddress` consolidated to the candidate impl (`0xaeec…d5e9`); unused `tokenomicsTwo/ThreeAddress` removed.

**Audit (`scripts/audit_chains/audit_contracts_setup.js`)**
- 8 new check functions: `checkUniswapPriceOracle`, `checkBalancerPriceOracle`, `checkNeighborhoodScanner`, `checkBridge2Burner` (+ Arbitrum variant for the extra `l1Olas` immutable), `checkBuyBackBurnerImpl`, `checkBuyBackBurnerProxy`, `checkLiquidityManagerImpl`, `checkLiquidityManagerProxy`.
- Verifies each BBB / LM proxy's EIP-1822 storage slot points at the expected implementation, owner is non-zero, and immutables match per-chain globals.
- Loads per-chain `{oracles,pol,utils}/globals_*.json` and falls back to public RPCs when `ALCHEMY_API_KEY_{MAINNET,MATIC}` are absent.
- `TokenomicsProxy` slot check now reads `tokenomicsAddress` (was `tokenomicsFourAddress`); will assert green once the DAO vote upgrades the proxy.
- Celo / Mode `TargetDispenserL2` wrapped in `try/catch` — pre-existing Wormhole→OP-stack migration left those entries renamed.
- `scripts/audit_chains/ownable_owners.csv` regenerated: 20 rows. All Tokenomics/Treasury/Dispenser/Depository/TargetDispensers show `ownerCategory=dao_executor`; freshly deployed BBB/LM proxies show `autonolas_deployer` (worklist for the transfer scripts below). Base BBB proxy reports `other` — owned by the legacy agents.fun deployer `0x6F7a…9Be7` (m/44'/60'/9'/0/0).

**ABIs** — `abis/0.8.30/`
- 13 new ABIs (Oracle / BBB impl + proxy / LM impl + proxy / Scanner / Bridge2Burner variants).
- Refreshed `Tokenomics.json` to match the new candidate implementation (19,602 bytes; was 24,549 bytes from the prior impl).

**Ownership-transfer scripts** (mirror `script_01_buy_back_burner_change_implementation.sh`)
- `scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh <network>`
- `scripts/deployment/pol/script_04_liquidity_manager_change_owner.sh <network>`
- Target owner = `bridgeMediatorAddress` (L2) or `timelockAddress` (L1), same convention as the deploy scripts' `treasury` argument.
- Pre-flight `owner()` check fails fast with a clear message if the signer's derivation path doesn't match the current proxy owner — covers the Base BBB legacy-owner case (header notes how to flip `derivationPath` to `m/44'/60'/9'/0/0` temporarily).
- No-op if already owned by the target.

## Roll-out

1. Merge this PR.
2. Run the transfer scripts (CSV is the worklist — every row with `ownershipChangeRequired=yes`):
   ```bash
   ./scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh eth_mainnet
   ./scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh gnosis_mainnet
   ./scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh polygon_mainnet
   ./scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh arbitrum_mainnet
   ./scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh optimism_mainnet
   # base_mainnet — flip derivationPath to m/44'/60'/9'/0/0 first, then revert.
   ./scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh base_mainnet

   ./scripts/deployment/pol/script_04_liquidity_manager_change_owner.sh eth_mainnet
   ./scripts/deployment/pol/script_04_liquidity_manager_change_owner.sh optimism_mainnet
   ./scripts/deployment/pol/script_04_liquidity_manager_change_owner.sh base_mainnet
   ```
3. Re-run `node scripts/audit_chains/audit_contracts_setup.js` and confirm all rows flip to `ownerCategory=dao_executor`.
4. When the DAO votes the new Tokenomics impl in, also bump the `Tokenomics` impl entry in `docs/configuration.json` from `0x1ce…0e93` → `0xaeec…d5e9` so the audit's `PROXY_TOKENOMICS()` assertion goes green.

## Test plan
- [x] `node scripts/audit_chains/audit_contracts_setup.js` — exit 0, all 8 chains processed, only the expected pre-vote `TokenomicsProxy.PROXY_TOKENOMICS()` mismatch.
- [x] `./node_modules/.bin/eslint . --ext .js,.jsx,.ts,.tsx` — 0 errors.
- [x] `npx hardhat test test/Tokenomics.js test/TokenomicsProxy.js` — 49 passing.
- [x] Cross-file address consistency check across oracles/pol/utils per chain — 0 mismatches.
- [x] On-chain proxy impl-slot reads for all 6 chains — every BBB/LM proxy points to the JSON-configured implementation.
- [x] `gitleaks protect --staged` — no leaks.
- [ ] Manual: run the change-owner scripts after merge and confirm CSV flips to `dao_executor`.
- [ ] Manual: post-DAO-vote, re-audit to confirm `PROXY_TOKENOMICS()` green.


kupermind (dismissed):

scripts/deployment/globals_mainnet.json
```diff
@@ -16,17 +16,15 @@
...
   "epochLen": "1209600",
   "donatorBlacklistAddress": "0xE85791B18F5df42163092Acc5C9da1c479AFEa9d",
-  "tokenomicsAddress": "0x87f89F94033305791B6269AE2F9cF4e09983E56e",
+  "tokenomicsAddress": "0xaeeC8bC8E5Fe28BC4dF2e9586b222924b8a0d5e9",
   "tokenomicsProxyAddress": "0xc096362fa6f4A4B1a9ea68b1043416f3381ce300",
   "treasuryAddress": "0xa0DA53447C0f6C4987964d8463da7e6628B30f82",
   "genericBondCalculatorAddress": "0x1521918961bDBC9Ed4C67a7103D5999e4130E6CB",
   "depositoryAddress": "0x52A043bcebdB2f939BaEF2E8b6F01652290eAB3f",
   "dispenserAddress": "0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0",
-  "tokenomicsTwoAddress": "0xD599ED4a5c45Fa66AA3C474b00701Fd452a2C30e",
```
DavidMinarsch, line 25: what's that?

scripts/deployment/globals_mainnet.json
```diff
@@ -16,17 +16,15 @@
...
   "OLAS_ETH_PairAddress": "0x09D1d767eDF8Fa23A64C51fa559E0688E526812F",
   "depositoryTwoAddress": "0xfF8697d8d2998d6AA2e09B405795C6F4BEeB0C81",
   "XOLAS_WXDAI_PairAddress": "0x27df632fd0dcf191C418c803801D521cd579F18e",
-  "tokenomicsThreeAddress": "0x2C1633765c66a7CF0EC801a1046347bBee7344C9",
```
DavidMinarsch, line 29: and that?

scripts/deployment/globals_mainnet.json
kupermind: Previous implementation deployments, somehow we were storing each of them separately. There is no need to do that.

scripts/deployment/globals_mainnet.json
kupermind: Same

77ph (dismissed):
# Audit review — PR #287 deployment of oracle + BBB stack

Scope: consistency review of address records vs on-chain state, audit-script logic (8 new check functions), ownership-transfer scripts, cross-chain Bridge2Burner variant config, regression check against oracle config audit 2026-05-11.

Methodology: line-by-line read of new audit JS functions + spot-verify mainnet BBB proxy impl-slot + LM proxy impl-slot + owners against on-chain state via public Ethereum RPC. Cross-check 7 oracle globals + 4 Bridge2Burner variant globals + 20-row ownership CSV against expected state.

**Verdict: Approve**. One Low, four Informational. Internal consistency is clean across all 6 chains for the verified surface. No blocking findings.

---

## L-1 (Low): Audit-script silent-pass

**Location**: `scripts/audit_chains/audit_contracts_setup.js:25-38`

`customExpect` catches assertion failures and logs them via `console.log` / `console.error`, but **does not increment a counter or propagate to process exit code**:

```javascript
function customExpect(arg1, arg2, log) {
    try {
        expect(arg1).to.equal(arg2);
    } catch (error) {
        console.log(log);
        // ... log + return normally
    }
}
```

`main()` does not aggregate mismatch counts. Script exits 0 even if multiple `customExpect` assertions failed throughout the run.

The test plan in this PR description says "exit 0, all 8 chains processed, only the expected pre-vote `TokenomicsProxy.PROXY_TOKENOMICS()` mismatch" — but "exit 0" does not mean "no mismatches", it means "no thrown exception that escaped main()". Verifying "only the expected mismatch" requires manual log review across ~150+ assertion calls. With future asymmetric per-chain checks added, an unexpected mismatch can be silently buried in log noise.

**Suggested mitigation** (one of):
1. Add a global mismatch counter; main() exits non-zero if counter > expected_known_failures.
2. Maintain an explicit allowlist of expected pre-vote mismatches; any mismatch not in the allowlist is a hard fail.
3. Document in the script header the manual review protocol (i.e., "exit 0 != pass; reader must verify zero warning lines in log").

---

## I-1 (Info): `expectedDao` OR-chain order inconsistent across check functions

**Locations**: `audit_contracts_setup.js:942` vs `:991` vs `:1072`

`BuyBackBurnerImpl.treasury` expected value is computed as:
```javascript
const expectedTreasury = utilsGlobals["bridgeMediatorAddress"] || utilsGlobals["timelockAddress"];  // L2-first
```

But `BuyBackBurnerProxy.expectedDao` is computed as:
```javascript
const expectedDao = utilsGlobals["timelockAddress"] || utilsGlobals["bridgeMediatorAddress"] || "";  // L1-first
```

Today, every per-chain `globals_*.json` has exactly ONE of the two set (L1 has timelockAddress, L2 has bridgeMediatorAddress) — so both orderings coincide on the value that IS set. The two checks produce consistent expected addresses for all current chains. **No active bug.**

If a future chain ever has both set in its globals (e.g., a cross-chain coordination context where both addresses are referenced), the two checks would disagree silently.

The existing `checkOwner` function at lines 122-148 uses explicit chainId-based branching:
```javascript
const expected = String(chainId) === "1"
    ? norm(globalsInstance["timelockAddress"])
    : norm(globalsInstance["bridgeMediatorAddress"]);
```

**Suggested mitigation**: mirror the chainId-based pattern in the new functions for forward-compatibility, or document the OR-chain assumption inline.

---

## I-2 (Info): `checkBytecode` compares last 43 bytes only

**Location**: `audit_contracts_setup.js:151-178`

`checkBytecode` compares only the last 43 bytes (`-86` hex chars) of deployed bytecode vs repo bytecode. Comment correctly explains this is to skip immutable variable info that the repo's deployed-bytecode artifact doesn't carry. The 43 bytes are the Solidity metadata CBOR suffix — a sanity check that compiler version + settings + source haven't drifted.

Adequate as a smoke-test. Caveat: a contract recompiled with different `optimizer_runs` or `evm_version` would change the metadata hash but might also be semantically equivalent — this would surface as a check failure. Likewise, two different contracts compiled with identical settings would have different metadata hashes only if their source bytes differ. The check catches obvious version drift; does not catch sophisticated bytecode swaps.

No action required.

---

## I-3 (Info): Shell scripts lack `set -e` / `set -u` / `set -o pipefail`

**Locations**: `scripts/deployment/utils/script_02_buy_back_burner_change_owner.sh`, `scripts/deployment/pol/script_04_liquidity_manager_change_owner.sh`

Neither script enables strict bash error handling. An intermediate `jq -r` returning `null` for a missing field, or `cast wallet address` failing on a missing Ledger, propagates as a string "null" or empty variable into subsequent steps — depending on later checks to catch it.

The pre-flight safety belt at lines 86-91 (script_02) / 79-83 (script_04) is the critical correctness layer: it confirms `currentOwner == deployer` before any state-mutating tx, blocking the wrong-signer class entirely. Combined with the idempotency no-op (94-97 / 85-88), the operational risk envelope is small.

**Suggested mitigation**: add `set -euo pipefail` at top of each script for defense-in-depth.

---

## I-4 (Info): Tx status reporting via `grep "status"` is fragile

**Locations**: `script_02_buy_back_burner_change_owner.sh:106`, `script_04_liquidity_manager_change_owner.sh:97`

```bash
result=$($castCmd)
echo "$result" | grep "status"
```

If `cast send` exits non-zero (RPC error, gas issue), the failure is captured in `result`. The final pipeline exit code is grep's, which exits 1 if no match found. Manual operators reading console output will see the absence of "status: 1" and notice; automated pipelines wrapping this script need to parse `result` directly.

No action required for the current manual-invocation use case.

---

## Verified clean (consistency)

- **F-2 regression check (vs oracle config audit 2026-05-11)**: all 7 chains' `oracles/globals_*.json` uniform at `maxStalenessSeconds = 86400`. F-2 finding **resolved**.
- **Slot constants in audit JS match source**: `SLOT_BUY_BACK_BURNER_PROXY` = `keccak256("BUY_BACK_BURNER_PROXY")` (matches `BuyBackBurnerProxy.sol:26`); `SLOT_PROXY_LIQUIDITY_MANAGER` = `keccak256("PROXY_LIQUIDITY_MANAGER")` (matches `LiquidityManagerProxy.sol:28`).
- **Mainnet BBB Proxy impl-slot spot-check**: storage at `0xfAd04813.../0xc6d7bd...19` reads `0xcf05126771a21a93da5a596d5cf67d8ed9f5e6e5` = `BuyBackBurnerUniswap` (configured). Match.
- **Mainnet LM Proxy impl-slot spot-check**: storage at `0x83Adc54B.../0xf7d1f641...cd` reads `0x0171d7178fbf3b48e75675dee2ea3efecef775c1` = `LiquidityManagerETH` (configured). Match.
- **Mainnet owner state**: BBB Proxy + LM Proxy both owned by `0xEB2A22b2...4E` = `AUTONOLAS_DEPLOYER`. Matches CSV (`ownershipChangeRequired = yes` for both).
- **Bridge2Burner per-chain variant config**: each of 4 chain-specific JSONs provides the expected immutables (Gnosis `gnosisOmniBridgeAddress`, Optimism `l2StandardBridgeProxyAddress`, Polygon `bridgeMediatorAddress`, Arbitrum `l2GatewayRouterAddress + l1OlasAddress`).
- **Ownership-transfer scripts**: pre-flight `currentOwner == deployer` check (script_02:86-91 / script_04:79-83) blocks wrong-signer class fast. Idempotency no-op (already-owned-by-target) present. newOwner fallback chain (`bridgeMediator || timelock`) matches deploy-script convention.
- **OR-chain timelock/bridgeMediator**: per-chain globals have exactly one of the two set; OR-chain works correctly for all current chains (see I-1 note for forward-compat).
- **CSV state (20 rows)**: every row's owner/category consistent with on-chain expectation per chain — autonolas_deployer for newly-deployed BBB/LM, dao_executor for already-transferred Tokenomics/Treasury/Dispenser/Depository/TargetDispensers, "other" for the documented Base BBB legacy owner (`0x6F7a...9Be7`).
- **Base BBB legacy case**: documented in both `script_02` header (lines 8-13) and audit-script CSV. Handling path is consistent.

---

audit-claude, 2026-05-18

77ph (approved):

kupermind (approved):

PR #306, opened 2026-07-03
fix(LiquidityManagerCore): fail-closed price guard + soft-priced exit floor

kupermind:
## Summary

Hardens `LiquidityManagerCore`'s price guard so it never prices an operation against a pool it cannot verify. Confined to one contract (`contracts/pol/LiquidityManagerCore.sol`) + tests; applied to the live proxies later via `changeImplementation` (separate PR).

### Guard fail-closed (both triggers)
`checkPoolAndGetCenterPrice` now reverts `NotEnoughHistory` when it cannot produce a verifiable 30-minute TWAP — both on a **freshly-created** pool (cardinality ≤ 1) and on an **inactive** pool (no trade within `SECONDS_AGO`). Previously both fell open to the raw, manipulable `slot0`. Every price-consuming caller (`convertToV3`, `changeRanges`, `increaseLiquidity`, permissionless `BuyBackBurner.buyBack`) inherits the revert. It is a refusal, not a permanent denial — one subsequent swap repopulates the observation buffer and the next call proceeds.

### Exits stay live — soft-priced floor (no stale minimums)
`decreaseLiquidity` keeps its original 4-arg signature and derives its slippage floor **at execution** from a new `_getExitSqrtPrice`: the TWAP when the pool is verifiable (with `slot0` bounded to it within `MAX_ALLOWED_DEVIATION`), otherwise `slot0` (fail-open). So a withdrawal is always possible on a quiet pool, is deviation-bounded on a mature pool, and — unlike owner-supplied absolute minimums — **never goes stale across a governance vote+timelock delay**. The residual on an unverifiable-but-liquid pool is a capital-bounded slip, never the empty-pool catastrophe (an exit pool always holds our own liquidity).

Deliberate distinction: **entries/trades fail-closed**; the **exit path fails-open-soft**.

### Other changes
- `collectFees`: drop the guard gate and burn **only the just-collected fee** (not the whole balance) via `_manageCollectedAmounts`; stays permissionless.
- `increaseLiquidity`: anchor `amountMin` to the TWAP center instead of raw `slot0`.
- Rename `oldestTimestamp` → `latestObsTimestamp`; remove the now-unused `_getObservationCardinality`.

## Tests
- **Unit (13):** `LiquidityManagerCorePriceGuard` (11) — guard fail-closed on both triggers + mature-deviation revert, `_getExitSqrtPrice` soft gate (quiet/fresh → slot0, mature+within → TWAP, mature+>10% → revert); `LiquidityManagerCoreCollectDecrease` (2) — `collectFees` scoped-burn (both token orderings).
- **Regression / PoC (5):** `LiquidityManagerPriceGuardRegression` — the fail-closed regressions; plus the fork-only exit-sandwich PoC (`LiquidityManagerExitSandwichFork`, manual). (18 tests across the three new suites in total.)
- **ETH & Base fork suites:** `BaseSetup` now pre-warms the freshly-created V3 pool (bump cardinality → seed wide-range liquidity → round-trip swaps spanning >`SECONDS_AGO`) so the fail-closed first seed succeeds; added fresh-pool-reverts and changeRanges-inactive-reverts cases.
- One legacy fork test (`testConvertToV3Conversion95ScanCollectFees`) is `vm.skip`-ped with a documented reason: it relied on the fail-open (its 10% drain swap moves slot0 >10% from the TWAP, so post-fix the follow-on entry ops correctly revert). Its behaviours are covered by the unit suite and `testV3BuyBackWithTwapSlippage`; a holistic rewrite is tracked as follow-up.

## Review-response fork proofs (findings #306.1, #306.2)

Two fork-only proof tests (self-skip off-fork; not in the CI `--mc` allowlist) empirically settle the two review findings **before** any code change — posted for review first.

- **`LiquidityManagerIncreaseDeadBandForkETH` — finding #306.1 (increase dead band): CONFIRMED.** Sweeps {range half-width} × {slot0-vs-TWAP deviation, driven by a real swap} on a real ETH-forked OLAS/WETH pool, measuring deviation exactly as the contract does. Because `_increaseLiquidity` anchors the desired amounts + `amountMin` to the TWAP while the NPM executes at `slot0`, the amount-space `amount{0,1}Min` check rejects mints the 10% pre-flight gate had already accepted. Dead-band cells (within the gate, yet `increaseLiquidity` reverts): half-width **120 → 0.65%**, **600 → 2.49% / 5.22%**, **1800 → 6.95%** deviation. The narrower the range, the lower the trigger (down to <1%). The test asserts the dead band exists on current code; it flips to `assertFalse` once `amountMin` is re-anchored to `slot0`.
- **`LiquidityManagerExitFailOpenForkETH` — finding #306.2 (fail-open exit slip): NOT exploitable.** Sweeps attacker slot0-manipulation size on a warm-then-inactive pool and measures the LM's realized exit loss (withdrawn amounts valued at the honest price). Result: the entry guard confines positions to warm pools, so the exit only meets fail-open when a pool later goes **inactive** — and a sandwich attacker's own front-run swap **re-activates** the pool, re-engaging the deviation gate against the stale TWAP. Every manipulation past ~10% is caught (exit reverts `Overflow`); every smaller one produces **zero measurable loss**. The test asserts `anyExploited == false`. So the "no empty-pool catastrophe" property holds empirically — via gate re-engagement, not a capital bound — and the optional owner-floor backstop is **not required**.

No contract change is proposed in this update; the `_increaseLiquidity` `amountMin` re-anchoring for #306.1 is held for review of these proofs first.

## Notes
- Isolated to `contracts/pol/` (foundry-tested); no interface/scanner/proxy change; storage layout preserved (external signature and a new `error` only) so a `changeImplementation` upgrade is safe.
- Deployment routine (`changeImplementation` script + README + runbook) and the `Vulnerabilities_list_tokenomics.md` / audit-diff docs land in follow-up PRs.
- Pre-push checks run locally: solhint (0 errors), gitleaks (no leaks), `yarn compile` (hardhat 0.8.30), `forge build`, and the CI foundry non-fork suite (all pass).


77ph:
**Internal audit — LiquidityManagerCore price-guard remediation: PASS (0 Critical / 0 High / 0 Medium).**

Reviewed `contracts/pol/LiquidityManagerCore.sol` hunk-by-hunk plus the permissionless `BuyBackBurner.buyBack`
V3 path that reuses the same guard. Both halves verified first-hand on a mainnet fork.

**What the change gets right.** The guard is correctly converted from *fail-open* to *fail-closed* on the
entry path (`checkPoolAndGetCenterPrice` reverts `NotEnoughHistory` on a pool with no verifiable 30-min TWAP,
and reverts on `slot0`-vs-TWAP deviation >10%, returning the TWAP price for minting), to a *deviation-gated
soft floor* on the exit (`_getExitSqrtPrice` keeps a withdrawal always possible on a quiet pool while
reverting a mature-pool manipulation), the increase `amountMin` is anchored to the TWAP center, and
`collectFees` now scopes the burn to just-collected amounts. Storage layout is unchanged, so
`changeImplementation` is upgrade-safe.

Two supporting results back the verdict: (1) the fix is **final in its class** — every value-bearing op
reads a pool price through exactly one of two guarded sources (the only `slot0` reader is
`_getPriceAndObservationIndexFromSlot0`, called only from those two), so both being safe on an unverifiable
pool covers the whole class; (2) the delta introduces **no new bug** — every added behaviour is a
conservative change (adds reverts on unverifiable entries, tightens an anchor, narrows a burn, adds an
always-exitable exit gate) and none locks funds, changes permissions, or adds a reentrancy surface.

**Findings — all Info/Low, none blocking merge:**
- **R1 (Low):** size `observationCardinality` to span 1800 s at peak churn for each POL pool; assert it in the seed runbook (closes the exit WRAP edge).
- **R2 (Low):** restore the `vm.skip`-ped `testConvertToV3Conversion95ScanCollectFees` as a pre-warmed round-trip (its skip is a valid invalidation, but a permanently-skipped fork test is coverage loss).
- **R3 (Info):** the audit-diff types `checkPoolAndGetCenterPrice` as `internal`; it is `public`. Doc nit.
- **R4 (Info, robustness):** the `applyDeviationGate` boolean on `_decreaseLiquidity` is safe today (only `changeRanges` passes `false`, and it pre-validates), but its safety rests on an unenforced, comment-only caller-precondition; a future `false` caller would silently re-open the exit fail-open. We agree both exit postures (soft withdrawal / strict reposition) are needed — we suggest making `_decreaseLiquidity` always gate and dropping the flag (`changeRanges` keeps its own preceding `checkPoolAndGetCenterPrice`), or passing the validated price as a typed parameter exactly as `_increaseLiquidity` already does here. Follow-up, not a blocker.
- **R5 (Info, maintainability):** `_getExitSqrtPrice` is ~90% a copy of `checkPoolAndGetCenterPrice` (and `_manageCollectedAmounts` of `_manageUtilityAmounts`). Suggest extracting the policy-free computation into one shared helper while keeping the fail-open-vs-closed policy explicit per caller — de-dup without a gate-toggling flag.

Full report and the two regression PoCs (`LiquidityManagerPriceGuardRegression` + `LiquidityManagerExitSandwichFork`) will be delivered as an internal-audit record (`audits/internal18`) and added to the test suite. Safe to merge; R1–R5 recommended as robustness follow-ups.

77ph:
Cross-note for #306 (full reply on #308): the internal18 audit carried here via #308 has been reviewed and confirmed correct. The post-audit commit `a56ab8a` is test/CI hardening only — the contract (`LiquidityManagerCore.sol`) is byte-identical to the audited state (`0c8257d` → `a56ab8a` diff is empty), so the internal18 verdict stands as-is. No security items open across #306 / #307 / #308.

The two robustness follow-ups are agreed as non-blocking dev-side changes and are **not yet in the contract**: (R4) drop the `applyDeviationGate` boolean and always run the exit-price compute in `_decreaseLiquidity` — fail-safe default over an unenforced caller precondition, `changeRanges` keeps its own preceding fail-closed check; (R5) extract the policy-free compute and keep the policy explicit per call-site. When they land as a commit here we'll re-review that delta specifically.

kupermind:
The R4/R5 robustness follow-ups have landed as a single commit on this branch: `f562a45`. Passing back for the delta re-review you flagged. It's behavior-preserving — **no ABI / storage layout / event change** — a pure robustness + de-duplication refactor.

**R4 — fail-safe exit gate.** Dropped the `applyDeviationGate` boolean on `_decreaseLiquidity`; it now always runs `_getExitSqrtPrice`. The anti-manipulation gate can no longer be silently skipped by a caller omitting the flag — the "gate-off-without-validation" call is now *unrepresentable* rather than safe-only-by-an-unenforced-precondition. `changeRanges` keeps its own preceding fail-closed `checkPoolAndGetCenterPrice`, and nothing touches the pool between that guard and the decrease, so the now-always-on gate is guaranteed to pass on the reposition path — it costs one redundant TWAP `observe` there (gas), not a functional change. On a genuinely quiet pool `_getExitSqrtPrice` still fails open and returns slot0, so exits stay always-possible.

**R5 — de-dup the security-critical primitives.**
- Extracted the policy-free price compute shared by `checkPoolAndGetCenterPrice` (entry, fail-closed) and `_getExitSqrtPrice` (exit, soft) into `_getPoolPriceFacts(pool) → (slot0Sqrt, twapAvailable, twapSqrt, deviation)`. Each callsite keeps its fail-open-vs-closed policy explicit; the slot0 read / inactivity check / observe availability / deviation math now live in one place, so the two paths can't drift apart. No security-gate-toggling boolean was introduced (per your R5 guidance).
- De-duped the byte-identical burn/transfer tail of `_manageUtilityAmounts` and `_manageCollectedAmounts` into one `_manageAmounts` primitive (renamed from `_manageCollectedAmounts`); the balance-based owner path now delegates to it. The explicit-amounts signature is what still lets `collectFees` scope to just the collected fees (VL#15 unchanged).

Net **−31 lines**. The unsafe branch R4 called out is now structurally impossible, and the drift risk R5 called out is closed.

**Verification (all green):**
- non-fork: `LiquidityManagerCorePriceGuard` 11/11, `LiquidityManagerCoreCollectDecrease` 2/2, `LiquidityManagerPriceGuardRegression` 5/5 (I1–I5 unchanged)
- fork: `LiquidityManagerETH` 24 pass / 1 pre-existing skip, `LiquidityManagerExitSandwichFork` (L1 PoC) 1/1, `LiquidityManagerBase` 8 pass / 1 skip
- `forge build` + `yarn compile` clean; branch CI green.

The regression net and the exit-sandwich PoC both still pass unchanged, which is the intended check that the refactor didn't move the security behavior — only its structure.

77ph:
R4/R5 delta (`f562a45`) re-reviewed — **PASS, no security findings.** Behavior-preserving, no ABI / storage / event change; verified by static equivalence plus on-fork execution.

**R4 — fail-safe exit gate.** Confirmed: `applyDeviationGate` dropped, `_decreaseLiquidity` always runs `_getExitSqrtPrice`, and the skip-the-gate call is now unrepresentable rather than safe-by-precondition. Only two callers; on the `changeRanges` path there is no pool interaction between `checkPoolAndGetCenterPrice` and the decrease, so the always-on gate is guaranteed to pass there (a redundant `observe`, gas only) — the reasoning holds. The redundant gate on `changeRanges` is a no-op today (that path is already protected by its own entry guard) but is deliberate forward-defense: if a future refactor ever introduced a pool interaction between the guard and the decrease, the always-on gate would catch a manipulation the old skip-path would have missed.

**R5 — de-dup.** `_getPoolPriceFacts` cleanly isolates the policy-free primitive; the entry (fail-closed) and exit (soft) paths keep their policy explicit at the callsite, with no gate-toggling flag introduced. The `_manageAmounts` unification is byte-identical to the old inline tail, and `collectFees` still scopes to just-collected fees.

**Equivalence.** Entry keeps the same revert set and order (ZeroValue → NotEnoughHistory → Overflow → return TWAP); exit keeps fail-open-on-unverifiable plus the deviation-gated slot0 (priced off slot0, not the TWAP, so a fair exit doesn't spuriously revert); burn/transfer/emit unchanged.

**Tests.** Green on `f562a45`: non-fork price-guard / regression / collect-decrease = 11 / 5 / 2 = 18; ETH fork suite 24 passed (+1 pre-existing deliberate skip, `testConvertToV3Conversion95ScanCollectFees`, whose coverage is in the collect-decrease unit suite); exit-sandwich fork PoC 1 passed. Both skips are pre-existing and legitimate — this commit changed no test file, so the skip set is unchanged. The Optimism/Balancer fork suite I couldn't run in my environment (fails at `setUp` with BAL#500); it reproduces identically at the pre-commit baseline, so it's an environment limitation, not something this change introduced — the reported 8 (+1 skip) is consistent.

**Optional, non-blocking:** the refactor lands without a new test. Nothing is left uncovered — the existing net exercises entry / exit / changeRanges / burn paths, and R4's always-on invariant is now structurally enforced (no boolean to omit). If you'd like the R4 intent explicit in-suite, a one-liner pinning "a direct `decreaseLiquidity` on a >MAX-deviation pool reverts Overflow" and "`changeRanges` succeeds on a verifiable pool" would do it, but neither is required.

Nothing blocks — R4/R5 look good.

77ph:
Follow-up: `96057d0c` verified — closes the optional test nit from the R4/R5 review. Test-only, no contract change (the `f562a45` verdict stands).

- The `testConvertToV3Conversion95ScanCollectFees` skip is removed on both suites: **ETH 24/1 → 25/0**, Base 8/1 → 9/0. On an ETH fork the rewritten test **PASSES end-to-end** (gas ~2.76M) with a real `amountsOut[0] > 0` fees-collected assertion — full lifecycle convert → collect → reconvert → increase → TWAP-anchored buyBack.
- The round-trip fee generation (OLAS→WETH→OLAS) is sound: both legs accrue fees but slot0 returns to ~the pre-warmed TWAP, so the downstream entry ops satisfy the deviation gate. The one-way-drain refusal is still covered (regression suite + `testCheckPoolAndGetCenterPrice_FlashManipulationReverts`), so no coverage is lost.
- Base fixture fixes (`_seedWideLiquidity`, `setMaxSlippages([WETH],[1000])`) are legitimate and mirror ETH — not a weakening. I couldn't run the Base (Slipstream/Balancer) suite in my environment, but the changes mirror ETH and the ETH mirror passes, consistent with the reported 9/0.
- No sibling regression — the other 24 ETH tests still pass.

Nothing blocks.

kupermind:
**R2 closed** in `96057d0` — the `vm.skip`-ped `testConvertToV3Conversion95ScanCollectFees` is restored and green on both suites (test-only; no production code touched). This clears the last open follow-up across #306/#307/#308.

Rather than assert the revert, I restored the full positive lifecycle (95% convert → collectFees → reconvert → increaseLiquidity → buyBack) under the fail-closed guard — the deviation-revert on a one-way drain is already pinned by the regression suite (I3) and `testCheckPoolAndGetCenterPrice_FlashManipulationReverts`, so a negative test here would only duplicate that while discarding the lifecycle coverage.

**The fix, per your R2 wording (“restore as a pre-warmed round-trip”):**
- Fees are generated by a price-neutral round-trip (OLAS→WETH then the WETH straight back) instead of the one-way 10% drain. Both legs accrue fees on the LM position, but slot0 returns to ~the pre-warmed TWAP, so the downstream entry ops satisfy the deviation gate. That alone restored the ETH suite.

**Base (Slipstream) needed two more, both surfaced by tracing the buyBack’s `amountOutMinimum` vs the actual fill:**
- **Re-seed depth before the buyBack** (`_seedWideLiquidity`). The Base harness removes the warm-up seed after warming (so it doesn’t steal the LM’s fees in the collectFees assertion), leaving the pool too thin for a 0.5 WETH buyBack; on-chain the pool would carry other LPs. The seed is balanced at the current price, so it moves neither slot0 nor the TWAP.
- **Configure `setMaxSlippages([WETH],[1000])` in Base setUp** (mirrors ETH). It was missing, so `mapTokenMaxSlippages[WETH] == 0` made the V3 buyBack demand the exact zero-slippage oracle quote (the H-02 fail-closed posture) and revert on any price impact. This is a test-setUp omission, not a production issue — the deploy README already documents that per-token slippage must be configured before the first V3 buyBack; the test now mirrors that required post-deploy config.

**Result:** ETH `LiquidityManagerETH` 25 pass / 0 skipped (was 24/1); Base `LiquidityManagerBase` 9 pass / 0 skipped (was 8/1). No other test changed; branch CI green.

77ph (dismissed):

DavidMinarsch:
## Review: fail-closed price guard + soft-priced exit floor

Verified locally: **18/18 new unit tests pass** — `LiquidityManagerCorePriceGuard` 11, `LiquidityManagerCoreCollectDecrease` 2, `LiquidityManagerPriceGuardRegression` 5.

The core design is right, and the split is the right one: entries fail closed, exits fail open-soft. `_getPoolPriceFacts` correctly isolates the shared primitive so the entry and exit policies cannot drift apart, and each caller applying its own policy explicitly at the call site is the clearest way to express that. `collectFees` dropping the guard is also correct — fee collection is not price-sandwichable — and scoping the burn to the just-collected `amounts` rather than `balanceOf(this)` is a clean VL#15 fix.

Credit as well for adding all three new suites to the CI foundry allowlist. That is better hygiene than the Dispenser stack, where the new fork suites run nowhere.

### 1. [High] The TWAP-anchored `amountMin` in `_increaseLiquidity` is not commensurable with the entry gate, and introduces a new revert mode

Before this PR, `_increaseLiquidity` read `slot0` itself. Desired amounts were computed at `P_slot0` and the position manager executes at `P_slot0`, so the round trip `getLiquidityForAmounts` → `getAmountsForLiquidity` returns exactly the amounts the NPM then consumes. `amountMin = desired * (MAX_BPS - maxSlippage) / MAX_BPS` was therefore **always** satisfied — the post-flight check was effectively vacuous. Wanting to fix that is fair.

But the amounts are now derived at `P_twap` while the NPM still executes at `P_slot0`. When the two differ, the NPM re-derives liquidity at `P_slot0` from the desired pair; one side binds and the other lands *below* its desired value. The problem is that a 10% deviation in **price space** does not map to a 10% shortfall in **amount space**. For a concentrated range the shortfall is amplified by how close `P_slot0` sits to a range boundary, and it reaches 100% as soon as `P_slot0` leaves `[tickLower, tickUpper]`.

That directly invalidates the calibration guidance in the constant's own comment (`LiquidityManagerCore.sol:124-126`):

> the companion deploy parameter `maxSlippage` should be set to the same percentage so the V3 NPM `amount{0,1}Min` post-flight amount-space check doesn't reject mints that this pre-flight check has already accepted.

With a narrow range there is now a real dead band: `checkPoolAndGetCenterPrice` accepts at (say) 4% deviation and `_increaseLiquidity` reverts unconditionally. The guidance held while `amountMin` was anchored to the execution price; it does not hold once the anchor moves to the TWAP.

Worth reconsidering what the TWAP anchoring buys. The price actually transacted at is `slot0` either way — manipulation resistance comes from the pre-flight deviation gate, not from the floor. Two coherent options:
- keep `amountMin` anchored to `slot0` (the execution price) and rely on the 10% gate for manipulation resistance, accepting that the floor only guards intra-transaction movement; or
- derive `maxSlippage` from the position's range width so the price-space and amount-space gates are actually commensurable.

Either way: **none of the 18 new tests exercise `P_twap != P_slot0` across a narrow range**, which is precisely the case this change alters. A unit test that sets the two apart within the 10% gate and asserts the increase still succeeds would either confirm the calibration or surface the dead band.

### 2. [Medium] The exit-sandwich PoC covers the branch that is not load-bearing

`LiquidityManagerExitSandwichFork.t.sol` proves the deviation gate reverts on a **verifiable** pool. But the design's actual risk surface is the fail-open branch — inactive pool, or freshly-created / cardinality <= 1 — where `_getExitSqrtPrice` returns raw `slot0` and `amountMin` is then derived from that same manipulated price. On that branch the slippage check provides **zero** manipulation protection by construction, since the floor moves with the attack.

The "capital-bounded slip, never the empty-pool catastrophe" claim is carrying the whole argument and is not demonstrated for that branch. On an inactive pool whose liquidity is largely our own, a searcher front-running a timelock execution can move the price a long way, and the bound is whatever their capital allows rather than anything structurally small.

If the residual is accepted, it deserves either a PoC that quantifies it on a real forked pool, or a backstop: an *optional* owner-supplied absolute floor applied only when non-zero. That keeps the anti-staleness property the PR rightly wants (the derived floor is used whenever the owner passes zero) while giving governance a way to bound the worst case on a pool it knows is quiet.

### 3. [Low] The PoC's assertion is too loose

`vm.expectRevert()` with no selector passes on *any* revert, including an unrelated one. For a test framed as the empirical proof of the exit claim, assert the `Overflow` selector explicitly. It also drives the attack with `vm.mockCall` on `slot0()` rather than an actual swap, so real pool mechanics (tick crossing, observation writes) are not exercised — worth noting in the header comment, since the file presents itself as a real-pool PoC.

### 4. [Low] Test count

The body says "Unit (12)"; the three new suites contain 18 tests between them.

DavidMinarsch:
## Still open — no response on this branch

The Dispenser stack (#309–#315) has had its review round addressed; this PR and #307 haven't been touched since my review on 12 Aug (no commits, no reply — the previous activity is the July R4/R5 round). Re-raising the one item I'd want an answer on before merge, since it is the only finding I have across either stack where I believe the code can revert during normal operation rather than merely being untidy.

### `_increaseLiquidity`'s TWAP-anchored `amountMin` vs `slot0` execution

Before this PR, `_increaseLiquidity` read `slot0` itself: desired amounts were computed at `P_slot0` and the position manager executes at `P_slot0`, so the `getLiquidityForAmounts` → `getAmountsForLiquidity` round trip returns exactly what the NPM then consumes, and `amountMin = desired * (MAX_BPS - maxSlippage) / MAX_BPS` was **always** satisfied. The post-flight check was vacuous. Wanting to fix that is fair.

Now the amounts come from `P_twap` while the NPM still executes at `P_slot0`. When they differ, the NPM re-derives liquidity at `P_slot0` from the desired pair, one side binds, and the other lands below its desired value. The issue is that a 10% deviation in **price space** does not map to a 10% shortfall in **amount space** — for a concentrated range the shortfall is amplified by how close `P_slot0` sits to a range boundary, and reaches 100% once it leaves `[tickLower, tickUpper]`.

That invalidates the calibration guidance in the constant's own comment (`LiquidityManagerCore.sol:124-126`):

> the companion deploy parameter `maxSlippage` should be set to the same percentage so the V3 NPM `amount{0,1}Min` post-flight amount-space check doesn't reject mints that this pre-flight check has already accepted

It held while `amountMin` was anchored to the execution price. It does not hold once the anchor moves to the TWAP: with a narrow range there is a dead band where `checkPoolAndGetCenterPrice` accepts at (say) 4% deviation and `_increaseLiquidity` reverts unconditionally.

Worth reconsidering what the TWAP anchoring buys, given the price actually transacted at is `slot0` either way — manipulation resistance comes from the pre-flight gate, not from the floor. Either keep `amountMin` on `slot0` and rely on the 10% gate, or derive `maxSlippage` from the range width so the two gates are commensurable.

Either way, **none of the 18 unit tests exercise `P_twap != P_slot0`**, which is exactly the case this change alters. A test that separates them *within* the 10% gate and asserts the increase still succeeds would settle it in one direction or the other — and if it fails, it fails for a reason worth knowing before this goes onto a live proxy.

The other two items from that review still stand and are cheaper: the exit-sandwich PoC covers the verifiable-pool branch rather than the fail-open branch that carries the actual residual, and its bare `vm.expectRevert()` passes on any revert.

DavidMinarsch (changes requested):
## Re-review — including `audits/internal18/README.md`, which I had not read before

I reviewed the internal audit doc in this PR for the first time (it was in the file list; I only covered the contract and tests previously). It is strong work — R4's discriminator between "a flag selecting equally-safe behaviours" and "a flag toggling a security control whose off-branch is safe only by unenforced precondition" is the right frame, and R5's "extract the policy-free computation, keep the policy explicit at each callsite" is exactly what `_getPoolPriceFacts` now does. Verified in the current head: `applyDeviationGate` is gone (0 occurrences), `_getPoolPriceFacts` is present, and the `vm.skip` is removed. R2–R5 all closed.

That makes one row of its §3 refutation table the load-bearing disagreement between us.

### 1. [Blocking] The delta-refutation for the `_increaseLiquidity` TWAP anchor is incorrect — with numbers

§3 refutes the "stricter min → DoS legit increase?" candidate as:

> On a verifiable pool (required post-FIX-1) TWAP ≈ `slot0` within 10%, so a fair increase satisfies the min

This equates a **price-space** bound (the 10% deviation gate) with an **amount-space** bound (`maxSlippage` applied to amounts). They are not interchangeable for concentrated liquidity, and the gap grows as the range narrows.

Desired amounts are computed at `P_twap`; the position manager re-derives liquidity at `P_slot0`, one side binds, and the other lands short. Simulating the actual V3 math (`getAmountsForLiquidity` / `getLiquidityForAmounts`), with `maxSlippage = 10%` and the entry gate permitting up to 10% deviation:

| range shape | increase reverts above | dead band (gate passes, mint reverts) |
|---|---|---|
| deployed shape, tickShifts `-25000/+15000` | **6.56%** deviation | 6.56% – 10% |
| ±50% | 2.38% | 2.38% – 10% |
| ±25% | 1.24% | 1.24% – 10% |
| ±10% | 0.51% | 0.51% – 10% |

So the dead band exists **even at the wide range the fork tests and runbook use**. At exactly the 10% deviation the guard is designed to permit, the deployed shape gives `a0/d0 = 0.853` against a `0.90` floor — `checkPoolAndGetCenterPrice` returns successfully and `increaseLiquidity` then reverts unconditionally.

I want to be fair about severity, which is narrower than my first review implied: at today's wide ranges this is a band near the top of the gate, not something you would hit in calm markets. But it is exactly the regime the guard exists for — a pool that has just moved several percent — and it tightens sharply the moment POL uses narrower ranges, which is the whole point of concentrated liquidity.

The same error appears in the constant's own comment (`LiquidityManagerCore.sol:124-126`: *"`maxSlippage` should be set to the same percentage so the … post-flight amount-space check doesn't reject mints that this pre-flight check has already accepted"*). That guidance held while `amountMin` was anchored to the execution price; it does not hold once the anchor moves to the TWAP.

Two coherent resolutions:
- **Keep `amountMin` on `slot0`** (the execution price) and rely on the pre-flight gate for manipulation resistance — which is what actually bounds the price transacted at, since the NPM executes at `slot0` either way; or
- **derive `maxSlippage` from the range width** so the two gates are genuinely commensurable.

Either way this needs a test that separates `P_twap` from `P_slot0` *within* the gate and asserts the increase still succeeds — none of the 18 unit tests do, which is why neither the audit nor CI caught it.

### 2. [Blocking, mechanical] `audits/internal18/` is already taken — this is your merge conflict

The PR is `CONFLICTING/DIRTY`. The sole conflict is `audits/internal18/README.md`, add/add: this branch adds *"Internal Audit 18 — LiquidityManagerCore price-guard fail-closed remediation"*, while `main` already carries *"Internal Audit 18 — autonolas-tokenomics (Dispenser rework — releasing audit)"*, merged via #312.

Two different audits claiming one number. The Dispenser one should keep 18 — it is merged and `internal19` cross-references it. Audit directories on `main` run 2–19, so **renumber this one to `audits/internal20/`** and update its own heading. No other file in #306 or #307 references `internal18`, so the rename is self-contained.

### 3. Still open from the previous round

- The exit-sandwich PoC covers the verifiable-pool branch; the fail-open branch that carries the actual residual is untested.
- That PoC's bare `vm.expectRevert()` passes on any revert — assert the `Overflow` selector.

R1 (size `observationCardinality` to span 1800 s at peak churn and assert it in the runbook) is a genuine pre-deploy item and belongs on the release checklist rather than in code.

kupermind:
## Empirical proofs for findings #306.1 and #306.2 (posted for review before any fix)

Two fork proof tests settle both findings with reproducible evidence rather than argument. Both run on a real ETH-mainnet-forked OLAS/WETH V3 pool, are fork-only (self-skip off-fork; not in the CI `--mc` allowlist), and are on this branch.

### #306.1 — increase dead band: **CONFIRMED**

`test/LiquidityManagerIncreaseDeadBandForkETH.t.sol` sweeps {range half-width} × {slot0-vs-TWAP deviation, driven by a **real swap**} — a real swap is required because the NPM's mint reads the pool's *storage* `slot0` (a `vm.mockCall` on `slot0()` wouldn't move it). Per cell it measures the deviation exactly as `_getPoolPriceFacts` does, confirms it's within the 10% gate, and records whether `increaseLiquidity` reverts.

Dead-band cells (deviation within the 10% gate, yet `increaseLiquidity` reverts):

| range half-width (ticks) | deviation | within gate | increase reverts |
|---|---|---|---|
| 120 | 0.65% | ✓ | ✓ |
| 600 | 2.49% | ✓ | ✓ |
| 600 | 5.22% | ✓ | ✓ |
| 1800 | 6.95% | ✓ | ✓ |

The mechanism is exactly as described: `_increaseLiquidity` anchors the desired amounts and `amountMin` to the TWAP center while the NPM executes at `slot0`; the amount-space `amount{0,1}Min` check then rejects mints the pre-flight gate accepted. The narrower the range, the smaller the deviation that triggers it (down to <1%). The test currently asserts the dead band exists; it flips to `assertFalse(anyDeadBand)` as the regression guard once a fix lands.

**Proposed fix (held for review):** re-anchor `amountMin` to `slot0` (the execution price). Manipulation resistance stays in the pre-flight 10% deviation gate — where it belongs — and the floor then does its actual job (an intra-tx slippage bound relative to the execution price).

### #306.2 — fail-open exit slip: **NOT exploitable**

`test/LiquidityManagerExitFailOpenForkETH.t.sol` sweeps attacker slot0-manipulation size on a warm-then-inactive pool and measures the LM's realized exit loss (withdrawn amounts valued at the honest pre-manipulation price).

| attacker OLAS in | slot0 deviation | exit outcome | LM loss |
|---|---|---|---|
| 50k – 1M | 0.11% – 2.18% | succeeds | 0 |
| 5M | 10.26% | reverts (`Overflow`) | — |
| 20M – 80M | 33% – 100% | reverts (`Overflow`) | — |

Two structural reasons, both observed:
1. The **entry guard fails closed**, so the LM only ever holds a position on a pool that was verifiable at mint; the exit meets fail-open only when such a pool later goes **inactive**.
2. A sandwich attacker's own front-run swap **writes an observation and re-activates the pool**, re-engaging the deviation gate against the still-stale TWAP: any manipulation past ~10% is caught (exit reverts), anything smaller produces no measurable loss.

So the "capital-bounded slip, never the empty-pool catastrophe" property holds empirically — via gate re-engagement rather than a capital bound — and the optional owner-supplied floor is **not required**. The test asserts `anyExploited == false`. (The "freshly-created / cardinality ≤ 1" sub-branch is unreachable for the LM's own positions: the entry guard prevents minting on such a pool in the first place.)

### Ask

Please review these two proofs before I touch `_increaseLiquidity`. If the #306.1 evidence and the proposed `slot0` re-anchoring look right, I'll apply the fix and flip the dead-band assertion to the regression guard. Remaining coverage — extending the #306.1 sweep to Base/Optimism, and a Balancer→UniV3 converter for Polygon — is planned as follow-up.

77ph:
@DavidMinarsch @kupermind — reviewed the full thread and re-checked the code first-hand on `fix/lm-price-guard-failclosed` (`b118ad9`). Short version: **David's #306.1 is correct, and we should own that it is a miss in our July sign-off** — our #306 note called the delta "no new bug… tightens an anchor," and that anchor change is precisely what creates the dead band. Detail + one addition below.

### #306.1 — confirmed, and it is a miss on our side

`_increaseLiquidity` anchors the desired amounts and `amount{0,1}Min` to the **TWAP center** (`checkPoolAndGetCenterPrice`), while the NPM executes the mint at **slot0**. When the two differ (even inside the 10% gate) the NPM re-derives liquidity at slot0 from the desired pair, one leg binds and the other lands below its TWAP-anchored floor → revert. As David notes, a 10% deviation in price space is not a 10% shortfall in amount space; near a range boundary it amplifies (→100% once slot0 leaves the range), so a sub-1% deviation can trip it on a narrow range. kupermind's fork sweep reproduces exactly this. Nothing to add to the mechanism — it is right.

Why we should have caught it: the **exit path already does it correctly** — `_decreaseLiquidity` anchors `amountMin` to `_getExitSqrtPrice` (slot0, "the exact price the position manager withdraws at"), with the manipulation check kept as a separate pre-flight deviation gate. The entry path violates that exact principle its sibling documents. A direct entry-vs-exit comparison of the `amountMin` anchor would have surfaced it, and ours didn't. That's on us; flagging it plainly.

### One addition: the same anchor mismatch is on the initial-mint path, not only `increaseLiquidity`

`_optimizeTicksAndMintPosition` → `_mintV3` builds `amountsIn` and `aMin = amountsIn*(MAX_BPS-maxSlippage)/MAX_BPS` at `sqrtP` = the TWAP center, and `_mintV3` (LiquidityManagerETH.sol) **ignores** its `centerSqrtPriceX96` argument and forwards those TWAP-anchored mins to `NPM.mint()`, which executes at slot0 — with the range placed around the TWAP center (`_calculateTicksAndMintPosition`). That is the same dead band as #306.1, on `convertToV3`'s primary (first-mint) path, and arguably a touch more exposed since slot0 can sit at the fresh range's boundary. So the proposed **slot0 re-anchoring should cover both the mint and the increase paths** (the exit is already fine); a fix limited to `_increaseLiquidity` would leave `convertToV3`'s initial mint able to revert in the same band.

### #306.2 — agree it is not exploitable, no fix needed

kupermind's second proof is sound: the entry guard fails closed, so the LM only ever holds a position on a pool that was verifiable at mint; and a sandwich front-run writes an observation that **re-activates** the pool, re-engaging the deviation gate against the stale TWAP — so anything past ~10% reverts and anything smaller shows no measurable loss. The "capital-bounded slip, never the empty-pool catastrophe" property holds via gate re-engagement, and the optional owner floor isn't required.

### Direction

Endorse the fix as described — re-anchor `amountMin` to slot0 (the execution price) and keep manipulation resistance in the pre-flight 10% deviation gate, applied to **both** the mint and increase paths — and flip the dead-band assertion to the regression guard. Happy to re-review that delta specifically when it lands (both paths + the flipped assertion), same as the R4/R5 round.

— 77ph

DavidMinarsch:
## Re: the two empirical proofs — #306.1 endorsed, #306.2 accepted with one untested sub-branch

Thanks for settling these with evidence rather than argument, and for posting before touching the code.

### #306.1 — confirmed independently; apply the proposed fix

Your fork sweep reproduces what I derived analytically from the V3 amount math, which is a good sign for both. My crossover points (the deviation above which the mint reverts, `maxSlippage = 10%`) versus your observed dead-band cells:

| shape | my analytic crossover | your observed revert |
|---|---|---|
| ±1.2% range ≈ 120 ticks | ~0.5% | 0.65% ✓ |
| ±6.2% range ≈ 600 ticks | <1% | 2.49%, 5.22% ✓ |
| deployed `-25000/+15000` | 6.56% | 1800 ticks @ 6.95% ✓ |

Consistent throughout — your cells are sample points above the crossover rather than the crossover itself, so the two agree.

**Re-anchoring `amountMin` to `slot0` is the right fix**, and worth stating why rather than just agreeing: the NPM executes at `slot0` no matter what the floor is anchored to, so the floor cannot provide manipulation resistance — that genuinely lives in the pre-flight deviation gate. Anchoring it to a price the transaction will *not* execute at converts it from a slippage bound into a price-agreement requirement, which is what produces the dead band. Post-fix the floor does its actual job (an intra-tx bound against the execution price) and the 10% gate does its own, with no overlap and no gap.

Please apply it and flip the assertion to `assertFalse(anyDeadBand)` as you proposed. Good call keeping the sweep as the permanent regression guard rather than a one-off.

### #306.2 — I accept the mechanism, and withdraw the owner-supplied floor

The re-activation argument is a genuinely better answer than the capital bound the PR originally claimed, and it is one I had not considered: the attacker's own front-run swap writes an observation, the pool reads as active again, and the deviation gate re-engages against a still-stale TWAP. Your table shows it cleanly — everything past ~10% reverts, everything below produces no measurable loss. The optional owner-supplied floor is unnecessary. Withdrawn.

**One sub-branch the PoC does not cover.** `_getPoolPriceFacts` has *two* fail-open triggers, and the test only exercises the first:

1. **Inactive pool** — `latestObsTimestamp + SECONDS_AGO < block.timestamp`. This is what `vm.warp(block.timestamp + SECONDS_AGO + 60)` sets up, and it is the branch your re-activation argument defeats. ✓
2. **`observe(1800)` reverts** — `LiquidityManagerCore.sol:1201-1202` returns `(slot0, false, 0, 0)` on `!success`, skipping the gate identically.

The re-activation argument does not rescue (2). An attacker's swap writes an observation, but writing evicts the oldest entry, so the buffer's span gets *shorter*, not longer — `observe(1800)` still reverts, `twapAvailable` stays false, and the gate stays off for the exit. Unlike (1), the attack does not disarm itself.

That is precisely the condition audit finding **R1** describes: a pool whose `observationCardinality` is too small to span 1800 s at peak churn. R1 is currently rated *Low, config* — but on this branch it is the sole precondition for the exit's only manipulation defence, which makes it a security precondition rather than a deployment nicety. Note it is also distinct from the "cardinality ≤ 1" case you correctly rule out via the entry guard: a pool can be well past cardinality 1, pass the entry guard at mint, and still wrap below 1800 s later under churn.

Two ways to close it, either is fine:
- extend the sweep with a deliberately under-sized cardinality pool and show the loss (I expect this one *is* exploitable, which would be worth knowing); or
- state the dependency explicitly in the audit and make the cardinality-spans-1800 s assertion a hard release gate rather than a runbook suggestion — see the related note on #307.

### Still outstanding

The `audits/internal18/` collision is unresolved and is this PR's merge conflict — `main` already has *"Internal Audit 18 — autonolas-tokenomics (Dispenser rework)"* from #312, and `internal19` cross-references it. Renumber this one to `audits/internal20/` (2–19 are taken) and update its heading; nothing else in #306 or #307 references the path.

kupermind:
## Fix landed (`47d8b86`) — #306.1 dead band re-anchored to slot0, both entry paths

Applied the endorsed fix, including the initial-mint path raised in review:

- **New `_slot0MinAmounts(slot0, tickLower, tickUpper, desired)`** — `getLiquidityForAmounts` then `getAmountsForLiquidity` at slot0, floored by `maxSlippage` — i.e. what the position manager actually consumes at the execution price.
- **Initial-mint path (`_optimizeTicksAndMintPosition`):** range placement stays anchored to the TWAP center (the scanner keeps `sqrtP` = TWAP, manipulation-resistant); only `amountMin` moves to slot0 via `_slot0MinAmounts`. Fixes `convertToV3`'s first mint and `changeRanges`' reposition.
- **Increase path:** `increaseLiquidity` and `convertToV3`'s increase branch keep `checkPoolAndGetCenterPrice` as the pre-flight gate, then pass **slot0** to `_increaseLiquidity`, so its liquidity math + `amountMin` anchor to the execution price.
- **Exit path (`_decreaseLiquidity`) unchanged** — it already anchored to `_getExitSqrtPrice` (slot0); this brings entry in line with its sibling.

The dead-band fork proof (`LiquidityManagerIncreaseDeadBandForkETH`) is flipped to a regression guard (`assertFalse(anyDeadBand)`): the same sweep now finds **zero** dead-band cells.

**Verification:** `LiquidityManagerETH` fork **25/25** (real `convertToV3` mint + `increaseLiquidity` + `changeRanges`), `LiquidityManagerBase` fork **9/9** (Slipstream mint/increase), POL unit suites **18/18**, exit fail-open + exit-sandwich unchanged, solhint clean, hardhat compile OK.

Ready for the delta re-review (both paths + the flipped assertion).

kupermind:
## #306.1 fixed on both entry paths (mint + increase); plus the buffer-wrap fail-open trigger

Thanks for the careful re-check — and for the mint-path addition, which was the important catch. Both points are now handled on the branch (`14f2d3c`).

### #306.1 — re-anchored to slot0 on **both** entry paths

The fix (`47d8b86`) moves the `amount{0,1}Min` floor from the TWAP center to `slot0` (the execution price) via a new `_slot0MinAmounts(slot0Sqrt, tickLower, tickUpper, desired)` helper, and — per the review addition — it covers the **initial-mint path**, not only `increaseLiquidity`:

- `_optimizeTicksAndMintPosition` now derives `aMin` from `slot0` and passes it into `_mintV3`. Both concrete impls (`LiquidityManagerETH._mintV3`, `LiquidityManagerOptimism._mintV3`) forward the passed `amountsMin` straight to `NPM.mint()` and ignore the center-price argument — so once the passed mins are slot0-anchored, the ignored argument is moot and the first-mint band closes too.
- `_increaseLiquidity` (public `increaseLiquidity` + the `convertToV3` increase branch) reads `slot0` and anchors its floor there.
- The scanner still places the **range** around the TWAP center; only the amount-space floor moved. Manipulation resistance stays where it belongs — in the pre-flight 10% deviation gate.
- The exit path was already correct (`_getExitSqrtPrice` = slot0), so it's unchanged.

The dead-band fork sweep flipped to the regression guard (`assertFalse(anyDeadBand)`) and now reports **0** in-gate dead-band cells (same sweep that previously found the 0.65% / 2.49% / 5.22% / 6.95% cells).

### Buffer-wrap fail-open trigger — reachable, but not a decreaseLiquidity value-loss

The earlier #306.2 proof covered only the **inactive-pool** fail-open trigger (self-defeating: the attacker's own front-run re-activates the pool and re-engages the gate). I've now added the **second** trigger — `observe(1800)` reverting because the observation buffer wrapped below 1800s under churn — as `test_exit_failOpen_bufferWrap_noValueLoss`:

- It is **genuinely reachable**: re-activation does *not* rescue it (writing an observation shortens the buffer's span, it doesn't extend it). The test confirms the gate is off — entry fails closed, and a 20M-OLAS manipulation the gate would otherwise catch does **not** revert the exit.
- But it is **not a value-loss hole** for `decreaseLiquidity`: removing liquidity at the manipulated price returns amounts worth `>=` the fair value at the true price (the exit `amountMin` guards the token *split*, not value). Measured loss is **0 bps**.

So I've corrected the #306.2 framing: the "no catastrophe" property holds on *both* triggers, via gate re-engagement on the inactive one and via the LP-removal value identity on the buffer-wrap one. Sizing `observationCardinality` for peak churn so `observe(1800)` never reverts is therefore **defense-in-depth** (keep the gate on, avoid the unbalanced-split residual) rather than a loss fix — and I've promoted it to a hard, per-pool release gate in the deployment README + runbook on the companion PR.

DavidMinarsch:
## Re-review — `47d8b86`, `14f2d3c`, renumbering

Verified on this head: POL unit suites **18/18** (PriceGuard 11, CollectDecrease 2, PriceGuardRegression 5), and the dead-band sweep is flipped to `assertFalse(anyDeadBand, "…regression")`.

### #306.1 — fix is correct

`_slot0MinAmounts` is the right shape, and the split you chose is better than what I proposed: **range placement stays on the TWAP center** (so the manipulation-resistant price still decides *where* liquidity goes) while **only the floor moves to slot0** (so it is satisfiable at the price the NPM actually executes at). I had lumped those together; separating them keeps the anti-manipulation property in the placement and removes the dead band from the floor.

Gate ordering checked at all three call sites — `checkPoolAndGetCenterPrice` runs before `_getPriceAndObservationIndexFromSlot0` in `increaseLiquidity` (`:1058-1061`), the `convertToV3` increase branch (`:785-789`), and via the callers of `_optimizeTicksAndMintPosition` — so the slot0 that gets anchored to is always one the 10% gate has already accepted. No path reaches the new helper ungated.

Worth recording explicitly for whoever reads this next: post-fix, `amount{0,1}Min` is **vacuous** — desired amounts and the floor are both computed at slot0 in the same transaction the NPM executes in, so the check can never bind. That is correct and intentional (it is what the pre-#306 code did, and what upstream V3 integrations do), but it means the pre-flight 10% deviation gate is now the *sole* entry protection. `MAX_ALLOWED_DEVIATION` is carrying all of it, so it deserves to be reviewed as a security parameter rather than a sanity bound.

### #306.2 — fully resolved, both sub-branches

The buffer-wrap test closes the gap I raised, and the conclusion is stronger than "no measurable loss" — it holds for a structural reason worth putting in the audit: an LP position's value is **minimised at the true price**. Burning liquidity `L` over `[A,B]` at a manipulated `Pm` and valuing the basket at the honest `Pt` gives

```
dV/dPm = (L / 2√Pm) · (1 − Pt/Pm)
```

which is negative below `Pt` and positive above it. So *any* manipulation away from the true price hands the LM a basket worth **more** at the true price, not less — the attacker pays swap fees and slippage to improve the victim's exit. That is why `decreaseLiquidity` is not sandwichable in either direction, and it is independent of the gate. Combined with the inactive-pool branch being self-defeating via re-activation, the exit is sound on both triggers. My request for an owner-supplied floor is withdrawn.

### One thing left — the heading

The renumbering landed for the directory but not inside the file. `audits/internal20/README.md` line 1 still reads:

```
# Internal Audit 18 — LiquidityManagerCore price-guard fail-closed remediation
```

`internal18` on `main` is the Dispenser rework audit, so the document now contradicts its own path. One line. No other file references either path — I grepped `docs/` and `scripts/` on both this branch and #307 and found no links, so nothing else needs touching.

Happy to approve once that heading says 20.

kupermind:
## Heading fixed → "Internal Audit 20"; both recorded points folded into the doc

- **Heading (`3c9ea62`).** `audits/internal20/README.md` line 1 now reads **`# Internal Audit 20`**. Root cause of the miss: the directory rename was staged, but the in-file title edit landed after the `git add`, so the merge commit captured the pre-edit line. Fixed and verified.

Thanks for the close read on the split — agreed that keeping **range placement on the TWAP center** while moving **only the floor to slot0** is the cleaner decomposition (anti-manipulation stays in the placement + the pre-flight gate; the floor just needs to be satisfiable at the executed price).

While in the file I also reconciled the write-up with the shipped fix and recorded your two points (`6885876`):

- **`MAX_ALLOWED_DEVIATION` is now the sole entry defence** — recorded as **R6**. You're right that post-fix `amount{0,1}Min` is vacuous by construction (desired + floor both at slot0 in the executing tx), so the 10% pre-flight gate carries all entry manipulation resistance and should be reviewed as a security parameter, not a sanity bound. I also fixed the stale rows that still described the *interim* TWAP-center anchor (FIX-2a, the §2.1 C5 row, and the §3 delta row) — the audit predated the slot0 re-anchor and those hadn't caught up.
- **Exit convexity** — added your closed-form `dV/dPm = (L / 2√Pm)·(1 − Pt/Pm)` to the withdraw-sandwich refutation, noting it's negative below `Pt` / positive above, so any manipulation hands the LM a basket worth *more* at the true price — independent of the gate, and consistent with the buffer-wrap proof measuring 0 bps with the gate provably off. Glad that lets the owner-floor request be withdrawn.

Ready for the approval pass.

DavidMinarsch (approved):
Approving. The heading reads `Internal Audit 20` and everything since the verified head is docs-only (`git diff pr306h..HEAD -- contracts/ test/` is empty), so the 18/18 unit result and the flipped dead-band guard still stand.

The reconciliation commit is the right call, and notably it **replaces** the incorrect refutation row rather than footnoting it. That is the correct treatment here — unlike internal18/internal19, which were accurate at their pins and merely overtaken, this row described interim code that no longer exists in the tree, so leaving it with a superseded note would have preserved a false statement about the shipped design. R1–R5 → R1–R6 is consistent throughout (no stale references left).

**R6 is a better articulation than my original note.** I framed it as "worth knowing"; recording it as a numbered finding that promotes `MAX_ALLOWED_DEVIATION` from a sanity bound to a security parameter — with the explicit instruction to review the 10% against the shallowest POL pool's single-block manipulability and to treat future changes as security changes — is what makes it survive into the next round. Same for folding the `dV/dPm` closed form into the withdraw-sandwich refutation: it turns an empirical "0 bps measured" into a structural result that holds independent of the gate.

### One non-blocking nit — FIX-2a overstates the mint path

The FIX-2a row reads:

> On both entry paths — the initial mint (`_optimizeTicksAndMintPosition`) and `_increaseLiquidity` — `amountMin` **and the liquidity math** are anchored to **`slot0`**

That is accurate for `_increaseLiquidity`, which receives `slot0Sqrt` and runs `getLiquidityForAmounts` at it. It is not accurate for the mint path: `optimizeLiquidityAmounts` still runs at `sqrtP` (the TWAP center) and only `aMin` is computed at `slot0` via `_slot0MinAmounts` — exactly as the code comment at `:598-601` says ("Range placement stays anchored to the TWAP center … but amountMin is anchored to slot0").

The behaviour is correct — `_slot0MinAmounts` re-derives liquidity from `amountsIn` at `slot0` and floors the resulting amounts, so `aMin` is always satisfiable at execution and the dead band is closed. It is only the sentence that overreaches, and it overreaches on precisely the TWAP-vs-`slot0` split that is the point of the fix, so it is worth getting right. Suggest: "*`amountMin` is anchored to `slot0` on both entry paths; the liquidity math follows `slot0` on the increase path, while the mint keeps the TWAP center for range placement and amount optimisation.*"

Worth one adjacent sentence too: because mint amounts are still optimised at the TWAP center while the NPM consumes a `slot0`-derived split, a `slot0 ≠ TWAP` within the gate leaves larger leftovers on mint. Those sweep to treasury via `_manageUtilityAmounts`, so it is deployed-capital efficiency rather than loss — but it is the visible residual of the split and a reader will otherwise wonder where it went.

Not blocking — fix on the way past if you touch the file again.

77ph:
@kupermind @DavidMinarsch — re-reviewed the merged fix (`47d8b86` + the renumbering) as offered. **Confirmed correct and complete; closing this from our side.**

- **Both entry paths covered.** The initial-mint path we flagged is closed the same way as `increaseLiquidity`: `_optimizeTicksAndMintPosition` keeps `sqrtP` = TWAP for range *placement* but derives `aMin` via the new `_slot0MinAmounts` at slot0; the increase path passes slot0 through `_increaseLiquidity` and recomputes the desired amounts at slot0. All three entries (`convertToV3`, `changeRanges`, `increaseLiquidity`) still run the fail-closed `checkPoolAndGetCenterPrice` before the raw-slot0 read, so manipulation resistance is unchanged. `_slot0MinAmounts` mirrors the position manager's own `getLiquidityForAmounts → getAmountsForLiquidity` at slot0, so the floor is satisfiable by exactly the `maxSlippage` margin and degrades to a single-sided floor when slot0 sits outside the range. The dead-band sweep flipped to `assertFalse(anyDeadBand)` is the right regression guard. No new bug.

- **One emphasis, endorsing your own R6.** As the internal20 note now records, once `amountMin` is anchored to the same slot0 the mint executes at, the amount-space floor is **vacuous by construction** — so `MAX_ALLOWED_DEVIATION` (10%) is the *sole* entry manipulation defence, not a sanity bound. That is the correct and unavoidable trade-off (a tight amount-floor *is* the dead band). Worth pinning the 10% against the **shallowest POL pool's single-block manipulability** and treating any future change to it as a security change, not a tuning knob. The exit-side convexity argument (`dV/dPm = (L/2√Pm)(1 − Pt/Pm)`, minimum at `Pm = Pt`) is sound, including out of range.

Thanks for the quick turnaround on the mint-path catch. One doc-reconciliation carries over to #307 (the audit-diff / vulnerabilities-list still describe the entry anchor as the pre-fix TWAP-center design) — noting it there.

— 77ph

PR #215, opened 2025-06-23
doc: internal audit 7

77ph:
- Internal audit based on DefaultTargetDispenserL2 update

PR #307, opened 2026-07-03
docs/deploy(pol): LiquidityManager changeImplementation upgrade routine + price-guard audit-diff

kupermind:
Deployment/docs companion to the price-guard fix (#306). No contract change.

## Contents
- **`scripts/deployment/pol/script_05_liquidity_manager_change_implementation.sh`** (new) — rolls a new `LiquidityManager*` impl onto the **existing** proxy via `changeImplementation(address)`. The LM proxies are still owned by the **Autonolas deployer EOA** (ownership left with the deployer while POL is not operational), so this is a plain single-signer `cast send`, not a Timelock/DAO proposal. Mirrors the BBB `utils/script_01`.
- **`scripts/deployment/pol/README.md`** — adds an "Upgrading the implementation" section: storage-safe swap (no proxy/scanner redeploy), and the **upgrade-before-seed** ordering (the fixed guard is fail-closed, so the first `convertToV3` reverts on a brand-new/quiet pool → pre-warm the pool before the first seed). The pre-warm block now makes sizing `observationCardinality` for **peak** churn a hard, per-pool release gate (audit R1).
- **`docs/liquidity_migration_runbook.md`** (new) — **included in this PR**, reframed as an operational procedure for the fixed system (see the note below).
- **`docs/lm_price_guard_audit_diff.md`** (new) — a short tagged prior-state → new-state map of the fix, for reviewers.
- **`docs/Vulnerabilities_list_tokenomics.md` §26** — item 2 corrected to the **soft-priced exit floor** (`_getExitSqrtPrice`), replacing the earlier owner-supplied-minimums description (the fix evolved to avoid stale minimums across a governance delay).

## The migration runbook IS in this PR (this corrects an earlier body)
`docs/liquidity_migration_runbook.md` (all 251 lines) **is published here**, reframed as an operational procedure for the fixed system. It carries **no** bounty-dispute or exposure material — an earlier internal draft described the issue as open/unfixed, and that reframe landed before this PR. An earlier version of this PR body wrongly stated the runbook was withheld; that claim has been removed. **The runbook is not held back.**

## Held back (not in this PR)
- The **`#15`/`#16` removal** and dropping VL#26's "planned/interim" language wait for the deployment PR (per the disclosure sequencing — the public list must not say "closed" before the impl is live).
- *(Not the runbook — see the section above; it is included here.)*

## Checks
- gitleaks: no leaks (new deployment script scanned).
- No contract/Solidity change, so ESLint/solhint/compile/tests are unaffected (that surface is in #306).


77ph:
**Internal audit — deployment runbook + price-guard audit-diff: PASS.**

Reviewed the `changeImplementation` upgrade routine and the price-guard audit-diff that accompany the
LiquidityManagerCore remediation (#306). The upgrade path is storage-layout safe (base↔fix state-variable
layout is identical: slot0 `owner`/`maxSlippage`/`_locked`, slot1 map; the added error and renamed local
occupy no storage), the deploy script guards against an empty implementation/proxy, and the
upgrade-before-seed ordering is correctly documented.

Two doc/runbook items carried from the code audit:
- **R1 (Low):** the runbook already bumps `observationCardinality` at seed — make it an explicit
  **assertion** that each POL pool's cardinality spans `SECONDS_AGO = 1800 s` at peak churn. This closes the
  only exit path that can still fall open (a pool whose buffer is too small to interpolate the window even
  after a swap).
- **R3 (Info):** the audit-diff describes `checkPoolAndGetCenterPrice` as `internal`; in code it is `public`
  (it must be — `BuyBackBurner` calls it cross-contract). Correct the signature in the doc.

Recommend #307 merge together with #306, and that the real deployment proceed only after this audit is on
record. The full internal-audit report (`audits/internal18`) covers the code findings in detail.

77ph (approved):
`8bfe36aa` re-reviewed — **PASS, approving.**

Verified the audit-diff reconciliation against the shipped code at #306 `96057d0c`:

- **Storage-safe upgrade confirmed.** The only storage-occupying members (`owner`, `maxSlippage`, `_locked`, `mapPoolAddressPositionIds`) are unchanged in name/type/order/width vs the deployed `v1.4.3` impl. Every other delta is a function body, the one renamed `error`, or an internal helper (`_getPoolPriceFacts`, `_manageAmounts`, the dropped `applyDeviationGate` param) — none occupies a slot. `changeImplementation` is storage-safe.
- **[R4] holds:** `_decreaseLiquidity(pool, id, decreaseRate)` — the boolean is gone and it unconditionally runs `_getExitSqrtPrice`; the "gate-off-without-validation" call is now unrepresentable.
- **[R5] holds:** `checkPoolAndGetCenterPrice` / `_getExitSqrtPrice` share the policy-free `_getPoolPriceFacts` (no gate-toggling flag; each call-site keeps its own fail-open-vs-closed posture), and the burn/transfer tail is the single `_manageAmounts` primitive.
- **[VL#15]/[FIX-3] holds:** `collectFees` burns only the just-collected fee via `_manageAmounts`, not the whole balance; stays permissionless.
- `script_05` still guards against `changeImplementation(0x0/null)`, and the pre-seed ordering note (upgrade before seeding, then pre-warm per the runbook) is correct.

One non-blocking doc nit (optional): the new baseline paragraph calls `v1.4.3` "the tag that first introduced `LiquidityManagerCore` (see `CHANGELOG.md`, 'Added — Protocol-Owned Liquidity')". The contract actually predates `v1.4.3` (it's present at `v1.4.2`), and that heading isn't in the `v1.4.3` CHANGELOG. The material claim is correct and unaffected — `v1.4.3` is the impl currently behind the proxies (nothing touched the file between `v1.4.3` and this PR), so it is the right storage baseline. Consider rewording "the tag that first introduced" → "the released impl currently behind the proxies".

77ph (dismissed):

DavidMinarsch:
## Review: LiquidityManager `changeImplementation` routine + price-guard audit diff

Companion review to #306. No contract change here, so this is about the deployment routine and the docs — which for an EOA-signed upgrade on a contract that will hold POL is the whole control surface.

### 1. [High] The body and the diff disagree about the migration runbook

The "Held back (not in this PR)" section states:

> The **full migration runbook** is intentionally **not published here**: its current internal draft describes the issue as open/unfixed with bounty-dispute and exposure detail and needs a reframe (operational procedure for the fixed system) before it goes public.

The diff adds `docs/liquidity_migration_runbook.md` — all 251 lines of it.

I read the added file and the content is clean: it is an operational procedure for the fixed system, and it carries none of the bounty-dispute or exposure material the body warns about. So this reads as a stale body after the reframe actually landed, not a disclosure incident. But the body is what a reviewer approves against, and someone trusting that section would not open a 251-line public-facing document. Please correct the body before merge — on a disclosure-sequenced PR, "this file is withheld" is exactly the claim that needs to match the diff.

### 2. [Medium] The upgrade branch skips the ownership handover, on the path you are actually taking

Runbook §1.1 step 4 reads:

> **Deploy `LiquidityManagerETH` + `LiquidityManagerProxy`; set Timelock as owner** (or upgrade the existing proxy to the fixed impl via `changeImplementation`, deployment README).

The parenthetical presents the upgrade as an *alternative to* setting the Timelock as owner — and the upgrade branch is the one that applies to the current deployment, since the proxies already exist and are EOA-owned. Steps 5 and 6 then transfer V2 LP from Treasury to the LiquidityManager and call `convertToV3`. Following that branch literally, POL is seeded into a proxy still owned by the deployer EOA. §1.2 step 5 has the same shape for the L2s.

That contradicts the runbook's own safety framing a few sections later:

> **Inert to non-owners while idle:** every value-moving LM function (`convertToV3`, `increaseLiquidity`, `transferToken`, …) is `onlyOwner`, and the owner is the L2 `BridgeMediator`.

and it undercuts the "no funds at risk" argument in VL#26, which rests on the LiquidityManager being unseeded and non-custodial while ownership sits with the deployer. The EOA-ownership decision is defensible precisely *because* POL is not operational — so the transition out of that state is the step that needs the gate.

Make the ownership handover an explicit precondition on the seed step, in both the runbook and `scripts/deployment/pol/README.md`. The README's "Ordering — upgrade before seeding" block is the natural place: it already enumerates the pre-warm preconditions, and "ownership moved to Timelock / BridgeMediator" belongs in the same list.

### 3. [Medium] `script_05` reports success on a failed upgrade

The script ends with:

```bash
result=$($castCmd)
echo "$result" | grep "status"
```

It never inspects the receipt status and always exits 0, so a reverted `changeImplementation` prints a status line and looks like a completed run. It also never reads back the `PROXY_LIQUIDITY_MANAGER` slot to confirm the swap actually landed.

For a plain single-signer `cast send` with no proposal review in front of it, that post-condition check is the only safety net there is. A `cast storage`/`cast call` read-back asserted against `liquidityManagerAddress`, plus a non-zero exit on mismatch, would close it. (Same class as the deploy-script finding on `autonolas-governance` #215, where the post-deploy checks likewise print rather than assert.)

The null/empty guards on both globals fields are good — worth extending them to the literal zero address for symmetry with the Dispenser proxy script.

### 4. [Low] `changeImplementation` accepts an EOA

`LiquidityManagerCore.sol:624` guards the zero address but not code size. A `delegatecall` to an address with no code returns `success = true` with empty returndata, so the proxy would answer every call with nothing — including `changeImplementation` itself, making the brick permanent. The same gap exists in `Tokenomics` and in the new `Dispenser` (see the note on #309), so it is convention rather than a regression, but it is most acute here: this is the one upgrade path with a single EOA signer and no timelock in front of it.

### 5. Good calls

- Holding the #15/#16 removal and the VL#26 "planned/interim" language until the implementation is live is the right sequencing.
- The VL#26 rewrite accurately reflects what #306 actually does, including correcting the earlier owner-supplied-minimums description to the soft-priced exit floor, and the "accepted residuals" section now names both liveness cases rather than one.
- The README's storage-safety note (function bodies + one `error`, no layout change, do not re-run `deploy_01`/`deploy_03`) is exactly the detail that gets lost between PRs.

DavidMinarsch:
## Still open — no response on this branch

Same note as #306: this PR hasn't been touched since my review on 12 Aug (no commits, no reply), while the Dispenser stack has had its round addressed. Two of the three items are cheap; re-raising them so they don't get lost when #306 moves.

### 1. The body and the diff still disagree about the runbook

"Held back (not in this PR)" states the full migration runbook is *intentionally not published* because the draft "describes the issue as open/unfixed with bounty-dispute and exposure detail". The diff adds `docs/liquidity_migration_runbook.md` — 251 lines.

I read it: the content is clean and operational, and carries none of that material, so this reads as a stale body after the reframe landed rather than a disclosure incident. But the body is what a reviewer approves against, and this is a public repo — on a disclosure-sequenced PR, "this file is withheld" is the one claim that has to match the diff. A body edit closes it.

### 2. The upgrade branch skips the ownership handover

Runbook §1.1 step 4 reads:

> **Deploy `LiquidityManagerETH` + `LiquidityManagerProxy`; set Timelock as owner** (or upgrade the existing proxy to the fixed impl via `changeImplementation`, deployment README)

The parenthetical presents the upgrade as an alternative *to* the ownership transfer — and the upgrade branch is the one that applies here, since the proxies already exist and are EOA-owned. Steps 5–6 then transfer LP and seed POL. Followed literally, POL lands in a proxy still owned by the deployer EOA. §1.2 step 5 has the same shape for the L2s.

That contradicts the runbook's own safety framing at lines 93-95 ("the owner is the L2 `BridgeMediator`") and undercuts the "no funds at risk" argument in VL#26, which rests on the LiquidityManager being unseeded and non-custodial. The EOA-ownership decision is defensible precisely *because* POL is not operational — so the transition out of that state is what needs the gate. Making the handover an explicit precondition on the seed step, in both the runbook and `scripts/deployment/pol/README.md`, closes it.

### 3. `script_05` reports success on a failed upgrade

```bash
result=$($castCmd)
echo "$result" | grep "status"
```

Never inspects the receipt, always exits 0, and never reads back `PROXY_LIQUIDITY_MANAGER` to confirm the swap landed. For a single-signer EOA `cast send` with no proposal review in front of it, that post-condition check is the only safety net.

Worth noting the Dispenser stack already moved off this exact pattern in response to the same finding — `deploy_23_vote_weighting.sh` (governance) asserts its immutable slots and exits non-zero on mismatch, and `cfceafb` (#314) hardened the impl-lock to assert its receipt rather than grep it. Same two-line change here.

kupermind:
Addressed:
- **①** Body corrected — the migration runbook **is** published and reframed (clean, no bounty/exposure material); the stale "withheld" claim is fixed.
- **②** You are right, and it aligns with the script header's own rationale ("ownership left with the deployer *while POL is not operational*"): seeding makes the LM custodial, so ownership must be the Timelock / `BridgeMediator` **before** the first seed. Made that an explicit precondition on the seed step in the runbook (§1.1, §1.2) and the README ordering block (`ffb0cbd`). The upgrade itself stays a plain EOA `changeImplementation`.
- **③** `script_05` now asserts the tx receipt **and** reads back the `keccak256("PROXY_LIQUIDITY_MANAGER")` slot vs `liquidityManagerAddress` (exit non-zero on mismatch), plus zero-address guards — same pattern as governance `deploy_23` / tokenomics `cfceafb`.
- **④** Accepting the EOA-`changeImplementation` `extcodesize` gap as convention (as on Tokenomics/Dispenser). Deliberately *not* adding the codesize check: it is a known false-positive source that would brick legitimate EOA testing / redeployment from the single-signer path, which is exactly the path in use here.

DavidMinarsch (changes requested):
## Re-review

Nothing has changed on this branch since 12 Aug, so the three items below still stand. Restating them briefly with what has moved around them.

### 1. Body and diff still disagree about the runbook

"Held back (not in this PR)" says the migration runbook is intentionally unpublished because the draft carries bounty-dispute and exposure detail. The diff adds `docs/liquidity_migration_runbook.md`, 251 lines. I have read it — the content is clean and operational and carries none of that material, so this reads as a body that was never updated after the reframe landed. On a disclosure-sequenced PR in a public repo, that is the one claim that has to match the diff. A body edit closes it.

### 2. The upgrade branch still skips the ownership handover

Runbook §1.1 step 4 presents `changeImplementation` as an alternative *to* setting the Timelock as owner, and that is the branch that applies here — the proxies already exist and are EOA-owned. Steps 5–6 then transfer LP and seed POL, so followed literally the POL lands in a proxy still owned by the deployer EOA. §1.2 step 5 has the same shape for the L2s, and it contradicts the runbook's own safety framing at lines 93-95 ("the owner is the L2 `BridgeMediator`") plus the "no funds at risk" argument in VL#26, which rests on the LiquidityManager being unseeded. Making the handover an explicit precondition on the seed step — in both the runbook and `scripts/deployment/pol/README.md` — closes it.

### 3. `script_05` still reports success on a failed upgrade

`result=$($castCmd); echo "$result" | grep "status"` never inspects the receipt, always exits 0, and never reads back `PROXY_LIQUIDITY_MANAGER`. For a single-signer EOA `cast send` with no proposal review in front of it, that post-condition check is the only safety net there is.

Worth noting the rest of the codebase has now moved past this pattern in response to the same finding: `deploy_23_vote_weighting.sh` (governance, merged) asserts its immutable slots and exits non-zero on mismatch; `cfceafb` (tokenomics #314, merged) hardened the impl-lock to assert its receipt; and `script_dispenser_change_managers.sh` reads back `voteWeighting()` and fails on mismatch. This script is the last one still grepping.

### Cross-reference to #306

Two things there bear on this PR:

- The internal audit in #306 lands at `audits/internal18/`, which is already occupied on `main` by the Dispenser rework audit (merged via #312). That is #306's merge conflict; it needs renumbering to `audits/internal20/`. Since this PR is the documentation companion, worth confirming nothing here links to `audits/internal18` for the LiquidityManager audit — I grepped and found no reference, but the audit-diff doc is the likely place for one to appear later.
- R1 from that audit — size `observationCardinality` to span 1800 s at peak churn for each POL pool, and assert it — is a pre-deploy check that belongs in this PR's runbook rather than in code. The README's pre-warm block (step 4, "confirm off-chain that `observe([1800, 0])` succeeds") is close but checks the current buffer rather than that cardinality is sized for peak activity. Those differ precisely when it matters: a busy pool can satisfy the check today and wrap below 1800 s during the next volatility burst, which is the state that makes `_getExitSqrtPrice` fail open.

DavidMinarsch:
## Re: `ffb0cbd` + `f9d82f0` — two of three closed

Verified both on the current head.

**#307.3 — `script_05` now asserts.** It parses the receipt status, fails on `0 (failed)` or a missing status line, and reads back the implementation from `keccak256("PROXY_LIQUIDITY_MANAGER")` with a non-zero exit on mismatch. That closes it — and it was the last script in either repo still grepping rather than asserting.

**#307.2 — ownership handover is now a precondition.** README step 5 ("hand proxy ownership to the Timelock / `BridgeMediator`") with the reasoning stated inline — the EOA ownership is justified only while POL is non-operational, and seeding is what makes the LM custodial. That is exactly the gate that was missing, and putting the *why* next to it is what stops it being dropped as ceremony later.

**#307.1 — still open.** The PR body still carries the "Held back (not in this PR)" section stating the migration runbook is intentionally unpublished, while the diff adds all 251 lines of `docs/liquidity_migration_runbook.md`. The file content is fine; it is the body that needs the edit. Small, but on a disclosure-sequenced PR in a public repo it is the one claim that has to match the diff, and it is what a reviewer approves against.

### One item arriving from #306

The exit fail-open PoC on #306 proves the *inactive-pool* branch is self-defeating (the attacker's own swap re-activates the pool and re-engages the deviation gate). It does not cover the second fail-open trigger — `observe(1800)` reverting because the observation buffer no longer spans 1800 s — and the re-activation argument does not rescue that one, since writing an observation shortens the buffer's span rather than extending it.

That makes audit finding **R1** load-bearing for this PR specifically: the README's pre-warm block currently says to "confirm off-chain that `observe([1800, 0])` succeeds", which tests the buffer *today*. R1 asks for something stronger — that `observationCardinality` is sized to span 1800 s at each pool's **peak** churn. Those come apart exactly when it matters: a pool can pass the current check, then wrap below 1800 s during the next volatility burst, which is the state in which the exit gate silently switches off. Worth promoting to a hard release gate with a per-pool number rather than a confirm-once step.

kupermind:
## Addressed 1–3 + the cross-PR R1; 4 declined with rationale

Sorry for the silence on this branch while the #306 proofs were in flight — caught up now (`346e3c6`).

**1. [High] body vs. runbook disagreement — corrected.** The body no longer claims the runbook is withheld: the "Held back" section now lists only the `#15`/`#16` removal and the VL#26 "planned/interim" language, and a new "Migration runbook — included and reframed" section states the runbook **is** published here (operational procedure, no bounty/exposure material) and explicitly corrects the earlier withheld claim. So the section a reviewer approves against now matches the diff.

**2. [Medium] ownership handover as a seed precondition — done.** Made explicit in both places:
- Runbook §1.1 / §1.2 now carry the handover as a precondition on the seed step (not an alternative to it).
- `scripts/deployment/pol/README.md` "Ordering — upgrade before seeding" block lists "ownership moved to Timelock / BridgeMediator" alongside the pre-warm preconditions.

**3. [Medium] `script_05` reports success on a failed upgrade — done.** The script now asserts the receipt `status`, reads back the `PROXY_LIQUIDITY_MANAGER` slot via `cast storage` and compares it to `liquidityManagerAddress`, and exits non-zero on receipt failure or slot mismatch. The null/empty globals guards were extended to the literal zero address, matching the Dispenser proxy script.

**4. [Low] `changeImplementation` accepts an EOA — not changing here.** Agreed it's real, but as noted it's a shared convention across `Tokenomics`, `Dispenser`, and this contract rather than a regression in this PR — so adding a code-size check only here would diverge one of three sibling upgrade paths. Tracking it as a cross-contract consistency item rather than folding it into the price-guard PR. (The single-EOA-signer acuity is now mitigated on the process side by the #3 read-back and the #2 handover precondition.)

### Cross-PR R1 (arriving from #306): cardinality sizing promoted to a hard release gate

The point that the pre-warm check "tests the buffer today" rather than guaranteeing it under churn is right, and it's load-bearing for a reason worth spelling out: when the observation buffer wraps below 1800s, `observe(1800)` reverts and `_getExitSqrtPrice` fails open to raw `slot0` — i.e. the **exit** deviation gate silently switches off. Unlike the inactive-pool fail-open, a front-run swap does **not** rescue this (writing an observation shortens the buffer's span). I've added a fork proof of exactly this branch on #306 (reachable with the gate off; no value extraction on a liquidity removal, but the gate should still stay on).

So the pre-warm block in both the README and runbook §3.1 now makes **sizing `observationCardinality` for PEAK churn a hard, per-pool release gate** (`N >= ceil(1800 / min_block_time) x peak_ticks_crossed_per_block`, rounded up) rather than a confirm-once step, with the exit-gate-disabling consequence spelled out next to the check.

DavidMinarsch:
## Re-review — `346e3c6`

R1 is now genuinely a release gate. The commit does the thing I was asking for and then goes further: it states *why* the buffer-wrap branch is different from the inactive-pool one — a front-run swap re-activating the pool does not rescue it, because writing an observation shortens the buffer's span rather than extending it — and gives a per-pool sizing formula rather than a "confirm it works today" step. The distinction between the mint failing closed (safe) and the exit gate failing open (not safe) under the same undersized buffer is exactly the asymmetry that makes this worth a hard gate.

With #306's buffer-wrap proof now showing no value loss on that branch either (an LP basket burned at a manipulated price is worth *more* at the true price), R1 is defense-in-depth rather than a loss fix — but it is still the right call to gate on it, since it keeps the deviation gate engaged instead of relying on that convexity argument holding under every future change to the exit path.

**#307.2** (ownership handover as a numbered precondition) and **#307.3** (`script_05` asserting the receipt and reading back the implementation slot) both verified closed on this head.

### Still open — the body

The one remaining item is unchanged: the PR body still carries the "Held back (not in this PR)" section saying the migration runbook is intentionally unpublished, while the diff adds all 251 lines of `docs/liquidity_migration_runbook.md`. The file content is fine — it is the body that needs the edit. Once that is corrected this is good to go from my side.

kupermind:
## Body #1 — the runbook is not in the "Held back" list (corrected earlier; now made unmissable)

I think a cached render of the body is in the way here — the "Held back" edit actually landed at **11:11** today, before both of your notes on this, so the version you're quoting predates the fix. The current body already:

- lists **only** the `#15`/`#16` removal + the VL#26 "planned/interim" language under "Held back (not in this PR)" — the runbook is **not** in that list; and
- carries a dedicated section stating `docs/liquidity_migration_runbook.md` **is** published here (reframed, no bounty/exposure material), explicitly correcting the earlier withheld claim.

To make sure it can't be missed on a skim (and to bump the edit timestamp so a reload serves fresh content), I've just re-edited the body (**15:38**) so the exact section you're quoting now reads:

> ## Held back (not in this PR)
> - The **`#15`/`#16` removal** and dropping VL#26's "planned/interim" language wait for the deployment PR …
> - *(Not the runbook — see the section above; it is included here.)*

plus a top-level **"The migration runbook IS in this PR"** section above it. A hard-refresh should show it. If you're still seeing the old "runbook is withheld" wording after reloading, let me know and I'll screenshot the live body — but the diff and the body now agree.

DavidMinarsch (approved):
Approving. And #307.1 was my error, not a caching issue — worth being precise about so it doesn't recur: I checked with `grep -c "Held back"` against the body, which matches the `## Held back (not in this PR)` **section header** that legitimately remains for the `#15`/`#16` and VL#26 items. It never tested the actual claim. The body has read correctly since 11:11; I reported it open twice off a bad check. The current body carries "The migration runbook IS in this PR" above a "Held back" list that explicitly excludes it — body and diff agree.

Everything else verified on this head:

- **#307.2 — ownership handover.** A precondition on the seed step in runbook §1.1/§1.2 and in the README ordering block, with the rationale inline (EOA ownership is justified only while POL is non-operational; seeding is what makes the LM custodial). That reasoning next to the step is what stops it being dropped as ceremony later.
- **#307.3 — `script_05`.** Asserts the receipt status, reads back `keccak256("PROXY_LIQUIDITY_MANAGER")` via `cast storage` against `liquidityManagerAddress`, exits non-zero on either failure, and the globals guards now cover the literal zero address. Last script in either repo still grepping rather than asserting — now closed.
- **R1 — cardinality sizing.** Promoted to a hard per-pool release gate with a sizing formula, and the exit-gate-disabling consequence spelled out next to the check. The `observe(1800)` wrap is the one fail-open branch a front-run swap does not disarm, so gating on it rather than confirming it once is the right call.

**#307.4 — accepting the decline.** The consistency argument carries it: adding a code-size check to one of three sibling upgrade paths (`Tokenomics`, `Dispenser`, `LiquidityManagerCore`) would diverge them, and it is already recorded as accepted risk on #309. Tracking it as a cross-contract item is the right home. I would push back on the second half of the rationale though — a codesize check would not "brick legitimate EOA testing", since pointing a proxy at an EOA is never a legitimate target; the reason to defer is consistency, and the process mitigations (#3's read-back, #2's handover) are what actually reduce the acuity here.

Scope confirmed docs/scripts only — `git diff origin/main...HEAD -- contracts/` is empty, so #306's merge doesn't interact with this.

One mechanical note: the branch is behind `main` after `a6091f5`. Worth updating before merge so the runbook and audit-diff land against the shipped `_slot0MinAmounts` code rather than the pre-fix tree.

### Post-merge, not blocking

`Vulnerabilities_list_tokenomics.md` #15/#16/#26 stay open by design until the implementation is live on the proxies. That is correct sequencing, but it does mean `main` now carries the fixed price guard while the public list still describes fail-open behaviour. The window closes when `script_05` actually rolls it — worth keeping the two steps adjacent on the release checklist so the list doesn't sit stale after the deploy.

77ph:
@kupermind @DavidMinarsch — companion to the #306 re-review. The deploy/ownership/script surface all checks out on this head; one doc-reconciliation is still outstanding.

### Confirmed
- **`script_05` (`f9d82f0`)** — the receipt-status check plus the read-back of `keccak256("PROXY_LIQUIDITY_MANAGER")` (`0xf7d1f641…`) against `liquidityManagerAddress` is correct: that constant is exactly the slot `LiquidityManagerProxy` `sstore`/`sload`s, and the last-20-bytes extraction is the right decode. For a single-signer `changeImplementation` this post-condition is the whole safety net, and it's sound.
- **Ownership-before-seed (`ffb0cbd`)** — making the Timelock / `BridgeMediator` handover a precondition on the *seed* step (not an alternative to the upgrade) is the right gate, and stating the *why* inline keeps it from being dropped as ceremony later.
- **R1 cardinality gate (`346e3c6`)** — agree it should be a hard, per-pool release gate, and the buffer-wrap-vs-inactive asymmetry (a front-run swap *shortens* the buffer's span, so it doesn't self-heal) is the right reason. One wording note: since the #306.2 buffer-wrap proof shows the exit stays loss-free even with the gate off (LP-burn convexity — the basket is worth *more* at the true price), this is defense-in-depth rather than a fund-loss fix, so "a sandwich can move the exit price → exploitable" slightly overstates it. An operator who under-sizes loses the belt-and-suspenders layer, not funds. Worth keeping the gate; just aligning the prose with the convexity result would make the risk model exact.

### One doc-reconciliation still outstanding
The internal20 write-up was reconciled to the shipped slot0 anchor (`6885876`), but three sibling docs in this PR still describe the entry `amountMin` as the pre-#306.1 **TWAP-center** design. Since this PR's audit-diff is explicitly the map "a reviewer traces each change" against, it's worth bringing them in line with the deployed code before merge:

1. **`docs/lm_price_guard_audit_diff.md`** — the `[FIX-2a]` legend ("increase TWAP anchor") and its row ("New state … `_increaseLiquidity` anchors `amountMin` to the TWAP center") describe the interim design. The shipped code anchors the entry floor to **slot0** via `_slot0MinAmounts` — so this row inverts the single most security-relevant change (post-fix the amount floor is vacuous by construction; the 10% deviation gate is the sole entry defence — internal20 R6).
2. **`docs/Vulnerabilities_list_tokenomics.md` §26 step (3)** — "*Anchors the entry slippage floor to a trusted price — `_increaseLiquidity`'s `amountMin` is derived from the TWAP center rather than raw slot0*" is the opposite of what shipped. (§26 also still reads "code fix planned" though #306 is merged — worth flipping to fixed/closed.)
3. **`docs/Vulnerabilities_list_tokenomics.md` §16** — this one *correctly* describes the current code as slot0-anchored, but its recommended "architectural fix — use the TWAP-derived center price as the anchor for `amountsMin`" is exactly the dead band #306.1 established and removed. As written, §16 and §26 point in opposite directions. A one-line note that #306.1 made the slot0 anchor intentional (with the deviation gate as the manipulation defence, R6) reconciles them.

None of this changes the code — it's making the reviewer-facing docs match the deployed slot0 anchor, the same reconciliation already applied to internal20. Happy to re-check once it lands.

— 77ph

PR #308, opened 2026-07-07
audit(pol): internal18 report + price-guard regression PoCs for #306

77ph:
Stacks on `fix/lm-price-guard-failclosed` (#306) so the fix, its regression tests, and the internal audit record land together. Diff here is audit-only (3 files); merging this into the fix branch makes #306 carry everything.

**Contents**
- `audits/internal18/README.md` — internal audit report. Verdict **PASS** (0 Critical / 0 High / 0 Medium); 5 findings (R1–R5), all Info/Low, none blocking. Includes a final-in-class completeness argument (two guarded price sources, single `slot0` reader) and a no-new-bug delta proof.
- `test/LiquidityManagerPriceGuardRegression.t.sol` — non-fork regression suite (I1–I5): entry fails closed on fresh/stale pools, entry deviation gate reverts, mature manipulated exit reverts, quiet pool stays always-exitable. Turns red if the fail-open class re-opens.
- `test/LiquidityManagerExitSandwichFork.t.sol` — mainnet-fork PoC: a manipulated `slot0` exit reverts while an honest exit succeeds (requires an `ETH_RPC` fork URL).

**Findings:** R1 cardinality sizing (Low), R2 restore the skipped legacy test (Low), R3 doc signature nit (Info), R4 `applyDeviationGate` caller-precondition trap (Info, robustness), R5 guard duplication (Info, maintainability). Details in the report.

**CI note:** the fork PoC needs a mainnet `ETH_RPC` secret; if CI has none it should be conditionally skipped so it does not block — the non-fork regression runs everywhere.

### DIFF
--- a/audits/internal18/README.md
+++ b/audits/internal18/README.md
@@ -1,0 +1,220 @@
+# Internal Audit 18 — LiquidityManagerCore price-guard fail-closed remediation
+
+**Scope:** PR #306 (`fix/lm-price-guard-failclosed`) + PR #307 (`docs/lm-guard-deployment-runbook`),
+base `main` (`029b5574`). Reviewed file: `contracts/pol/LiquidityManagerCore.sol` (production) and the
+accompanying deployment runbook / audit-diff (docs). POL subsystem: `convertToV3`, `increaseLiquidity`,
+`decreaseLiquidity`, `changeRanges`, `collectFees`, and the price-guard helpers
+`checkPoolAndGetCenterPrice` / `getTwapFromOracle` / `_getExitSqrtPrice`, plus the permissionless
+`BuyBackBurner.buyBack` V3 path that reuses the same guard.
+
+**Verdict: PASS — 0 Critical / 0 High / 0 Medium.** 5 findings, all Info/Low, none blocking (R1–R5).
+The remediation correctly converts the price guard from *fail-open* to *fail-closed* on entries and to a
+*deviation-gated soft floor* on exits, closes the fee-collection scope-burn, and is storage-layout safe
+for a `changeImplementation` upgrade.
+
+---
+
+## 1. What the change does
+
+The pre-fix guard could, on a pool with **no verifiable 30-minute TWAP** (freshly-created /
+`observationCardinality <= 1`, or inactive for more than `SECONDS_AGO = 1800 s`), fall back to the raw,
+single-block-manipulable `slot0` price and size/price a value-bearing operation from it (fail-open). The
+remediation:
+
+| Change | Effect |
+|---|---|
+| **FIX-1** — `checkPoolAndGetCenterPrice` fail-closed | On an unverifiable pool the entry path now **reverts** (`NotEnoughHistory`) instead of returning `slot0`; when a TWAP exists it reverts if `|slot0 − TWAP| / TWAP > MAX_ALLOWED_DEVIATION` (10%) and returns the **TWAP-derived** price for minting. |
+| **FIX-2a** — `_increaseLiquidity` anchor | `amountMin` and the liquidity math are anchored to the caller's TWAP center (passed as a typed parameter), not the manipulable `slot0`. |
+| **FIX-2b** — `_getExitSqrtPrice` soft floor | New exit/maintenance price source: reverts on `slot0`-vs-TWAP deviation `>10%` on a **verifiable** pool, else skips the gate (fail-open) so a withdrawal is **always possible** on a quiet pool; returns raw `slot0` (the exact price the position manager withdraws at). |
+| **FIX-3** — `collectFees` scope-burn | New `_manageCollectedAmounts` burns/transfers only the **just-collected** fees, not `balanceOf(this)`, so separately-staged OLAS is no longer griefable and the donation-inflation vector is closed; the guard is removed from `collectFees` (fee collection consumes no price and must stay live). |
+| **FIX-5** — cleanup | Rename `oldestTimestamp -> latestObsTimestamp` (the value read is the *latest* observation, not the oldest — the old name was a misnomer) and removal of the now-unused `_getObservationCardinality` helper. |
+
+Both halves were verified first-hand on a mainnet fork (see §6).
+
+---
+
+## 2. Completeness — the fix is final in its class
+
+**Class C = "unverifiable-pool price fail-open".** A bug is in C iff it lets a value-bearing operation be
+sized/priced from a pool price the contract cannot verify, such that an attacker who moves that `slot0`
+shifts the operation in their favour.
+
+**Surface closure (why the enumeration is exhaustive).** Every value-bearing op that touches a pool price
+does so through exactly one of **two** internal price sources — entry/trade via `checkPoolAndGetCenterPrice`,
+exit/maintenance via `_getExitSqrtPrice` (or its gate-skip twin inside `_decreaseLiquidity`, which is
+dominated by a preceding `checkPoolAndGetCenterPrice` in `changeRanges`). The only `slot0` reader is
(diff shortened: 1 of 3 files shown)


kupermind:
## Review of #308 (internal18 audit + regression PoCs)

Reviewed the audit report and both PoC suites against the actual `LiquidityManagerCore` code on the base branch, and ran what I could locally.

**Verdict: strong and mergeable with a few wiring/hygiene fixes.** The report's reasoning is sound, the completeness/no-new-bug arguments hold up against the code, and both PoCs are technically correct. My findings below are mostly about how the tests are wired into the repo, not about their correctness.

### What I independently verified

| Check | Result |
|---|---|
| `LiquidityManagerPriceGuardRegression` (non-fork, I1–I5) | **5/5 PASS** locally (`forge test --mc LiquidityManagerPriceGuardRegression`) |
| `LiquidityManagerExitSandwichFork::test_L1_...` on a live ETH fork | **PASS** (manipulated `slot0` exit reverts `Overflow`; honest exit succeeds) — confirms L1 empirically |
| Mock pool vs `checkPoolAndGetCenterPrice` / `_getExitSqrtPrice` / `getTwapFromOracle` decode paths | consistent (7-field `slot0`, `observationIndex` from word 2, flat `observe` → TWAP tick 0) |
| R3 (doc types `checkPoolAndGetCenterPrice` as `internal`; code is `public`) | confirmed — it is `public` (must be, `BuyBackBurner` calls it) |

### Findings to address in this PR

**1. (must-fix) The regression suite is not wired into CI, which contradicts the file's own header.** The file states it "must live in the project CI (foundry non-fork suite) and stay GREEN," but CI's foundry step (`.github/workflows/workflow.yaml:58`) runs a hardcoded `--mc` allowlist that does not include `LiquidityManagerPriceGuardRegression`. As-is, the "permanent certified gate" never runs in CI and can silently rot. Fix: append `&& forge test --mc LiquidityManagerPriceGuardRegression` to that line (it's pure-mock, runs offline). The fork PoC should **not** be added there — no RPC in CI.

**2. (should-fix) The fork PoC inherits the concrete `LiquidityManagerETHTest`, so it re-declares and re-runs the entire ETH fork suite under a second contract.** `forge test --list --mc LiquidityManagerExitSandwichFork` shows **26 tests** in the child — its own L1 test plus all 25 inherited ETH fork tests. Any run that matches `--mc LiquidityManager` (or the full fork set) executes the whole ETH suite a second time — expensive and confusing. Fix: inherit `BaseSetup` instead of `LiquidityManagerETHTest`. Same `import`, identical setUp (the `LiquidityManagerETHTest` override is just `super.setUp()`), and `BaseSetup` already exposes everything the test uses (`TOKENS`, `FEE_TIER`, `FACTORY_V3`, `liquidityManager`, `convertToV3`). I re-ran the L1 test after this change mentally against the setUp path — behavior is preserved.

**3. (should-fix) The fork PoC has no RPC-absent guard.** The PR body says it "should be conditionally skipped" if there's no `ETH_RPC`, but the code doesn't implement that — it's saved from breaking CI only because CI's allowlist happens to exclude it. Anyone running `forge test` unfiltered without `--fork-url` hits a hard setUp failure. Add a self-skip at the top, e.g. `vm.skip(bytes(vm.envOr("ETH_RPC", string(""))).length == 0);`. Note the header documents the run as `--fork-url $ETH_RPC` but nothing reads that env — the fork is only ever supplied via CLI, so either implement the env read + skip or align the doc.

**4. (nit, strengthens the gate) Bare `vm.expectRevert()` in the regression suite.** For a suite whose whole job is to catch a *specific* regression class, an untyped `expectRevert` also passes if the call reverts for the wrong reason — e.g. I3 would still "pass" if it reverted `NotEnoughHistory` or `ZeroValue` instead of the deviation `Overflow`. Use typed selectors: `NotEnoughHistory.selector` for I1/I2, `Overflow.selector` (ideally `abi.encodeWithSelector(Overflow.selector, deviation, MAX_ALLOWED_DEVIATION)`) for I3/I4, so a regression that reverts differently is caught rather than masked.

**5. (nit) Scope line.** The report scopes "#306 + #307, base `main (029b5574)`," but this branch stacks only on #306 and carries no #307 files. Worth a one-line note that the #307 docs are reviewed-but-not-carried here.

### Agreement / positives

- R4 and R5 are good calls. Agree with the R4 recommendation to delete the `applyDeviationGate` boolean and always run `_getExitSqrtPrice` in `_decreaseLiquidity` — `changeRanges` keeps its own preceding fail-closed `checkPoolAndGetCenterPrice`, so its strict posture is unchanged and the now-always-on gate is only a redundant `observe` (gas, not risk). Fail-safe-defaults over an unenforced caller precondition. R5's "extract the policy-free compute, keep policy explicit at each callsite" is the right de-dup (not a flag).
- The completeness map (two guarded price sources, single `slot0` reader `_getPriceAndObservationIndexFromSlot0` called only from those two) matches the code — the enumeration is genuinely exhaustive.
- No private identifiers leak in the report body — clean for a public repo.

### Merge recommendation

Findings 1–3 are worth fixing before merge (1 especially — otherwise the certified gate isn't actually a gate). 4–5 are nits. R1/R2/R4/R5 are dev-side follow-ups on #306, not blockers for #308. With 1–3 addressed I'd merge this into `fix/lm-price-guard-failclosed` so the fix, its regression, and the audit record land together.

77ph:
Thanks for the thorough review — re-checked your notes and the follow-up commit (`a56ab8a`) against the code. Agreed on all points; summary of where each lands:

**PoC-wiring items 1–4 — closed by `a56ab8a`, verified by running them:**
- (1) regression suite is now in the CI `--mc` allowlist — the three names resolve to real contracts and run 11 / 2 / 5 = 18 tests green (no hollow filter).
- (2) the fork PoC now inherits `BaseSetup` instead of the concrete ETH test — the duplicate ~26-test re-run is gone, PoC compiles clean and references nothing from the concrete class; the original ETH suite still runs under its own name.
- (3) fork PoC now self-skips off-fork. I went with a `block.chainid != 1` guard rather than an `ETH_RPC`-env check — it's robust regardless of how the fork URL is supplied (off-fork → 1 skipped, no error; on an ETH fork → runs). Flagging the deviation from your suggestion in case you'd prefer the env form.
- (4) typed reverts in place. `NotEnoughHistory` gets the full `abi.encodeWithSelector(..., pool)` match you asked for. For the deviation cases I used `expectPartialRevert(Overflow.selector)` instead of a full `encodeWithSelector(Overflow, deviation, MAX_ALLOWED_DEVIATION)` — the `deviation` arg is price-dependent so it can't be hard-coded, but it still pins the revert to `Overflow` and rejects any wrong-reason revert. Suite stays 5/5 green.

**Open nit (5):** the internal18 scope line reads "#306 + #307" though this branch carries only the #306 contract change — cosmetic, noted; the audit is closed so we'll leave the report as-is.

**R4 / R5 — agreed, and understood as dev-side follow-ups (not #308 blockers):** dropping the `applyDeviationGate` boolean and always running the exit-price compute in `_decreaseLiquidity` (fail-safe default rather than an unenforced caller precondition — `changeRanges` keeps its own preceding fail-closed check, so its posture is unchanged), plus extracting the policy-free compute and keeping the policy explicit per call-site. When that lands as a commit on #306 we'll re-review that delta specifically.

**Net:** the contract itself is byte-identical to the internal18-reviewed state (`0c8257d` → `a56ab8a` diff on `LiquidityManagerCore.sol` is empty), so internal18 stands as-is — everything after the audit is test/CI hardening of the regression net, which strengthens it. No security items open across #306 / #307 / #308; the only follow-ups are the report scope-line note above and the R4/R5 contract change when you make it.

PR #309, opened 2026-07-23
refactor(Dispenser): proxy-based Dispenser behind DispenserProxy

kupermind:
First PR of the Dispenser rework (PR-A of the series): makes the Dispenser **proxy-based**, following the `TokenomicsProxy` / `LiquidityManagerProxy` pattern, so future Dispenser logic fixes become one-tx implementation swaps behind a **stable address** instead of a whole-stack redeploy. The L1 deposit processors bind `l1Dispenser` immutably and the L2 target dispensers bind their L1 processor immutably — with a stable proxy address those bindings never have to change again.

**Contents**
- `contracts/proxies/DispenserProxy.sol` — EIP-1822-style proxy: implementation slot `keccak256("PROXY_DISPENSER")`, `constructor(implementation, initData)` delegatecall-init, payable Gnosis-Safe-style fallback (the Dispenser has payable claim functions).
- `contracts/Dispenser.sol`:
  - Constructor now sets **bytecode immutables only**: `olas`, `tokenomics`, `retainer`/`retainerHash`, `defaultMinStakingWeight`, `defaultMaxStakingIncentive`. Behind a proxy these are re-settable via an implementation swap, so "immutable" no longer means locked forever — every impl deploy must pass the correct constructor args (documented in-code).
  - `tokenomics` promoted to immutable — it points at the stable TokenomicsProxy and is read on every claim (hot path, zero-SLOAD). `changeManagers` drops its tokenomics param; `treasury` / `voteWeighting` stay mutable.
  - New `initialize(treasury, voteWeighting, maxNumClaimingEpochs, maxNumStakingTargets)` guarded by `AlreadyInitialized`; new owner-gated `changeImplementation(address)` (same shape/name as LiquidityManagerCore / BuyBackBurner).
  - Storage layout frozen append-only (no gap — flat contract, Tokenomics convention).
- Tests: new `test/DispenserProxy.t.sol` (9 tests — proxy guards, init/re-init, upgrade preserving state while swapping logic/immutables); `test/Dispenser.t.sol` + both hardhat Dispenser suites now run **through the proxy**, with deploy order inverted (Tokenomics proxy → Dispenser impl+proxy → repoint) since `tokenomics` is a constructor immutable.

**No behavior change** to claim/nominee/withheld logic in this PR — bug fixes (vuln-list #8, #9, #12, #25) land as the follow-up PR on top of this base; the L2 `migrate()` fix (#10) and deploy scripts + fork tests follow separately.

**Verified:** forge Dispenser 3/3 + DispenserProxy 9/9 + Depository/Treasury green; full hardhat run green (exit 0); eslint/solhint clean on changed files. CI's `--mc Dispenser` filter picks up the new proxy suite automatically.

### DIFF
--- a/contracts/Dispenser.sol
+++ b/contracts/Dispenser.sol
@@ -228,6 +228,9 @@ error WrongAmount(uint256 provided, uint256 expected);
 /// @dev Caught reentrancy violation.
 error ReentrancyGuard();
 
+/// @dev The contract is already initialized.
+error AlreadyInitialized();
+
 /// @dev The contract is paused.
 error Paused();
 
@@ -263,7 +266,7 @@ contract Dispenser {
     }
 
     event OwnerUpdated(address indexed owner);
-    event TokenomicsUpdated(address indexed tokenomics);
+    event ImplementationUpdated(address indexed implementation);
     event TreasuryUpdated(address indexed treasury);
     event VoteWeightingUpdated(address indexed voteWeighting);
     event StakingParamsUpdated(uint256 maxNumClaimingEpochs, uint256 maxNumStakingTargets);
@@ -280,16 +283,28 @@ contract Dispenser {
     event AddNomineeHash(bytes32 indexed nomineeHash);
     event RemoveNomineeHash(bytes32 indexed nomineeHash);
 
+    // Dispenser proxy address slot
+    // keccak256("PROXY_DISPENSER") = "0x8bd249c73459f2c50400ebdc57436101fc7d9a76908baf1ba5be362b47b48f83"
+    bytes32 public constant PROXY_DISPENSER = 0x8bd249c73459f2c50400ebdc57436101fc7d9a76908baf1ba5be362b47b48f83;
     // Maximum chain Id as per EVM specs
     uint256 public constant MAX_EVM_CHAIN_ID = type(uint64).max / 2 - 36;
+
+    // Immutables live in the implementation bytecode, not proxy storage: they are re-settable by deploying
+    // a new implementation and calling changeImplementation. Every new implementation deploy must pass the
+    // correct constructor args — a wrong value silently rewires the proxy.
     uint256 public immutable defaultMinStakingWeight;
     uint256 public immutable defaultMaxStakingIncentive;
     // OLAS token address
     address public immutable olas;
+    // Tokenomics proxy address (stable by construction — the proxy address never changes)
+    address public immutable tokenomics;
     // Retainer address in bytes32 form
     bytes32 public immutable retainer;
     // Retainer hash of a Nominee struct composed of retainer address with block.chainid
     bytes32 public immutable retainerHash;
+
+    // Storage layout below is consumed via the DispenserProxy delegatecall and is FROZEN: never remove,
+    // reorder, retype or insert in the middle — new variables are appended strictly after the last mapping.
 
     // Max number of epochs to claim staking incentives for
     uint256 public maxNumClaimingEpochs;
@@ -302,8 +317,6 @@ contract Dispenser {
     // Pause state
     Pause public paused;
 
-    // Tokenomics contract address
-    address public tokenomics;
     // Treasury contract address
     address public treasury;
     // Vote Weighting contract address
@@ -320,36 +333,27 @@ contract Dispenser {
     // Mapping for epoch number => refunded all the staking inflation due to zero total voting power
     mapping(uint256 => bool) public mapZeroWeightEpochRefunded;
 
-    /// @dev Dispenser constructor.
+    /// @dev Dispenser implementation constructor: sets the implementation-bytecode immutables only.
+    /// @notice The mutable state is set via initialize() delegatecall-ed by the DispenserProxy constructor.
     /// @param _olas OLAS token address.
-    /// @param _tokenomics Tokenomics address.
-    /// @param _treasury Treasury address.
-    /// @param _voteWeighting Vote Weighting address.
+    /// @param _tokenomics Tokenomics proxy address.
+    /// @param _retainer Retainer address in bytes32 form.
+    /// @param _defaultMinStakingWeight Default min staking weight (bounded by uint16).
+    /// @param _defaultMaxStakingIncentive Default max staking incentive (bounded by uint96).
     constructor(
         address _olas,
         address _tokenomics,
-        address _treasury,
-        address _voteWeighting,
         bytes32 _retainer,
-        uint256 _maxNumClaimingEpochs,
-        uint256 _maxNumStakingTargets,
         uint256 _defaultMinStakingWeight,
         uint256 _defaultMaxStakingIncentive
     ) {
-        owner = msg.sender;
-        _locked = 1;
-        // Staking incentives must be paused at the time of deployment because staking parameters are not live yet
-        paused = Pause.StakingIncentivesPaused;
-
         // Check for at least one zero contract address
-        if (_olas == address(0) || _tokenomics == address(0) || _treasury == address(0) ||
-            _voteWeighting == address(0) || _retainer == 0) {
+        if (_olas == address(0) || _tokenomics == address(0) || _retainer == 0) {
             revert ZeroAddress();
         }
 
         // Check for zero value staking parameters
-        if (_maxNumClaimingEpochs == 0 || _maxNumStakingTargets == 0 || _defaultMinStakingWeight == 0 ||
-            _defaultMaxStakingIncentive == 0) {
+        if (_defaultMinStakingWeight == 0 || _defaultMaxStakingIncentive == 0) {
             revert ZeroValue();
         }
 
@@ -363,15 +367,69 @@ contract Dispenser {
 
         olas = _olas;
         tokenomics = _tokenomics;
-        treasury = _treasury;
-        voteWeighting = _voteWeighting;
 
         retainer = _retainer;
         retainerHash = keccak256(abi.encode(IVoteWeighting.Nominee(retainer, block.chainid)));
-        maxNumClaimingEpochs = _maxNumClaimingEpochs;
-        maxNumStakingTargets = _maxNumStakingTargets;
         defaultMinStakingWeight = _defaultMinStakingWeight;
         defaultMaxStakingIncentive = _defaultMaxStakingIncentive;
+    }
+
+    /// @dev Initializes the dispenser proxy storage.
+    /// @notice Delegatecall-ed by the DispenserProxy constructor; the caller becomes the owner.
+    /// @param _treasury Treasury address.
+    /// @param _voteWeighting Vote Weighting address.
+    /// @param _maxNumClaimingEpochs Maximum number of epochs to claim staking incentives for.
+    /// @param _maxNumStakingTargets Maximum number of staking targets on a single chain Id.
+    function initialize(
+        address _treasury,
+        address _voteWeighting,
+        uint256 _maxNumClaimingEpochs,
+        uint256 _maxNumStakingTargets
+    ) external {
+        // Check that the proxy storage has not been initialized yet
+        if (owner != address(0)) {
+            revert AlreadyInitialized();
+        }
+
+        // Check for at least one zero contract address
+        if (_treasury == address(0) || _voteWeighting == address(0)) {
+            revert ZeroAddress();
+        }
+
+        // Check for zero value staking parameters
+        if (_maxNumClaimingEpochs == 0 || _maxNumStakingTargets == 0) {
+            revert ZeroValue();
+        }
+
+        owner = msg.sender;
+        _locked = 1;
+        // Staking incentives must be paused at the time of deployment because staking parameters are not live yet
+        paused = Pause.StakingIncentivesPaused;
+
+        treasury = _treasury;
+        voteWeighting = _voteWeighting;
+        maxNumClaimingEpochs = _maxNumClaimingEpochs;
+        maxNumStakingTargets = _maxNumStakingTargets;
+    }
+
+    /// @dev Changes the dispenser implementation contract address behind the proxy.
+    /// @param implementation Dispenser implementation contract address.
+    function changeImplementation(address implementation) external {
+        // Check for the contract ownership
+        if (msg.sender != owner) {
+            revert OwnerOnly(msg.sender, owner);
+        }
+
+        // Check for the zero address
+        if (implementation == address(0)) {
+            revert ZeroAddress();
+        }
+
+        // Store the implementation address under the designated storage slot
+        assembly {
+            sstore(PROXY_DISPENSER, implementation)
+        }
+        emit ImplementationUpdated(implementation);
     }
 
     /// @dev Checkpoints specified staking target (nominee in Vote Weighting) and gets claimed epoch counters.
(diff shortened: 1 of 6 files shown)


77ph:
Reviewed the full Dispenser-rework stack (**#309 → #310 → #311**) as the release gate for the five `Vulnerabilities_list_tokenomics.md` items **#8 / #9 / #10 / #12 / #25**. Verified firsthand against `main` (`029b557`): built the tree, ran the forge suites, traced the staking-incentive accounting, and swept for regressions.

### This PR (proxy refactor) — behaviour-preserving ✔

- The `main → refactor/dispenser-proxy` diff is **plumbing only** (errors/events/storage decls, `constructor → initialize`, `changeImplementation`, `changeManagers` signature). No `claim` / nominee / `calculateStakingIncentives` / withheld logic is touched — the "no behavior change" claim holds.
- `PROXY_DISPENSER = keccak256("PROXY_DISPENSER") = 0x8bd249…48f83` verified; `changeImplementation` writes that exact slot the fallback reads → the upgrade path is self-consistent; owner-gated + zero-checked.
- `initialize` is re-init-guarded (`owner != 0 → AlreadyInitialized`); `owner` is set to the deployer atomically inside the proxy constructor delegatecall, so there is **no front-run window**. The `DispenserProxy` suite covers proxy re-init **and** implementation-direct re-init reverts.
- Storage is `frozen append-only`; moving `tokenomics` from storage to an implementation immutable shifts slots, which is safe here because this is a **greenfield proxy deployment** (not an upgrade over pre-existing storage). Worth stating explicitly in the deploy runbook so a future maintainer never points the proxy at an old storage image.
- `forge test --mc Dispenser`: **17/17** (DispenserProxy 9, DispenserFixes 5, Dispenser 3).

Two notes, both informational:
- `changeImplementation` asserts non-zero but not is-contract. This matches `LiquidityManagerCore` / `BuyBackBurner`, so it is a consistent, owner-trusted convention — flagging only for the record.
- The standalone implementation can be `initialize`-d by anyone (in its own storage) — the usual UUPS consideration; harmless here since the implementation has no `selfdestruct` / arbitrary `delegatecall`, and the suite already tests the direct-reinit revert.

### Release/deploy gate (applies to the whole series)

The constructor signature changed (5 immutables) and `initialize` is new, so the deploy order inverts (Tokenomics proxy → Dispenser impl+proxy → repoint). The in-code warning that *"a wrong constructor immutable silently rewires the proxy"* makes the deploy arguments safety-critical. The contract changes are sound to merge, but **the on-chain deploy should wait on the planned PR-D — the deploy scripts + an L1 mainnet-fork test of the proxied claim path — since none of that is in this stack yet** and the current forge tests use mocks.

No regression outside the diff: the `Dispenser.changeManagers` signature change (3→2 args) is not called by any deployment/proposal script (those call `changeManagers` on `Tokenomics` / `Treasury`, not the Dispenser).

contracts/Dispenser.sol
```diff
@@ -280,17 +283,29 @@ contract Dispenser {
...
 
+    // Dispenser proxy address slot
+    // keccak256("PROXY_DISPENSER") = "0x8bd249c73459f2c50400ebdc57436101fc7d9a76908baf1ba5be362b47b48f83"
+    bytes32 public constant PROXY_DISPENSER = 0x8bd249c73459f2c50400ebdc57436101fc7d9a76908baf1ba5be362b47b48f83;
     // Maximum chain Id as per EVM specs
     uint256 public constant MAX_EVM_CHAIN_ID = type(uint64).max / 2 - 36;
+
+    // Immutables live in the implementation bytecode, not proxy storage: they are re-settable by deploying
+    // a new implementation and calling changeImplementation. Every new implementation deploy must pass the
+    // correct constructor args — a wrong value silently rewires the proxy.
```
kupermind, line 294: Who are you writing this for? It's a kindergarten knowledge, just post create comments for variables that don't have them

contracts/Dispenser.sol
```diff
@@ -280,17 +283,29 @@ contract Dispenser {
...
     uint256 public immutable defaultMinStakingWeight;
     uint256 public immutable defaultMaxStakingIncentive;
     // OLAS token address
     address public immutable olas;
+    // Tokenomics proxy address (stable by construction — the proxy address never changes)
```
kupermind, line 299: Stop stupid comments - "Tokenomics proxy address" is enough

contracts/Dispenser.sol
```diff
@@ -320,36 +333,27 @@ contract Dispenser {
...
     constructor(
         address _olas,
         address _tokenomics,
-        address _treasury,
-        address _voteWeighting,
         bytes32 _retainer,
-        uint256 _maxNumClaimingEpochs,
-        uint256 _maxNumStakingTargets,
         uint256 _defaultMinStakingWeight,
         uint256 _defaultMaxStakingIncentive
```
DavidMinarsch, line 337: whats the rational for some moving and some not?

DavidMinarsch:
## Review: proxy-based Dispenser (PR-A) — plus stack-level findings

Verified locally on the final stacked state (`refactor/dispenser-view-calculate-staking-incentives`): **forge `--mc Dispenser` 19/19 pass** (Dispenser 3, DispenserProxy 9, DispenserFixes 7). I could not run the Hardhat suite — solmate resolution fails in my clone (submodule artifact, not a defect here) — so the Hardhat pass counts across the series are unverified.

### This PR

Verified rather than assumed: `keccak256("PROXY_DISPENSER")` = `0x8bd249c73459f2c50400ebdc57436101fc7d9a76908baf1ba5be362b47b48f83`, matching the constant in both `DispenserProxy.sol` and `Dispenser.sol`; `changeImplementation` is shape-identical to `Tokenomics.changeTokenomicsImplementation`; the storage layout is append-only; and `owner` (slot 2) is a sound re-init guard precisely because `changeOwner` rejects the zero address, so `owner` can never return to zero.

**[Low] `changeImplementation` accepts an EOA.** A `delegatecall` to an address with no code returns `success = true` with empty returndata, so the proxy fallback would `return(0, 0)` for every call — silently answering everything with nothing, *including* `changeImplementation` itself. That makes the brick permanent. The same gap exists in `Tokenomics`, so this is convention rather than a regression, but one `extcodesize` check is cheap insurance on the contract that now owns the entire staking-incentive flow.

**[Low] The implementation is left uninitialized.** Anyone can call `initialize()` on the impl and become its owner; `test_initialize_implDirectReinit_reverts` documents this as "harmless" rather than preventing it. The impact really is low — the impl holds no OLAS, and an attacker-set `treasury` / `voteWeighting` only affect the impl's own storage — but initializing/locking it in the deploy script costs nothing and removes the question.

**Noted, not a finding:** the proxy has `fallback() external payable` and no `receive()`, so a bare ETH send delegatecalls empty calldata into the Dispenser, which has no receive/fallback, and reverts. That is the correct behaviour for a contract that should only receive value through its payable claim functions.

---

## Stack-level findings (apply across #309 → #315)

### A. [High — will brick mainnet] `dispenserAddress` in globals silently changes meaning

`deploy_07a_dispenser.sh` writes the *implementation* to `dispenserAddress`; `deploy_07b_dispenser_proxy.sh` writes the live proxy to `dispenserProxyAddress`. That mirrors the Tokenomics convention (`tokenomicsAddress` impl / `tokenomicsProxyAddress` proxy) — but roughly fifteen existing scripts still read `dispenserAddress` meaning *the live Dispenser*:

- `scripts/deployment/deploy_08_09_change_managers.js:39,51` passes it to `Tokenomics.changeManagers(...)` and `Treasury.changeManagers(...)`. Setting Tokenomics' `dispenser` to the implementation means `refundFromStaking` — which gates on `dispenser == msg.sender` (`Tokenomics.sol:837`) — rejects calls arriving from the proxy, so **every staking claim reverts**.
- Every L1 deposit processor deploy script (e.g. `scripts/deployment/staking/deploy_04_optimism_deposit_processor.js:33`) passes it as the **immutable** `_l1Dispenser`, permanently binding the processor to the implementation.
- Also `deploy_10_14_change_ownerships.js`, `scripts/audit_chains/audit_contracts_setup.js:252,293,397,455`, and `scripts/fork/tokenomics_retain_calculation.js`.

Tokenomics got away with this convention because `tokenomicsProxyAddress` was introduced alongside its consumers. Here the key is being repurposed under existing consumers. Fix by either naming the implementation `dispenserImplAddress` and keeping `dispenserAddress` = live proxy, or sweeping every consumer in the same PR.

### B. [High — operational] The one-time migration cost is the entire bridging stack

`DefaultDepositProcessorL1.l1Dispenser` (`:41`) and `DefaultTargetDispenserL2.l1DepositProcessor` (`:80`) are both immutable. A new Dispenser address therefore means: redeploy every L1 deposit processor → redeploy every L2 target dispenser → migrate each chain's withheld amount. The proxy prevents this from ever happening *again*, which is the right call, but the series documents the benefit and not the one-time cost. This is also why the `migrate()` change in #311 matters more than its "Low" severity suggests.

### C. [Medium] This stack and governance #215 are mutually load-bearing

`autonolas-governance` #215 makes VoteWeighting's `dispenser` immutable — and it must be the Dispenser *proxy*, not the implementation. The Dispenser's `voteWeighting` stays mutable, which is what makes the circular dependency solvable: Dispenser proxy → VoteWeighting → `Dispenser.changeManagers(0, newVW)`. Neither series states this ordering. Combined with (A), this is exactly where a mainnet migration goes wrong.

### D. [Medium] No fork test runs in CI

The workflow's hardcoded `--mc` allowlist (`.github/workflows/*.yaml:58`) is deliberately avoided by all three fork suites (`StakingClaimForkETH`, `StakingL1ProcessorForkETH`, `StakingL2MigrateForkOP`). That is correct for a fork-less CI, but it means the only tests exercising real Tokenomics / Treasury / deposit-processor wiring never run automatically and will silently rot. Worth a scheduled job gated on an RPC secret.

### E. [Low] Stale ABIs

`abis/*/Dispenser.json` and `docs/configuration.json` still carry the old 9-arg constructor. #314 flags this as a release-artifact step, which I agree with — but given (A) draws from the same family of artifacts, it should be a blocking item on the release checklist rather than a note.

contracts/Dispenser.sol
kupermind: Immutables stay in the implementation construct, whereas those with the possibility to update must be initialized during the proxy deployment.

kupermind:
Consolidated status of the stack-level findings raised here — all now addressed across the stack (current hashes):

- **(A) [High] `dispenserAddress` globals collision** — swept in `9bfd58f` (#314): every live-Dispenser consumer now reads `dispenserProxyAddress` (Tokenomics/Treasury `changeManagers`, all eight L1 deposit-processor deploy + verify scripts, set-processors, ownership, audit, retain-calc, both proposals), keeping `dispenserAddress` = implementation per the `tokenomicsAddress`/`tokenomicsProxyAddress` convention.
- **(B) [High-op] one-time bridging-stack migration cost** — documented in the migration runbook (contracts-in-scope + phased procedure): all L1 processors redeploy, all L2 dispensers migrate; the proxy makes this the last full cascade.
- **(C) [Medium] governance #215 coupling / ordering** — resolved by the zero-VW `initialize` (`cf881af`, #314) + `script_dispenser_change_managers.sh`; the deploy order (proxy → `VoteWeighting(ve, proxy)` → `changeManagers`) is documented and now matches governance internal21, which was corrected post-merge to the same VW=0 procedure.
- **(D) [Medium] no fork test in CI** — N/A per our convention (forks never run in CI); the three fork suites (`StakingClaimForkETH`, `StakingL1ProcessorForkETH`, `StakingL2MigrateForkOP`) are manual, run against public RPCs.
- **(E) [Low] stale ABIs** — on the redeploy regen checklist (runbook Release-artifacts): `abis/*/Dispenser.json` (3-arg ctor, new initialize/changeManagers, 5-value `calculateStakingIncentives` return), `DefaultTargetDispenserL2.json` (4-arg `Migrated`), `configuration.json` post-rewiring.

PR-level lows:
- **`changeImplementation` accepts an EOA** — left as-is: owner/DAO-trusted convention matching Tokenomics / LiquidityManager / BuyBackBurner (not a regression).
- **Implementation left uninitialized** — `deploy_07a` locks the impl by initializing it post-deploy (`9bfd58f`), hardened in `cfceafb` to assert the tx receipt.

The `:337` immutable-vs-storage question is answered inline and documented on the constructor NatSpec (`cfceafb`, #314).

DavidMinarsch:
## Re: consolidated status — two items closed against a document I can't locate

Confirmed resolved from my side, verified in code on the rebased #315 head:

- **(A)** every live-Dispenser consumer now reads `dispenserProxyAddress` (`9bfd58f`). This was the mainnet-bricking one — `Tokenomics`/`Treasury` `changeManagers` now receive the proxy, so `refundFromStaking`'s `dispenser == msg.sender` gate resolves.
- **(C)** the zero-VW `initialize` (`cf881af`) plus `script_dispenser_change_managers.sh` genuinely break the cycle, and governance internal21 was corrected post-merge to the same VW=0 procedure — the two documents of record now agree.
- **Implementation left uninitialized** — `deploy_07a` locks it and `cfceafb` asserts the receipt. Closed.
- **`changeImplementation` accepts an EOA** — agreed, convention-consistent and not a regression. Recording it as accepted risk rather than closed, since the failure mode (permanent brick including the upgrade path) survives the rationale.
- **(D) fork tests in CI** — understood, and I'll defer to the convention. Same note: this is accepted risk, not closure. The assurance case for the whole rework rests on three suites that only run when someone remembers to run them.

### (B) and (E) — which runbook?

Both are closed against "the migration runbook":

> **(B)** … documented in the migration runbook (contracts-in-scope + phased procedure)
> **(E)** … on the redeploy regen checklist (runbook Release-artifacts)

I can't find that document. Searching `origin/main` and every branch in this stack (`#306`, `#307`, `#309`, `#313`, `#314`, `#315`), the only runbook in the repo is `docs/liquidity_migration_runbook.md` — and that one lives on **#307's** branch and covers the POL / LiquidityManager V2→V3 migration, not the Dispenser:

```
$ for ref in origin/main pr306 pr307 pr309 pr313 pr314 pr315; do
    git ls-tree -r --name-only $ref -- docs/ | grep -i runbook; done
pr307: docs/liquidity_migration_runbook.md
```

If a Dispenser migration runbook exists internally, that's fine — could you say where, so the re-audit can be pointed at it? If it's intended to live in this repo, (B) and (E) are still open rather than closed, and (B) in particular is the one item on this stack that nobody can execute from the repo alone: every L1 deposit processor redeploys, every L2 target dispenser migrates (the #311 path), and each chain's withheld amount has to be restored. That sequence is the highest-risk part of the whole rework and it currently has no reviewable artifact.

(E) is smaller but concrete — the ABI regen list is known and enumerable today: `abis/*/Dispenser.json` (3-arg ctor, new `initialize` / `changeManagers`, 5-value `calculateStakingIncentives` return), `DefaultTargetDispenserL2.json` (4-arg `Migrated`), and `configuration.json` post-rewiring.

contracts/Dispenser.sol
DavidMinarsch: _defaultMinStakingWeight and _defaultMaxStakingIncentive can change?

kupermind: They cannot — because they no longer exist. `_defaultMinStakingWeight` / `_defaultMaxStakingIncentive` were the default-fallback immutables, and #314 (`86b35bf`) removed them entirely along with the fallback: the staking min/max are now read solely from Tokenomics. On merged `main` the constructor is `(_olas, _tokenomics, _retainer)`, and those two params are gone.

kupermind:
Agreed on the accepted-risk framing for `changeImplementation`-accepts-EOA and fork-tests-not-in-CI — recording them as accepted risk, not closure. On the fork suites specifically we are reconsidering a manual/secret-gated scheduled CI job so the assurance case does not rest on someone remembering to run them.

On **(B)** and **(E)** — you are right that the migration runbook I referenced was not in the repo (it was a local, security-triage-linked working doc). We are publishing a **sanitized** public version as a reviewable artifact (full phased procedure incl. the L1-processor redeploy / L2-dispenser `migrate` / withheld-restore sequence, and the enumerable ABI-regen list). It will land in a follow-up PR off `main`; I will link it here so the re-audit can be pointed at it.

PR #310, opened 2026-07-24
fix(Dispenser): close vulnerabilities-list items 8, 9, 12, 25

kupermind:
PR-B of the Dispenser rework, **stacked on #309** (base: `refactor/dispenser-proxy`). Behavioral fixes for the four Dispenser items from `docs/Vulnerabilities_list_tokenomics.md` (referred to below as "item N" — the numbering is the doc's, not GitHub issues/PRs), each with a regression test proven to **fail on the pre-fix code** (the whole new suite was run against the parent commit: 0/5 pass there, 5/5 pass here).

**Fixes**
- **item 12 (High)** — `calculateStakingIncentives` is public and state-mutating; it set the one-way `mapZeroWeightEpochRefunded[j]` flag while leaving the actual refund to the caller, so a standalone external call could permanently strand a zero-weight epoch's staking inflation. The flag write and `Tokenomics.refundFromStaking` are now **atomic in the same call**; `totalReturnAmount` no longer includes zero-weight amounts (no caller double-refund).
- **item 9 (Low)** — the withheld-covered portion of claimed incentives is paid from OLAS already minted under a previous allocation, yet the current allocation was still counted as inflation spent. Both netting paths now return the `withheldUsed` amount to staking inflation (direct `refundFromStaking` on the single-claim path; via `totalAmounts[2]` on the batch path).
- **item 25 (Informative)** — `addNominee` now clears `mapRemovedNomineeEpochs` from a previous nominee lifecycle, so a removed-then-re-added nominee no longer permanently reverts every claim. The Dispenser no longer relies on the upstream VoteWeighting enforcing "remove is final".
- **item 8 (Informative)** — `changeManagers` only swaps `voteWeighting` while staking incentives are paused (`StakingIncentivesPaused`/`AllPaused`; new `Unpaused` error), forcing outstanding claims to be settled before the nominee set is orphaned. Treasury changes stay ungated. Deploy-time wiring is unaffected (the Dispenser initializes paused).

**Tests** — new `test/DispenserFixes.t.sol` (5): atomic zero-weight refund + no-double-refund on the subsequent claim; withheld-reuse inflation credit exact-delta (single + batch, incl. netted transfer/mint assertions); remove→re-add claimability; pause-gate matrix (Unpaused / DevIncentivesPaused revert, StakingIncentivesPaused / AllPaused pass). Runs via the DispenserProxy end-to-end with real Tokenomics + Treasury and mock VW / deposit processor.

**Verified:** new suite 5/5 (0/5 on parent); forge Dispenser 3/3 + proxy 9/9; hardhat Dispenser suites 31 passing; full `yarn test` green; solhint/eslint clean.

### DIFF
--- a/contracts/Dispenser.sol
+++ b/contracts/Dispenser.sol
@@ -234,6 +234,9 @@ error AlreadyInitialized();
 /// @dev The contract is paused.
 error Paused();
 
+/// @dev The contract is unpaused.
+error Unpaused();
+
 /// @dev Incentives claim has failed.
 /// @param account Account address.
 /// @param reward Reward amount.
@@ -709,14 +712,21 @@ contract Dispenser {
             if (transferAmounts[i] > 0) {
                 uint256 withheldAmount = mapChainIdWithheldAmounts[chainIds[i]];
                 if (withheldAmount > 0) {
+                    uint256 withheldUsed;
                     if (withheldAmount >= transferAmounts[i]) {
+                        withheldUsed = transferAmounts[i];
                         withheldAmount -= transferAmounts[i];
                         transferAmounts[i] = 0;
                     } else {
+                        withheldUsed = withheldAmount;
                         transferAmounts[i] -= withheldAmount;
                         withheldAmount = 0;
                     }
                     mapChainIdWithheldAmounts[chainIds[i]] = withheldAmount;
+
+                    // The withheld-covered portion is paid from OLAS already minted under a previous allocation;
+                    // add it to the return amount so the caller refunds it back to staking inflation
+                    totalAmounts[2] += withheldUsed;
                 }
             }
 
@@ -761,6 +771,14 @@ contract Dispenser {
 
         // Change Vote Weighting contract address
         if (_voteWeighting != address(0)) {
+            // Swapping the Vote Weighting contract orphans the per-nominee claim cursors recorded under the
+            // previous nominee set: any staking incentives not yet claimed become unreachable. Require staking
+            // incentives to be paused first, so all outstanding claims are settled before the swap
+            Pause currentPause = paused;
+            if (currentPause != Pause.StakingIncentivesPaused && currentPause != Pause.AllPaused) {
+                revert Unpaused();
+            }
+
             voteWeighting = _voteWeighting;
             emit VoteWeightingUpdated(_voteWeighting);
         }
@@ -832,6 +850,12 @@ contract Dispenser {
         }
 
         mapLastClaimedStakingEpochs[nomineeHash] = ITokenomics(tokenomics).epochCounter();
+
+        // Clear a possible removal epoch left from a previous nominee lifecycle: without this, a re-added
+        // nominee would permanently revert in _checkpointNomineeAndGetClaimedEpochCounters (the claim cursor
+        // just set above always satisfies firstClaimedEpoch >= epochRemoved). The Dispenser no longer relies
+        // on the upstream Vote Weighting enforcing "remove is final"
+        delete mapRemovedNomineeEpochs[nomineeHash];
 
         emit AddNomineeHash(nomineeHash);
     }
@@ -937,6 +961,9 @@ contract Dispenser {
 
     /// @dev Calculates staking incentives for a specific staking target.
     /// @notice Call this function via staticcall in order not to write in the nominee checkpoint map.
+    ///         A state-mutating call refunds zero-total-weight epochs back to staking inflation atomically with
+    ///         setting their one-way refunded flag, so a standalone external call cannot strand the refund;
+    ///         the returned totalReturnAmount accordingly does not include the zero-weight epoch amounts.
     /// @param numClaimedEpochs Specified number of claimed epochs.
     /// @param chainId Chain Id.
     /// @param stakingTarget Staking target corresponding to the chain Id.
@@ -1016,8 +1043,11 @@ contract Dispenser {
 
             // Check if the totalWeightSum is zero, then all staking incentives must be returned back to tokenomics
             if (totalWeightSum == 0) {
+                // The one-way refunded flag and the actual refund are performed atomically in the same call:
+                // this function is public, and setting the flag while deferring the refund to the caller would
+                // let a standalone external call permanently mark the epoch refunded without any refund executed
                 mapZeroWeightEpochRefunded[j] = true;
-                totalReturnAmount += stakingPoint.stakingIncentive;
+                ITokenomics(tokenomics).refundFromStaking(stakingPoint.stakingIncentive);
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


77ph:
Reviewed against `Vulnerabilities_list_tokenomics.md` **#8 / #9 / #12 / #25**, tracing the full staking-incentive accounting on top of #309.

- **#12 (High) — correct.** The one-way `mapZeroWeightEpochRefunded[j]` flag and `Tokenomics.refundFromStaking(...)` are now atomic in `calculateStakingIncentives`, and the zero-weight amount is dropped from `totalReturnAmount`. Traced both callers: the single path refunds `returnAmount` and `withheldUsed` separately; the batch path aggregates into `totalAmounts[2]` and refunds once — zero-weight is refunded inline and is not double-counted in either. `refundFromStaking` is `dispenser`-gated with a `uint96` overflow check, so the now-reachable "public state-mutating call performs the refund" path is authorized and bounded (once per epoch via the flag). No on-chain `staticcall` consumer of `calculateStakingIncentives` exists, so the added external call breaks no preview path.
- **#9 (Low) — correct.** `withheldUsed` is returned to staking inflation on both paths (direct `refundFromStaking` single, `totalAmounts[2]` batch), disjoint from the other refunds.
- **#25 (Informative) — correct.** `delete mapRemovedNomineeEpochs[nomineeHash]` in `addNominee` is option (a) from the vuln-list note; a no-op on a first-time add, and it unblocks a remove→re-add lifecycle. The Dispenser no longer depends on upstream VoteWeighting enforcing "remove is final" — good, especially now that `changeManagers` can repoint `voteWeighting`.
- **#8 (Informative) — correct.** The `voteWeighting` swap now requires `StakingIncentivesPaused` / `AllPaused`. Deploy-time wiring is unaffected since `voteWeighting` is set in `initialize`, not `changeManagers`.

`DispenserFixes.t.sol` is **5/5** on this branch. The suite is written against the new `Unpaused` error / 2-arg `changeManagers`, so it is tightly coupled to the fixes — a good sign it exercises the changed behaviour rather than passing vacuously.

contracts/Dispenser.sol
```diff
@@ -1016,8 +1043,11 @@ contract Dispenser {
 
             // Check if the totalWeightSum is zero, then all staking incentives must be returned back to tokenomics
             if (totalWeightSum == 0) {
+                // The one-way refunded flag and the actual refund are performed atomically in the same call:
+                // this function is public, and setting the flag while deferring the refund to the caller would
+                // let a standalone external call permanently mark the epoch refunded without any refund executed
                 mapZeroWeightEpochRefunded[j] = true;
-                totalReturnAmount += stakingPoint.stakingIncentive;
+                ITokenomics(tokenomics).refundFromStaking(stakingPoint.stakingIncentive);
```
kupermind, line 1050: This is not a great solution and will be rewritten in #315  to make this function **view**, stripping out the rest to upper function calls.

DavidMinarsch:
## Review: vuln-list items 8 / 9 / 12 / 25 (PR-B)

Stack-level findings for the whole series are on #309 (globals `dispenserAddress` rename, bridging-stack migration cost, governance #215 coupling, fork tests outside CI, stale ABIs). Verified locally: forge `--mc Dispenser` 19/19 on the final stacked state.

### item 9 (withheld netting) — correct, and I checked the leg that would have made it wrong

The fix is only sound if the withheld amount was *not* already credited back to inflation when it was synced from L2. It wasn't: `syncWithheldAmount` (`Dispenser.sol:1347`) only records `mapChainIdWithheldAmounts[chainId] += amount` and emits — it never calls `refundFromStaking`.

So the accounting closes: the withheld OLAS was minted exactly once under the earlier epoch's allocation and physically sits on L2; a later claim covered by it mints nothing, and `refundFromStaking(withheldUsed)` restores the current epoch's unspent allocation. Net minted equals net debited. Without the fix, inflation is over-debited by exactly `withheldUsed` — which is what the doc's item 9 describes.

### item 25 — right, and correctly reasoned

`delete mapRemovedNomineeEpochs[nomineeHash]` is necessary, not defensive: the claim cursor set on the line above is `epochCounter`, which always satisfies `firstClaimedEpoch >= epochRemoved`, so without the delete a re-added nominee reverts on every claim forever.

### item 8 (pause gate) — a checklist, not a control

The owner can pause, swap `voteWeighting`, and unpause within a single proposal, so this enforces a runbook step rather than an invariant. That is fine for an Informative item, but the PR body's "forcing outstanding claims to be settled before the nominee set is orphaned" overstates it.

Also worth correcting in the reasoning: `mapLastClaimedStakingEpochs` is keyed by `keccak256(Nominee(target, chainId))`, which is independent of *which* VoteWeighting instance is wired. The cursors therefore **survive** a swap. What is actually lost is pending claims for nominees that are not re-added to the new VoteWeighting — a narrower (and more precisely stateable) hazard than "orphaned cursors".

### item 12 — superseded two PRs later

The fix landed here (inline `refundFromStaking` from a `public` state-mutating function) is replaced in #315 by the `view` + caller-owns-effects design, and this PR's item-12 regression test is rewritten there. Not a defect — but anyone reviewing item 12 should assess #315's final state, not this diff. Worth either annotating this body or squashing the stack before merge.

### Note

`calculateStakingIncentives` remains `public` and state-mutating in this PR, so the NatSpec "Call this function via staticcall in order not to write in the nominee checkpoint map" is still unachievable for any epoch range containing a zero-weight epoch. #315 resolves this properly; flagging only so the intermediate state isn't mistaken for the intended one.

kupermind:
Item 8 wording corrected in `0396533` (#315): the `changeManagers` comment now states the claim cursors are `keccak256(Nominee)`-keyed and survive a Vote Weighting swap, and that what is lost is pending claims for nominees not re-added — not "orphaned cursors". Item 12's intermediate state is superseded by #315's view refactor, as you noted.

DavidMinarsch (dismissed):
Approving. Every item I raised on this PR is either fixed or answered, and I traced the accounting rather than taking the description on trust.

**item 9 (withheld netting) — verified correct.** The fix is only sound if the withheld amount wasn't already credited back to inflation when synced from L2, and it wasn't: `syncWithheldAmount` only records `mapChainIdWithheldAmounts[chainId] += amount`, it never calls `refundFromStaking`. So the withheld OLAS is minted exactly once under the earlier epoch and sits on L2; a later claim covered by it mints nothing, and `refundFromStaking(withheldUsed)` restores the current epoch's unspent allocation. Net minted equals net debited. Without the fix inflation is over-debited by exactly `withheldUsed`.

**item 25 — correct and necessary, not defensive.** The claim cursor set on the line above is `epochCounter`, which always satisfies `firstClaimedEpoch >= epochRemoved`, so without the `delete` a re-added nominee reverts on every claim forever.

**item 8 — wording corrected.** The `changeManagers` comment now states that cursors are `keccak256(Nominee)`-keyed and survive a Vote Weighting swap, and that what's lost is pending claims for nominees not re-added. That's the accurate framing. The gate remains a runbook enforcement rather than an invariant (the owner can pause, swap and unpause in one proposal), which is fine for an Informative item.

**item 12 — approving the intermediate state knowingly.** The inline `refundFromStaking` from a `public` function that lands here is replaced in #315 by the `view` + caller-owns-effects design, and the regression test is rewritten there. Normal for a stack; noting it so nobody reviews item 12 from this diff alone. #315's version is the one I checked in detail (sparse-array bounds, zero sentinel, batch dedupe via the `mapZeroWeightEpochRefunded` skip).

Note on sequencing: this merges into `refactor/dispenser-proxy`, not `main`. #309 is still `REVIEW_REQUIRED` and carries the open question about where the Dispenser migration runbook lives — findings (B) and (E) are currently closed against a document I can't locate in the repo. Nothing in this PR is blocked on that, but the stack can't reach `main` until #309 does.

PR #311, opened 2026-07-24
fix(TargetDispenserL2): surface withheldAmount in migrate() event (#10)

kupermind:
PR-C of the Dispenser rework, **stacked on #310** (base: `fix/dispenser-vuln-fixes`). Closes `Vulnerabilities_list_tokenomics.md` **#10** in `DefaultTargetDispenserL2`.

**Problem.** `migrate(newL2TargetDispenser)` transfers the full OLAS balance to the new L2 dispenser and permanently bricks the old one, but the new dispenser deploys with `withheldAmount == 0` and there's no L1-side setter — the DAO must restore it via `updateWithheldAmountMaintenance()` right after. The `Migrated` event previously carried only the OLAS balance, so the exact value to restore was off-chain knowledge and could be silently dropped (losing the inflation information for OLAS already minted and held on L2).

**Fix.** The `Migrated` event now includes `withheldAmount`, making the value-to-restore part of the on-chain migration record. Nothing else changes (OLAS balance still moves, owner still zeroed, contract still locks for good). It's a base-contract change inherited by all chain variants (Arbitrum / Gnosis / Optimism / Polygon `TargetDispenserL2`).

**Tests.** `test/StakingBridging.js` withheld-recovery migrate test now asserts the `Migrated` event carries `(sender, newDispenser, amount, withheldAmount)` and that the DAO's restored value on the new dispenser matches the emitted one. Suite green: **21 passing**. `yarn compile` clean; solhint unaffected.

Separate PR per the plan because this is a different contract family (staking stack) with its own suite. After this, the remaining item is **PR-D** (forge deploy scripts for the Dispenser impl + `DispenserProxy`, and L1 mainnet-fork tests of the proxied claim path).

### DIFF
--- a/contracts/staking/DefaultTargetDispenserL2.sol
+++ b/contracts/staking/DefaultTargetDispenserL2.sol
@@ -57,7 +57,8 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
     event Drain(address indexed owner, uint256 amount);
     event TargetDispenserPaused();
     event TargetDispenserUnpaused();
-    event Migrated(address indexed sender, address indexed newL2TargetDispenser, uint256 amount);
+    event Migrated(address indexed sender, address indexed newL2TargetDispenser, uint256 amount,
+        uint256 withheldAmount);
     event LeftoversRefunded(address indexed sender, uint256 leftovers);
 
     // receiveMessage selector (Ethereum chain)
@@ -516,6 +517,11 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
     ///         No further write interaction with the contract is going to be possible.
     ///         If the withheld amount is nonzero, it is regulated by the DAO directly on the L1 side.
     ///         If there are outstanding queued requests, they are processed by the DAO directly on the L2 side.
+    ///         The migrated OLAS balance AND the current withheldAmount are both emitted in the Migrated event:
+    ///         the new L2 target dispenser deploys with withheldAmount == 0, so the DAO must restore it via
+    ///         updateWithheldAmountMaintenance(withheldAmount) on the new contract right after migration, or the
+    ///         inflation information (OLAS already minted and held on L2) is lost. Emitting it here makes the
+    ///         exact value to restore part of the migration record rather than off-chain tribal knowledge.
     function migrate(address newL2TargetDispenser) external {
         // Reentrancy guard
         if (_locked > 1) {
@@ -556,7 +562,9 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
         // Zero the owner
         owner = address(0);
 
-        emit Migrated(msg.sender, newL2TargetDispenser, amount);
+        // Emit the withheld amount alongside the migrated OLAS balance: it is the exact value the DAO must
+        // restore on the new dispenser via updateWithheldAmountMaintenance(), so it must be in the record
+        emit Migrated(msg.sender, newL2TargetDispenser, amount, withheldAmount);
 
         // _locked is now set to 2 for good
     }
--- a/contracts/staking/DefaultTargetDispenserL2.sol
+++ b/contracts/staking/DefaultTargetDispenserL2.sol
@@ -57,6 +57,10 @@ abstract contract DefaultTargetDispenserL2 is IBridgeErrors {
     event Drain(address indexed owner, uint256 amount);
     event TargetDispenserPaused();
     event TargetDispenserUnpaused();
+    /// @dev Emitted on migrate(). NOTE: this 4-arg form changed the event signature/topic0 from the previous
+    ///      Migrated(address,address,uint256). During the L2 cutover both forms coexist on-chain (already-deployed
+    ///      dispensers emit the 3-arg form, new ones this 4-arg form), so any indexer, subgraph or monitor keyed on
+    ///      the old topic0 must handle both.
     event Migrated(address indexed sender, address indexed newL2TargetDispenser, uint256 amount,
         uint256 withheldAmount);
     event LeftoversRefunded(address indexed sender, uint256 leftovers);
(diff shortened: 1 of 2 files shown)


77ph:
`#10` fix reviewed — **event-only, as intended.** `Migrated` gains `withheldAmount` (the current storage value the DAO must restore via `updateWithheldAmountMaintenance` on the new dispenser); the `migrate()` fund transfer / owner-zeroing / permanent lock are unchanged, and the change is inherited by all chain variants (Arbitrum / Gnosis / Optimism / Polygon). One operational note: the 4-argument `Migrated` is an ABI change, so any off-chain indexer / migration tooling parsing the previous 3-arg signature needs updating.

For the record — this is the same contract as Immunefi report **85857** (`redeem()` not re-verifying a since-removed target). That report has since been finalized **BY-DESIGN / Informational, no change** (validity is fixed at the funding epoch; a target de-authorized on L1 never has its emissions forwarded to the L2 queue in the first place, so no reclaimable amount is stranded there). `#311` touches only the `migrate()` event and is unrelated to that question, so nothing further is owed here on that axis.

DavidMinarsch:
## Review: `migrate()` withheld amount in the event (PR-C)

Stack-level findings for the whole series are on #309. The change itself is small, correct, and inherited cleanly by all four chain variants.

### The framing is stronger than the fix

The body says the value to restore "was off-chain knowledge and could be silently dropped". It wasn't off-chain: `migrate()` leaves `withheldAmount` untouched in storage and only zeroes `owner`, so the old dispenser's `withheldAmount()` getter keeps returning the exact value indefinitely after migration. The event adds convenience and an indexable record — not recoverability.

That matters because item 10's actual risk is **the DAO forgetting to call `updateWithheldAmountMaintenance()` on the new dispenser**, and that risk is untouched here. If you want to close it rather than document it, the shape would be `migrate(address newL2TargetDispenser, uint256 expectedWithheldAmount)` reverting on mismatch — forcing the operator to acknowledge the value at migration time, which is when they still have it in hand. Given (B) on #309 — this migration is about to be executed on every L2 as part of the Dispenser cutover — that seems worth the extra parameter.

### [Low] `Migrated` topic0 changes

`Migrated(address,address,uint256)` → `Migrated(address,address,uint256,uint256)` is a new event signature. Already-deployed L2 target dispensers keep emitting the 3-arg form; only newly deployed ones emit the 4-arg form, so for the duration of the cutover both exist on-chain. Any indexer, subgraph or monitoring keyed on the old signature must handle both. Worth a line in the body and an ABI regeneration.

### Test nit

The test restores from `olas.balanceOf(newArbitrumTargetDispenserL2.address)` and then asserts `newDispenser.withheldAmount() == withheldToCarry`. That only holds because balance and withheld coincide in this scenario, so the assertion doesn't actually prove the emitted value was used. Restoring directly from the emitted `withheldAmount` (or constructing the case so balance ≠ withheld) would make the test test the thing it claims.

DavidMinarsch:
## Follow-up review — `080321a`

The test nit is properly fixed. The mint is now `defaultAmount + 50` while `updateWithheldAmountMaintenance` records `defaultAmount`, so the physical balance genuinely exceeds the accounting withheld amount — which is what makes the assertion meaningful:

- `Migrated` is asserted with `balanceToMigrate` and `withheldToCarry` as *distinct* values, so the withheld argument demonstrably reflects storage rather than the transferred balance;
- the restore on the new dispenser is driven by the emitted withheld value, and the follow-up `expect` confirms the full balance moved across separately.

That closes the "only passes because the two coincide" objection. Nothing further from me on this PR beyond the two points already raised: the `Migrated` topic0 change means indexers see both the 3-arg and 4-arg signatures during the L2 cutover, and item #10's operational risk (the DAO not calling `updateWithheldAmountMaintenance` at all) is recorded rather than closed — `migrate(newDispenser, expectedWithheldAmount)` would close it if you want that, given every L2 runs this migration as part of the Dispenser cutover.

kupermind:
Both remaining points are now recorded in `326379d`: the `Migrated` topic0 / 3-vs-4-arg coexistence is documented on the event NatSpec (indexers must handle both during the cutover), and the `migrate()` NatSpec explains why no `expectedWithheldAmount` parameter — pause / migrate / updateWithheldAmountMaintenance ship as one atomic DAO L1→L2 proposal already passing the emitted value, and the operator would read any "expected" value from this same storage, so a confirmation arg adds ceremony without safety. The 4-arg `Migrated` ABI regen is on the redeploy checklist.

DavidMinarsch (dismissed):
Approving. Small, self-contained, and both review items are resolved.

**The test now proves what it claims.** `080321a` mints `defaultAmount + 50` while recording `defaultAmount` as withheld, so the physical balance genuinely exceeds the accounting figure. `Migrated` is asserted with the two as distinct values, and the restore is driven by the emitted withheld amount rather than the migrated balance. That closes the "only passes because they coincide" objection — the earlier version would have passed against a contract that emitted the balance.

**The `expectedWithheldAmount` suggestion is correctly declined.** I proposed `migrate(newDispenser, expectedWithheldAmount)` reverting on mismatch, to force the operator to acknowledge the value. The NatSpec rebuttal is right: pause / migrate / `updateWithheldAmountMaintenance` ship as one atomic DAO L1→L2 proposal that already carries the emitted value, and the operator would read any "expected" figure from this same storage — so a confirmation argument adds ceremony without safety. Withdrawn.

**Topic0 coexistence is now documented** where an integrator will actually see it: already-deployed dispensers keep emitting the 3-arg `Migrated`, new ones the 4-arg form, and both exist on-chain through the L2 cutover. The note that the value to restore is the emitted `withheldAmount` and *not* the migrated balance is the right thing to have in the contract rather than only in a runbook.

One framing point I'd still make, not blocking: the event is a record, not a recovery mechanism — `withheldAmount` remains readable from the old contract after migration since `migrate()` only zeroes `owner`. The NatSpec now says this, which is what I was after.

Sequencing: this merges into `fix/dispenser-vuln-fixes` (#310), which I've also approved. The stack reaches `main` only via #309, still `REVIEW_REQUIRED`.

PR #312, opened 2026-07-28
docs(audit): internal18 — Dispenser rework releasing audit (#309/#310/#311)

77ph:
Standalone releasing-audit record for the Dispenser-rework stack (**#309 → #310 → #311**), so the audit result is legible independently of the per-PR review threads.

**Adds:** `audits/internal18/README.md`.

## Verdict: GREEN-LIGHT the code — CONDITIONAL on PR-D before the on-chain deploy

All five `Vulnerabilities_list_tokenomics.md` items (**#8/#9/#10/#12/#25**) are correctly closed, the proxy refactor is behaviour-preserving, and no regression was found — verified first-hand (built the tree, ran the forge suites 17/17, traced the staking-incentive accounting on both claim paths, recomputed the proxy slot, swept every changed signature for callers).

The on-chain **deploy** green-light is held on the planned **PR-D**: deploy scripts (the constructor/`initialize` signatures changed and a wrong constructor immutable "silently rewires the proxy") + an **L1 mainnet-fork test of the proxied claim path** + off-chain tooling updated for the 4-arg `Migrated`.

Highlights (full detail in the doc):
- **#309** proxy: `main→#309` diff is plumbing only; `PROXY_DISPENSER` slot recomputed and matches read/write sites; `initialize` re-init-guarded with no front-run window; storage `frozen append-only` (safe = greenfield deploy). DispenserProxy 9/9.
- **#12 (High)** flag + `refundFromStaking` now atomic; `totalReturnAmount` excludes zero-weight; single + batch traced → no double / missing refund; `refundFromStaking` is `dispenser`-gated + overflow-checked. **#9/#25/#8** correct per the vuln-list. DispenserFixes 5/5.
- **#311 (#10)** event-only; `withheldAmount` = value the DAO restores; inherited by all chain variants.
- **Regression:** `Dispenser.changeManagers` 3→2 args breaks no deployment/proposal script (those target Tokenomics/Treasury). Committed tree compiles clean.
- **85857** (same contract) is finalized BY-DESIGN/Informational and unrelated to #311.

### DIFF
--- a/audits/internal18/README.md
+++ b/audits/internal18/README.md
@@ -1,0 +1,86 @@
+# Internal Audit 18 — autonolas-tokenomics (Dispenser rework — releasing audit)
+
+**Audit date**: 2026-07-27
+**Audited change**: the Dispenser-rework PR stack **#309 → #310 → #311** (clean stack over `main` `029b557`):
+- **#309** `refactor/dispenser-proxy` (`22f5352`) — put the `Dispenser` behind a `DispenserProxy`.
+- **#310** `fix/dispenser-vuln-fixes` (`c60fbe5`, stacked on #309) — `Vulnerabilities_list_tokenomics.md` items **#8, #9, #12, #25**.
+- **#311** `fix/l2-migrate-withheld` (`396e0a5`, stacked on #310) — item **#10** (`DefaultTargetDispenserL2.migrate()`).
+
+**Deployment status**: pre-deploy. The three PRs are open; deploy scripts + fork tests are a separate follow-up (PR-D) and are **not** in this stack.
+**Auditor**: audit-claude
+**Scope of change** (release unit = cumulative diff `main..396e0a5`): 3 production contracts —
+`contracts/Dispenser.sol` (+157/−104 net over the stack), `contracts/proxies/DispenserProxy.sol` (new, 68 LOC), `contracts/staking/DefaultTargetDispenserL2.sol` (+10/−2). No other production contract changed; no deploy script changed.
+
+## Verdict
+
+**GREEN-LIGHT the code — CONDITIONAL on PR-D before the on-chain deploy.**
+
+All five vuln-list items (#8/#9/#10/#12/#25) are correctly closed, the proxy refactor is behaviour-preserving, and no regression was found. The contract changes are sound to merge. The **on-chain deployment**, however, should wait on the planned **PR-D** (deploy scripts + an L1 mainnet-fork test of the proxied claim path), because the constructor/`initialize` signatures changed and none of that deploy machinery is present or exercised yet (§5).
+
+## Method (everything verified first-hand)
+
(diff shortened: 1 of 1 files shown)


DavidMinarsch (changes requested):
## Review: internal18 — Dispenser rework releasing audit

The audit itself is solid work: the `PROXY_DISPENSER` slot is recomputed and confirmed (I independently derived the same `0x8bd2…8f83`), the greenfield-storage caveat in §1 is the right thing to have written down, and the disjoint-refund trace across both claim paths in §2 is correct for the code it was pinned to.

The problem is timing. This is dated **2026-07-27** and pinned to `396e0a5`, but it lands on `main` *after* the stack continued through #313 → #314 → #315 (all merged this morning). Merging it as a dated historical record is legitimate — but four passages read as current status, and one of them is the compliance table, which is the artifact an external re-auditor will actually read.

### 1. §2 and the compliance table certify an item-12 design that no longer exists

> `calculateStakingIncentives` now writes the one-way `mapZeroWeightEpochRefunded[j]` flag **and** calls `Tokenomics.refundFromStaking(...)` atomically in the same call

and the table row:

> | #12 zero-weight refund atomicity | High | #310 `calculateStakingIncentives` | ✔ correct (atomic; no double-refund) |

On merged `main`, `calculateStakingIncentives` is `public view` (`Dispenser.sol:985`). #315 removed the inline refund and moved the effects into the claim paths, precisely because a `public` state-mutating function that refunds is the weaker form. So the shipped fix for item #12 is the opposite of what this document certifies.

Related: §2's supporting reasoning —

> `calculateStakingIncentives` has only the two internal state-mutating callers (no on-chain `staticcall` consumer), so the added external call breaks no preview path

— is the exact rationale #315 reversed. Worth flagging in the header, because a reader who trusts it will conclude the view refactor was unnecessary.

### 2. §4's regression sweep is now false

> `Dispenser.changeManagers` (3→2 args) … not called by any deployment or proposal script

`scripts/deployment/script_dispenser_change_managers.sh` (added in #314) calls `changeManagers(address,address)` on the Dispenser proxy — it is the script that wires Vote Weighting in after deploy. True when written, false on `main`.

### 3. §5 — two stale facts, one satisfied condition

- "The constructor now takes **five** bytecode immutables" — #314 reduced it to three (`olas`, `tokenomics`, `retainer`).
- The **PR-D release condition** ("green-light requires deploy scripts + an L1 mainnet-fork test") is satisfied: #313 landed both, and #314/#315 followed.

### Suggested fix

A dated `> **Superseded 2026-08-13.**` block under the Verdict, noting: the stack continued to #313/#314/#315; item #12's shipped design is the `view` + caller-owns-effects form (see internal19 Part B); the constructor is 3-arg; the PR-D condition is met; and `script_dispenser_change_managers.sh` now calls `Dispenser.changeManagers`. That preserves the audit as a point-in-time record without leaving a superseded design certified as shipped — the same pattern `autonolas-governance` internal21 used for its own post-merge correction.

Nothing here is a finding against the code. It is entirely about the document not out-living its accuracy.

PR #313, opened 2026-07-30
test(Dispenser): deploy scripts + L1 mainnet-fork claim test (PR-D)

kupermind:
PR-D of the Dispenser-rework series (stacked on #311). **Deploy machinery + a real-contract fork test** — the release condition the auditor held the on-chain deploy on. No contract changes; the audited code from #309/#310/#311 is untouched.

### Deploy scripts (forge `.sh`)
- `deploy_07a_dispenser.sh` — Dispenser implementation (the 5 bytecode immutables; warns that a wrong immutable silently rewires the proxy).
- `deploy_07b_dispenser_proxy.sh` — `DispenserProxy` with `initialize()` calldata (deployer becomes owner, staking incentives paused until wired).
- `script_dispenser_change_implementation.sh` — owner-gated `changeImplementation` upgrade helper.

The deploy-order inversion (Tokenomics proxy → Dispenser impl+proxy → repoint) and the greenfield-only storage-image invariant are documented inline in the scripts.

### L1 mainnet-fork test — `test/StakingClaimForkETH.t.sol`
Runs against the **live** deployed Tokenomics proxy and Treasury (real storage: real inflation-curve position, 75% staking fraction, real Timelock owner), not fresh copies or mocks:
- **`test_migrationWiring_realContractsRepointed`** — deploys the new impl+proxy, repoints the real Timelock-owned Treasury/Tokenomics at it, asserts the wiring + owner-gated `changeImplementation`.
- **`test_proxiedClaim_realTreasuryMint_realInflationRefund`** — the money-path: `claim → Treasury.withdrawToAccount → real OLAS mint (60k OLAS) → deposit processor`, capped remainder refunded to live staking inflation.
- **`test_proxiedClaim_withheldNetting_refundsReusedInflation`** — withheld netting over real Tokenomics: transfer netted by withheld, reused-withheld portion refunded to inflation (#9 path on real contracts).

Vote Weighting (gauge voting is week-boundary/veOLAS-lock bound, infeasible to drive on a fork) and the L2 deposit processor (bridge plumbing) are deterministic stand-ins; every OLAS movement and inflation accounting is real.

Run: `forge test -f $FORK_ETH_NODE_URL --mc StakingClaimForkETH -vvv` — **3/3 pass**.

### CI safety
The fork-test contract name avoids the `Dispenser`/`Treasury`/`Depository` substrings, so the fork-less CI `forge test --mc Dispenser` allowlist does **not** pick it up (verified). Unit suites unchanged: `--mc Dispenser` 17/17.

### Out of scope (by design)
No `Vulnerabilities_list_tokenomics.md` edit — those entries come off only after the physical redeploy + rewiring closes the exposure on-chain, not when a fix branch merges. No CHANGELOG edit (tag-scoped).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,6 +56,7 @@ forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerETH -vvv       # ETH mainn
 forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerObservationCardinalityGasETH -vvv
 forge test -f $FORK_ETH_NODE_URL --mc UniswapPriceOracleETH -vvv
 forge test -f $FORK_ETH_NODE_URL --mc BuyBackBurnerUniswapETH -vvv
+forge test -f $FORK_ETH_NODE_URL --mc StakingClaimForkETH -vvv        # proxied Dispenser claim path vs live Tokenomics/Treasury
 forge test -f $FORK_BASE_NODE_URL --mc LiquidityManagerBase -vvv      # Base
 forge test -f $FORK_BASE_NODE_URL --mc BalancerPriceOracleBase -vvv
 forge test -f $FORK_BASE_NODE_URL --mc BuyBackBurnerBalancerBase -vvv
--- a/scripts/deployment/deploy_07a_dispenser.sh
+++ b/scripts/deployment/deploy_07a_dispenser.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+# Deploys the Dispenser implementation (proxy-based Dispenser). The constructor sets the
+# implementation-bytecode immutables only; the mutable state is set by initialize(), which is
+# delegatecall-ed by the DispenserProxy constructor in deploy_07b_dispenser_proxy.sh.
+#
+# WARNING: the constructor immutables are safety-critical — a wrong value silently rewires the proxy.
+# Double-check tokenomicsProxyAddress (NOT the tokenomics implementation) and retainerAddress.
+#
+# Globals fields consumed:
+#   olasAddress            : OLAS token address
+#   tokenomicsProxyAddress : Tokenomics PROXY address (implementation immutable)
+#   retainerAddress        : retainer in bytes32 form
+#   minStakingWeight       : default min staking weight (bounded by uint16)
+#   maxStakingIncentive    : default max staking incentive (bounded by uint96)
+# Globals fields written:
+#   dispenserAddress       : deployed Dispenser implementation address (consumed by deploy_07b_*)
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 0
+fi
+
+# Read variables using jq
+contractVerification=$(jq -r '.contractVerification' $globals)
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
+chainId=$(jq -r '.chainId' $globals)
+networkURL=$(jq -r '.networkURL' $globals)
+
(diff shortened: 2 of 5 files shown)


DavidMinarsch:
## Review: deploy scripts + `StakingClaimForkETH` (PR-D)

Stack-level findings for the whole series are on #309 — **(A) in particular lands squarely on this PR**, since it is the one that introduces the `dispenserAddress` (implementation) / `dispenserProxyAddress` (live proxy) split while ~15 existing scripts still read `dispenserAddress` as the live address. Worth reading before merging this one.

Otherwise: good addition. A fork test against the live Tokenomics proxy and Treasury — real inflation-curve position, real mint, real refund — is the right release gate for a change of this shape, and it is exactly the coverage the unit suites cannot provide.

### [Medium] `deploy_07b_dispenser_proxy.sh` validates only one of its five inputs

`dispenserAddress` gets a null/zero check; `treasuryAddress`, `voteWeightingAddress`, `maxNumClaimingEpochs` and `maxNumStakingTargets` do not. A missing globals key makes `jq -r` emit the literal string `null`, which then goes straight into `cast calldata "initialize(address,address,uint256,uint256)"`. This matters more after #314, where `maxNumClaimingEpochs` / `maxNumStakingTargets` become **set-once at `initialize()`** with no runtime setter — a bad value there is an implementation redeploy, not a fix-up transaction. Mirror the existing `dispenserAddress` check for all five.

### [Low] The body is stale on merge

`deploy_07a_dispenser.sh` is described here as taking "the 5 bytecode immutables", but #314 reduces the constructor to `(olas, tokenomics, retainer)` and rewrites the script. Since #314 is stacked directly on this, the description will be wrong the moment both land. Fine to leave if the stack is squashed; worth a note if not.

### Confirmed

The CI-safety claim checks out: the workflow's allowlist (`.github/workflows/*.yaml:58`) is a hardcoded list of `--mc` filters, and `StakingClaimForkETH` matches none of them. See (D) on #309 for the flip side — no fork suite runs in CI at all, so this and the two suites #315 adds are manual-only.

kupermind:
The `deploy_07b` all-inputs validation (treasury / maxNumClaimingEpochs / maxNumStakingTargets) is added in `9bfd58f` on #314, together with the #309-A `dispenserProxyAddress` sweep across every live-Dispenser consumer. Thanks for confirming (A) is swept.

DavidMinarsch (dismissed):
Approving. The fork test against live Tokenomics and Treasury is the right release gate for a change of this shape, and it's the coverage the unit suites structurally cannot provide — real inflation-curve position, real mint, real refund, real Timelock owner.

Two of my three items were fixed **later in the stack** rather than here, so I'm approving this diff knowing it is an intermediate state:

- **The `dispenserAddress` globals collision** — raised against this PR since it introduces the implementation/proxy split, and swept in `9bfd58f` on #314: every live-Dispenser consumer now reads `dispenserProxyAddress`. That was the mainnet-bricking one (`deploy_08_09_change_managers` would otherwise have pointed Tokenomics' and Treasury's `dispenser` at the implementation, and `refundFromStaking` gates on `dispenser == msg.sender`).
- **`deploy_07b` validating only `dispenserAddress`** — the `treasuryAddress` / `maxNumClaimingEpochs` / `maxNumStakingTargets` checks land in #314 too. That mattered more than a usual input-validation nit because #314 also makes the latter two set-once at `initialize()` with no runtime setter, so a bad value is an implementation redeploy.

So on this branch in isolation both gaps are still present. That's normal for a stack and I don't think it's worth reshuffling commits — but it does mean this PR shouldn't be merged to `main` independently of #314, which the stack ordering already enforces.

**Confirmed:** the CI-safety claim holds. The workflow's foundry step is a hardcoded list of `--mc` filters and `StakingClaimForkETH` matches none of them, so the fork-less CI doesn't pick it up. The flip side is the accepted-risk item on #309 — no fork suite runs in CI at all, here or in #315, so all three exist only as manual runs.

The body's "5 bytecode immutables" description of `deploy_07a_dispenser.sh` is superseded by #314's 3-arg constructor; worth a touch-up if the stack isn't squashed.

PR #314, opened 2026-08-03
refactor(Dispenser): drop default staking-param fallback + obsolete changeStakingParams

kupermind:
Stacked on #313. Addresses the review comments on #309 and the follow-up cleanup discussed there.

## What changed

**Remove the `defaultMinStakingWeight` / `defaultMaxStakingIncentive` immutables and their claim-time fallback.**
The fallback filled a `StakingPoint` with these defaults when `minStakingWeight == 0 && maxStakingIncentive == 0`. On a live Tokenomics that combination is unreachable when `stakingFraction > 0`:
- `Tokenomics.changeStakingParams` rejects zero for both parameters, and
- at epoch settlement the values are inherited from the previous (non-zero) epoch when not re-requested.

So the fallback was dead code. The block collapses to a plain `continue` when `stakingFraction == 0`.

**Remove `Dispenser.changeStakingParams` (and its `StakingParamsUpdated` event).**
It only set `maxNumClaimingEpochs` / `maxNumStakingTargets`, which are now established once at `initialize()`. No deployment or proposal script calls it. The Dispenser is proxy-upgradeable, so a future implementation can reintroduce a setter if one is ever needed.
> Note: this makes `maxNumClaimingEpochs` / `maxNumStakingTargets` set-once at `initialize()` — there is no longer a runtime setter for them.

**Constructor is now `(olas, tokenomics, retainer)`** — `deploy_07a_dispenser.sh` and every Dispenser test updated accordingly.

**Misc:** bump `Dispenser` + `DispenserProxy` pragma to `0.8.30`; add the `DispenserProxy` co-author; trim the over-explanatory storage/immutable comments flagged in the #309 review.

## Tests
- Forge unit: `forge test --mc Dispenser` — 17/17.
- Forge fork (proxied claim path vs live mainnet): `StakingClaimForkETH` — 3/3.
- Hardhat: full `yarn test` suite green. `DispenserFixes` now sets the staking params explicitly (mirroring the live invariant); the removed `changeStakingParams` DAO-setter test is dropped; the input-bound `Overflow` checks are re-exercised via a dedicated small-bounds `(1,1)` dispenser instance.
- solhint / eslint clean (no new errors).

## Out of scope (flagged)
- `abis/*/Dispenser.json` + `docs/configuration.json` still reference a stale Dispenser ABI (already stale since the #309 proxy refactor — old 9-arg constructor). ABI regeneration is a release-artifact step, handled at redeploy time.
- `Vulnerabilities_list_tokenomics.md` unchanged — those entries come off only after the physical redeploy + rewiring.

### DIFF
--- a/contracts/Dispenser.sol
+++ b/contracts/Dispenser.sol
@@ -369,8 +369,11 @@ contract Dispenser {
             revert AlreadyInitialized();
         }
 
-        // Check for at least one zero contract address
-        if (_treasury == address(0) || _voteWeighting == address(0)) {
+        // Check for a zero Treasury address. The Vote Weighting address is intentionally allowed to be zero
+        // here: on a fresh proxy deploy the Vote Weighting contract does not exist yet (it is deployed with
+        // this proxy's address baked in as an immutable), so it is wired in afterwards via changeManagers()
+        // while staking incentives stay paused. No claim path can run with a zero Vote Weighting.
+        if (_treasury == address(0)) {
             revert ZeroAddress();
         }
 
--- a/contracts/Dispenser.sol
+++ b/contracts/Dispenser.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.25;
+pragma solidity ^0.8.30;
 
 // Deposit Processor interface
 interface IDepositProcessor {
@@ -272,7 +272,6 @@ contract Dispenser {
     event ImplementationUpdated(address indexed implementation);
     event TreasuryUpdated(address indexed treasury);
     event VoteWeightingUpdated(address indexed voteWeighting);
-    event StakingParamsUpdated(uint256 maxNumClaimingEpochs, uint256 maxNumStakingTargets);
     event IncentivesClaimed(address indexed owner, uint256 reward, uint256 topUp, uint256[] unitTypes, uint256[] unitIds);
     event StakingIncentivesClaimed(address indexed account, uint256 chainId, bytes32 stakingTarget,
         uint256 stakingIncentive, uint256 transferAmount, uint256 returnAmount);
@@ -292,14 +291,9 @@ contract Dispenser {
     // Maximum chain Id as per EVM specs
     uint256 public constant MAX_EVM_CHAIN_ID = type(uint64).max / 2 - 36;
 
-    // Immutables live in the implementation bytecode, not proxy storage: they are re-settable by deploying
-    // a new implementation and calling changeImplementation. Every new implementation deploy must pass the
-    // correct constructor args — a wrong value silently rewires the proxy.
-    uint256 public immutable defaultMinStakingWeight;
-    uint256 public immutable defaultMaxStakingIncentive;
     // OLAS token address
     address public immutable olas;
-    // Tokenomics proxy address (stable by construction — the proxy address never changes)
+    // Tokenomics proxy address
     address public immutable tokenomics;
     // Retainer address in bytes32 form
     bytes32 public immutable retainer;
@@ -341,31 +335,14 @@ contract Dispenser {
     /// @param _olas OLAS token address.
     /// @param _tokenomics Tokenomics proxy address.
     /// @param _retainer Retainer address in bytes32 form.
-    /// @param _defaultMinStakingWeight Default min staking weight (bounded by uint16).
-    /// @param _defaultMaxStakingIncentive Default max staking incentive (bounded by uint96).
     constructor(
         address _olas,
         address _tokenomics,
-        bytes32 _retainer,
-        uint256 _defaultMinStakingWeight,
-        uint256 _defaultMaxStakingIncentive
+        bytes32 _retainer
     ) {
         // Check for at least one zero contract address
         if (_olas == address(0) || _tokenomics == address(0) || _retainer == 0) {
             revert ZeroAddress();
-        }
-
-        // Check for zero value staking parameters
-        if (_defaultMinStakingWeight == 0 || _defaultMaxStakingIncentive == 0) {
-            revert ZeroValue();
-        }
-
-        // Check for maximum values
-        if (_defaultMinStakingWeight > type(uint16).max) {
-            revert Overflow(_defaultMinStakingWeight, type(uint16).max);
-        }
-        if (_defaultMaxStakingIncentive > type(uint96).max) {
-            revert Overflow(_defaultMaxStakingIncentive, type(uint96).max);
         }
 
         olas = _olas;
@@ -373,8 +350,6 @@ contract Dispenser {
 
         retainer = _retainer;
         retainerHash = keccak256(abi.encode(IVoteWeighting.Nominee(retainer, block.chainid)));
-        defaultMinStakingWeight = _defaultMinStakingWeight;
-        defaultMaxStakingIncentive = _defaultMaxStakingIncentive;
     }
 
     /// @dev Initializes the dispenser proxy storage.
@@ -784,26 +759,6 @@ contract Dispenser {
         }
     }
 
-    /// @dev Changes staking params by the DAO.
-    /// @param _maxNumClaimingEpochs Maximum number of epochs to claim staking incentives for.
-    /// @param _maxNumStakingTargets Maximum number of staking targets available to claim for on a single chain Id.
-    function changeStakingParams(uint256 _maxNumClaimingEpochs, uint256 _maxNumStakingTargets) external {
-        // Check for the contract ownership
-        if (msg.sender != owner) {
-            revert OwnerOnly(msg.sender, owner);
-        }
-
-        // Check if values are zero
-        if (_maxNumClaimingEpochs == 0 || _maxNumStakingTargets == 0) {
-            revert ZeroValue();
-        }
-
-        maxNumClaimingEpochs = _maxNumClaimingEpochs;
-        maxNumStakingTargets = _maxNumStakingTargets;
-
-        emit StakingParamsUpdated(_maxNumClaimingEpochs, _maxNumStakingTargets);
-    }
-
     /// @dev Sets deposit processor contracts addresses and L2 chain Ids.
     /// @notice It is the contract owner responsibility to set correct L1 deposit processor contracts
     ///         and corresponding supported L2 chain Ids.
@@ -1020,15 +975,8 @@ contract Dispenser {
             ITokenomics.StakingPoint memory stakingPoint =
                 ITokenomics(tokenomics).mapEpochStakingPoints(j);
 
-            // Check for staking parameters to all make sense
-            if (stakingPoint.stakingFraction > 0) {
-                // Check for unset values
-                if (stakingPoint.minStakingWeight == 0 && stakingPoint.maxStakingIncentive == 0) {
-                    stakingPoint.minStakingWeight = uint16(defaultMinStakingWeight);
-                    stakingPoint.maxStakingIncentive = uint96(defaultMaxStakingIncentive);
-                }
-            } else {
-                // No staking incentives in this epoch
+            // No staking incentives in this epoch
+            if (stakingPoint.stakingFraction == 0) {
                 continue;
             }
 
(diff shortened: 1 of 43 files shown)


77ph (dismissed):
Reviewed as releasing auditor — **no blocking finding, safe cleanup.** Verified by hand + grounded on live mainnet (full write-up: internal19 doc #316).

- **The removed default-staking-param fallback is provably dead code on the live Tokenomics.** Three-way confirmation: `Tokenomics.changeStakingParams` rejects zero for both params (`Tokenomics.sol:767`); the checkpoint carries `min`/`max` forward unconditionally when not re-requested (`:1236–1237`); and on-chain the live Tokenomics proxy `0xc096362fa6f4A4B1a9ea68b1043416f3381ce300` (epochs 44–46) has `stakingFraction=75`, `maxStakingIncentive=60000e18`, `minStakingWeight=50` — all non-zero, so the `min==0 && max==0` trigger is absent and unreachable going forward. Even if that invariant were ever broken, the blast radius is benign — `minStakingWeight` is a threshold (`Dispenser.sol:1018`), not a divisor, and `max=0` caps `availableStakingAmount` at 0 → zero incentives, no div-by-zero / DoS / over-payment.
- **`Dispenser.changeStakingParams` has no live caller** — the two `changeStakingParams` references in `scripts/proposals/*` target the retained `Tokenomics.changeStakingParams`. `maxNumClaimingEpochs`/`maxNumStakingTargets` are set at `initialize()` (`:389–390`); set-once is fine given the proxy is upgradeable.
- **Storage-layout safe:** only immutables + a function + an event were removed (no storage variable), so the proxy layout is preserved; `DispenserProxy` is pragma + co-author only; `deploy_07a` correctly moves to the 3-arg `abi-encode`.
- **`forge test --mc Dispenser` → 17/17, re-run firsthand.** The `StakingClaimForkETH` fork test I did not re-run in-environment (it needs an archive mainnet RPC that isn't available to me right now) — but the live-state claim it exercises is independently grounded by the on-chain reads above.

One non-blocking hardening (**OBS-1**): the removal makes an implicit cross-contract invariant load-bearing (Tokenomics must never emit a `stakingFraction>0` StakingPoint with zero `min`/`max`). Worth a one-line NatSpec at the claim site or a note in `Vulnerabilities_list_tokenomics.md` so a future Tokenomics change surfaces the Dispenser dependency rather than silently distributing zero incentives.

Approving. The out-of-scope items you flagged (stale ABI regen, vuln-list clearing at redeploy) are correct to defer.

DavidMinarsch:
## Review: drop default staking-param fallback + obsolete `changeStakingParams`

Stack-level findings for the whole series are on #309.

### The dead-code argument is correct — but conditional, and the condition is asserted nowhere

I checked both legs of the justification and they hold:
- `Tokenomics.changeStakingParams` rejects zero for either parameter (`Tokenomics.sol:767`).
- Epoch settlement copies both forward when not re-requested (`Tokenomics.sol:1236-1237`).

So from the first `changeStakingParams` call onward, `minStakingWeight` / `maxStakingIncentive` are non-zero for every subsequent epoch. Good.

The gap is the window *before* that. `stakingFraction` is set independently via `changeIncentiveFractions`, so an epoch can exist with `stakingFraction > 0` and both staking params still zero — which is precisely the case the fallback was written for. With the fallback removed, such an epoch behaves very differently rather than reverting: `maxStakingIncentive == 0` clamps `stakingIncentive` to zero and dumps the entire amount into `returnAmount`, while `minStakingWeight == 0` disables the threshold check altogether.

What actually makes this safe is something the PR doesn't say: a fresh Dispenser's claim cursors all start at the current epoch, because `_addNominee` sets `mapLastClaimedStakingEpochs[nomineeHash] = ITokenomics(tokenomics).epochCounter()`. Historical epochs are therefore unreachable on the new deployment, so no zero-param epoch is ever traversed. (`retain()` reads neither parameter, so it is unaffected either way.)

Two asks:
1. State that reasoning in the PR / code comment — "the fallback was dead code" is true only given the greenfield-cursor property, and that property is what a future reviewer will need.
2. Add a deploy-time precondition to the runbook: assert `mapEpochStakingPoints(epochCounter).maxStakingIncentive != 0` on the live Tokenomics before wiring the new Dispenser. It is a one-line `cast call` and it converts the argument above from analysis into a check.

### `changeStakingParams` removal — get explicit DAO sign-off

This makes `maxNumClaimingEpochs` and `maxNumStakingTargets` set-once at `initialize()`. Defensible behind a proxy, and the body flags it — but it converts a one-transaction DAO knob into an implementation deploy plus a `changeImplementation` proposal. `maxNumClaimingEpochs` in particular is the lever you reach for when a claim becomes too expensive or a nominee's backlog grows, so losing the fast path has an operational cost. Worth naming that trade-off explicitly for the executive sign-off rather than carrying it as a cleanup.

Relatedly, this raises the stakes on the `deploy_07b` input validation gap flagged on #313: with no setter, a wrong value at `initialize()` is only fixable by redeploying the implementation.

### Confirmed

Constructor reduction to `(olas, tokenomics, retainer)` is consistently applied across `deploy_07a_dispenser.sh` and every Dispenser test; forge `--mc Dispenser` 19/19 on the final stacked state.

DavidMinarsch:
## Follow-up review — `cf881af` + `9bfd58f`

Finding (A) from #309 is swept properly, and the circular deploy is now genuinely broken rather than documented around. Verified: forge `--mc Dispenser` green on this head.

### Resolved

- **`dispenserProxyAddress` propagated to every live-Dispenser consumer** — `deploy_08_09_change_managers`, `deploy_10_14_change_ownerships`, `audit_chains/audit_contracts_setup` (all four assertion sites), all eight deposit-processor deploy scripts and their eight verify counterparts, `deploy_10_set_deposit_processors`, `fork/tokenomics_retain_calculation`, and both proposal scripts. That was the mainnet-bricking item — `Tokenomics.changeManagers` / `Treasury.changeManagers` now receive the proxy, so `refundFromStaking`'s `dispenser == msg.sender` gate resolves correctly.
- **`initialize` accepts a zero `voteWeighting`** — this is the right fix for the circular deploy, and better than the placeholder-address workaround. I checked the load-bearing claim in the comment ("No claim path can run with a zero Vote Weighting") and it holds through two independent barriers:
  1. the proxy initializes `Pause.StakingIncentivesPaused`, so `claimStakingIncentives` / batch revert `Paused()`;
  2. `addNominee` requires `msg.sender == voteWeighting`, which is unreachable while it is zero — so every `mapLastClaimedStakingEpochs` cursor is 0 and `_getClaimedEpochCounters` reverts `ZeroValue` regardless of the pause state.
  The #310 pause gate on `changeManagers` also lines up: a fresh proxy is already `StakingIncentivesPaused`, so the wiring call passes without an extra pause step.
- **`script_dispenser_change_managers.sh`** validates its inputs and asserts the post-condition (`voteWeighting()` read-back, `exit 1` on mismatch). Good — that is the pattern the other scripts should follow.
- **`deploy_07b`** now validates `treasuryAddress`, `maxNumClaimingEpochs` and `maxNumStakingTargets` (the #313 item), which matters more than usual now that the latter two are set-once with no runtime setter.
- **`deploy_07a` locks the implementation** (the #309 item). The reasoning is sound — the impl's storage is never used by the proxy, so the throwaway values are harmless and it just removes the question.

### New, low — nothing enforces the wiring before unpause

`setPauseState` has no `voteWeighting != address(0)` check, so an operator can unpause before running `script_dispenser_change_managers.sh`. It fails safe (claims revert `ZeroValue` per barrier 2 above rather than doing anything harmful), but the deployment invariant is currently *emergent* rather than enforced. A one-line guard on the unpause transition — or on `setDepositProcessorChainIds`, whichever you consider the "going live" step — would make it explicit and cheap.

### New, low — the impl-lock is unasserted

The lock step ends with `cast send … | grep -i "status"` and never checks the receipt, so a failed lock leaves the implementation unlocked and the script still exits 0. The inline comment already calls this best-effort and the impl genuinely cannot affect the proxy, so this is fine as-is — flagging only because the same `grep`-without-assert pattern is what the post-deploy checks elsewhere in this PR moved away from.

### Cross-repo note

`autonolas-governance` #215's internal21 CORRECTION currently prescribes breaking the cycle with a *placeholder non-zero* `voteWeighting`, which `cf881af` supersedes. Flagged there too — the two documents of record should agree on one procedure before the external re-audit reads both.

kupermind:
Thanks for the re-verify. The two documentation asks from the earlier round are done: the OBS-1 cross-contract invariant (Tokenomics never emits `stakingFraction > 0` with zero min/max) and the greenfield-cursor reasoning are in `0396533` (#315), and the Phase-0 `maxStakingIncentive != 0` / `minStakingWeight != 0` precondition is added to the migration runbook.

On your two new lows — the missing `voteWeighting != 0` guard on the unpause / going-live transition, and the unasserted impl-lock receipt — and the cross-repo internal21 supersede (governance #215 still prescribing the placeholder non-zero VW): noted, reviewing how to handle each; will follow up.

kupermind:
Follow-up items addressed in `cfceafb`:
- **Go-live guard:** `setPauseState` now reverts `ZeroAddress` on any transition to a staking-active state (not `StakingIncentivesPaused`/`AllPaused`) while `voteWeighting == 0` — the deploy invariant is now enforced, not emergent.
- **Impl-lock assert:** `deploy_07a` now checks the lock tx receipt status and surfaces a failure loudly (still non-fatal — the impl cannot affect the proxy).
- **Immutable-vs-storage rationale (#309:337):** documented on the constructor NatSpec.

Cross-repo: no longer an issue — governance `audits/internal21/README.md` was corrected post-merge (`5b31eb6`, 2026-08-13) with a "Superseded 2026-08-13" note documenting the shipped VW=0 order (impl → proxy VW=0 → `VoteWeighting(ve, proxy)` → `script_dispenser_change_managers.sh` → processors → unpause). The two documents of record now agree. forge 18/18, fork 3/3, hardhat 192.

DavidMinarsch (dismissed):
Approving. Forge suites verified locally on this head: **18/18** (Dispenser 3, DispenserProxy 10, DispenserFixes 5).

**The follow-up hardening is exactly right.** `setPauseState` now rejects any transition to a staking-active state while `voteWeighting == address(0)`, which turns the deploy invariant from emergent into enforced — claims already failed safe, but "fails safe by accident of three unrelated checks" is not something a future maintainer can rely on. The impl-lock asserts its receipt instead of grepping it, and the constructor NatSpec now records the immutable-vs-proxy-storage rationale (fixed identities in bytecode; repointable managers and init-time bounds in proxy storage).

**Allowing a zero `voteWeighting` at `initialize` is the right call** and better than the placeholder-address workaround it replaces. I checked the load-bearing claim and it holds through two independent barriers: the proxy initialises `StakingIncentivesPaused`, and `addNominee` requires `msg.sender == voteWeighting`, unreachable at zero — so every cursor is 0 and `_getClaimedEpochCounters` reverts `ZeroValue` regardless of pause state. The new `setPauseState` guard is now a third.

**The globals sweep is complete.** `9bfd58f` covers all four `audit_contracts_setup` assertion sites, both `changeManagers` targets, eight deposit-processor deploy scripts and eight verify counterparts, set-processors, ownership, retain-calc and both proposals.

### One item not addressed, non-blocking

I asked for two things on the dead-code argument and neither landed: the reasoning isn't recorded in the code, and there's no deploy-time precondition check.

The change **is** safe, and I verified that independently rather than taking the body's word: `Tokenomics.changeStakingParams` rejects zero (`:767`) and settlement inherits both params forward (`:1236-1237`), but the property that actually makes the fallback removal safe is different — a fresh Dispenser's cursors all start at the current epoch, because `_addNominee` sets `mapLastClaimedStakingEpochs[nomineeHash] = epochCounter`. Historical epochs are therefore unreachable on this deployment, so a pre-configuration epoch carrying `stakingFraction > 0` with zero params can never be traversed. (`retain()` reads neither parameter, so it's unaffected either way.)

That reasoning is what a future maintainer will need, and "the fallback was dead code" doesn't convey it — the fallback was *reachable* code made unreachable by a deployment property. Worth a comment at the `stakingFraction == 0` continue, plus a one-line `cast call` in the runbook asserting `mapEpochStakingPoints(epochCounter).maxStakingIncentive != 0` on the live Tokenomics before wiring. Neither blocks merge.

Separately, the loss of `changeStakingParams` makes `maxNumClaimingEpochs` set-once — the lever you'd reach for if a claim ever becomes too expensive now costs an implementation deploy. The body flags it; just make sure that's a conscious DAO decision rather than a cleanup side effect.

PR #315, opened 2026-08-03
refactor(Dispenser): make calculateStakingIncentives view; move effects to the claim path

kupermind:
Stacked on #314. Makes `calculateStakingIncentives` a `view` function and moves all state effects up into the claim paths — the cleaner form of the item-12 fix discussed on #310/#315.

## What changed

**`calculateStakingIncentives` is now `view`.** It no longer checkpoints the nominee, sets `mapZeroWeightEpochRefunded`, or refunds. It returns a new sparse `zeroWeightEpochs[]` (indexed by `j - firstClaimedEpoch`; a non-zero slot is a zero-total-weight epoch — epoch numbers are never zero, so a zero slot means "not zero-weight") and folds each zero-weight epoch's incentive into `totalReturnAmount`. Consequences:
- A standalone call mutates nothing, so it can never strand a zero-weight epoch's inflation — the item-12 hazard is gone *by construction* (no gating/`eth_call`-from trick needed).
- It is directly `staticcall`-able for off-chain estimation.

**The claim paths own the effects.** `claimStakingIncentives` / `_calculateStakingIncentivesBatch` now checkpoint the nominee in Vote Weighting *before* the calculation and set `mapZeroWeightEpochRefunded` for the returned zero-weight epochs. In the batch orchestrator the flag is set **before the next target's call**, so the in-loop `mapZeroWeightEpochRefunded` skip dedupes zero-weight epochs shared across targets — each is refunded exactly once for the whole batch. The zero-weight amount rides the single aggregated `refundFromStaking` (no per-epoch refund calls).

**Rename** `_checkpointNomineeAndGetClaimedEpochCounters` → `_getClaimedEpochCounters`. It never checkpointed anything — `checkpointNominee` was always a separate call after it; "counters" are the claim cursors. NatSpec corrected.

**`checkpointNominee` ordering (corrected after review).** The real `VoteWeighting.checkpointNominee` **does** revert `NomineeDoesNotExist` for a never-added nominee (via `_getWeight`) — it is *not* a Curve-style no-op. So `checkpointNominee` is now guarded by `_requireClaimableNominee(stakingTarget, chainId)`, called **before** the checkpoint in both claim paths: it reverts `ZeroValue` when the Dispenser has no claim cursor, mirroring the first check of `_getClaimedEpochCounters`. This keeps `ZeroValue` the authoritative not-claimable error and ensures `checkpointNominee` is only ever reached for a live nominee — matching `retain()`, which already checks the cursor before checkpointing. `MockVoteWeighting.checkpointNominee` is restored to faithfully revert `NomineeDoesNotExist`; the guard means it is only reachable if the Dispenser and Vote Weighting ever disagree. (This replaces the earlier no-op-mock approach, which rested on the incorrect "real VW no-ops" premise flagged in review.)

## Tests
- Reworked the item-12 test: asserts the view call mutates nothing (flag unset, pot unchanged) and the claim refunds exactly once.
- Added `test_fix12_batchZeroWeight_dedupRefundsOnce` (two targets sharing a zero-weight epoch in one batch → refunded once) and `test_fix12_separateClaims_zeroWeight_noDoubleRefund` (cross-tx flag persistence).
- Added `test_claim_neverAddedNominee_revertsZeroValueBeforeCheckpoint` locking the guard ordering (claim reverts `ZeroValue`, not `NomineeDoesNotExist`); the existing reverse-order batch `ZeroValue` test now also catches a reorder regression.
- **Fork tests (this PR also adds the two migration-leg suites):** `test/StakingL1ProcessorForkETH.t.sol` (new Dispenser proxy vs a real `EthereumDepositProcessor`, ETH fork) and `test/StakingL2MigrateForkOP.t.sol` (L2 `pause → migrate → updateWithheldAmountMaintenance` on an Optimism fork; now uses `balance != withheld` so the restore assertion proves the emitted value is used). Both dodge the CI `--mc` allowlist by name, so they are manual-only.
- Added `test_setPauseState_requiresVoteWeighting_beforeGoLive` after rebasing onto #314 — exercises the merged-state interaction of zero-VW `initialize` + the go-live guard + `_requireClaimableNominee`.
- forge Dispenser 22/22, fork `StakingClaimForkETH` + `StakingL1ProcessorForkETH` 6/6, `StakingL2MigrateForkOP` 2/2, full `yarn test` 192 passing, solhint clean.

## ABI changes (for the redeploy regen list)

This PR changes the `calculateStakingIncentives` return tuple — it now returns a fifth value, the sparse `zeroWeightEpochs[]`. It is a `public` function used for off-chain estimation, so `abis/*/Dispenser.json` must be regenerated at redeploy, alongside the #314 constructor change and the #311 4-arg `Migrated` signature.

## Note

This branch was rebased onto the updated #314 (`cfceafb`) so the shipping combination — zero-VW `initialize` + `_requireClaimableNominee` + the go-live guard — is under test on one branch (commit hashes shifted accordingly).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -57,12 +57,14 @@ forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerObservationCardinalityGasE
 forge test -f $FORK_ETH_NODE_URL --mc UniswapPriceOracleETH -vvv
 forge test -f $FORK_ETH_NODE_URL --mc BuyBackBurnerUniswapETH -vvv
 forge test -f $FORK_ETH_NODE_URL --mc StakingClaimForkETH -vvv        # proxied Dispenser claim path vs live Tokenomics/Treasury
+forge test -f $FORK_ETH_NODE_URL --mc StakingL1ProcessorForkETH -vvv  # new Dispenser proxy vs real EthereumDepositProcessor (L1 migration leg)
 forge test -f $FORK_BASE_NODE_URL --mc LiquidityManagerBase -vvv      # Base
 forge test -f $FORK_BASE_NODE_URL --mc BalancerPriceOracleBase -vvv
 forge test -f $FORK_BASE_NODE_URL --mc BuyBackBurnerBalancerBase -vvv
 forge test -f $FORK_POLYGON_NODE_URL --mc BuyBackBurnerBalancerPolygon -vvv   # Polygon
 forge test -f $FORK_ARBITRUM_NODE_URL --mc BuyBackBurnerBalancerArbitrum -vvv  # Arbitrum
 forge test -f $FORK_OPTIMISM_NODE_URL --mc BuyBackBurnerTransferV3Optimism -vvv  # Optimism
+forge test -f $FORK_OPTIMISM_NODE_URL --mc StakingL2MigrateForkOP -vvv         # L2 target dispenser pause/migrate/withheld cutover (L2 migration leg)
 forge test -f https://forno.celo.org --mc LPSwapCeloForkTest -vvv             # Celo
 ```
 
--- a/contracts/Dispenser.sol
+++ b/contracts/Dispenser.sol
@@ -410,12 +410,14 @@ contract Dispenser {
         emit ImplementationUpdated(implementation);
     }
 
-    /// @dev Checkpoints specified staking target (nominee in Vote Weighting) and gets claimed epoch counters.
+    /// @dev Gets the claimable epoch counters for a nominee, bounded by the current and removal epochs.
+    /// @notice View only — it does NOT checkpoint the nominee in Vote Weighting (that is a separate
+    ///         checkpointNominee call in the claim/retain flow); "counters" here are the claim cursors.
     /// @param nomineeHash Hash of a Nominee(stakingTarget, chainId).
     /// @param numClaimedEpochs Specified number of claimed epochs.
     /// @return firstClaimedEpoch First claimed epoch number.
     /// @return lastClaimedEpoch Last claimed epoch number (not included in claiming).
-    function _checkpointNomineeAndGetClaimedEpochCounters(
+    function _getClaimedEpochCounters(
         bytes32 nomineeHash,
         uint256 numClaimedEpochs
     ) internal view returns (uint256 firstClaimedEpoch, uint256 lastClaimedEpoch) {
@@ -670,12 +672,24 @@ contract Dispenser {
                     revert WrongAccount(retainer);
                 }
 
+                // Checkpoint staking target nominee in the Vote Weighting contract before the (view) calculation
+                IVoteWeighting(voteWeighting).checkpointNominee(stakingTargets[i][j], chainIds[i]);
+
                 // Get the staking incentive to send as a deposit with, and the amount to return back to staking inflation
-                (uint256 stakingIncentive, uint256 returnAmount, uint256 lastClaimedEpoch, bytes32 nomineeHash) =
+                (uint256 stakingIncentive, uint256 returnAmount, uint256 lastClaimedEpoch, bytes32 nomineeHash,
+                    uint256[] memory zeroWeightEpochs) =
                     calculateStakingIncentives(numClaimedEpochs, chainIds[i], stakingTargets[i][j], bridgingDecimals);
 
                 // Write last claimed epoch counter to start claiming from the next time
                 mapLastClaimedStakingEpochs[nomineeHash] = lastClaimedEpoch;
+
+                // Mark zero-weight epochs refunded (amount already in returnAmount). Setting the flag here, before
+                // the next target's calculateStakingIncentives call, dedupes zero-weight epochs shared across targets
+                for (uint256 k = 0; k < zeroWeightEpochs.length; ++k) {
+                    if (zeroWeightEpochs[k] != 0) {
+                        mapZeroWeightEpochRefunded[zeroWeightEpochs[k]] = true;
+                    }
+                }
 
                 stakingIncentives[i][j] = stakingIncentive;
                 transferAmounts[i] += stakingIncentive;
@@ -807,9 +821,9 @@ contract Dispenser {
         mapLastClaimedStakingEpochs[nomineeHash] = ITokenomics(tokenomics).epochCounter();
 
         // Clear a possible removal epoch left from a previous nominee lifecycle: without this, a re-added
-        // nominee would permanently revert in _checkpointNomineeAndGetClaimedEpochCounters (the claim cursor
-        // just set above always satisfies firstClaimedEpoch >= epochRemoved). The Dispenser no longer relies
-        // on the upstream Vote Weighting enforcing "remove is final"
+        // nominee would permanently revert in _getClaimedEpochCounters (the claim cursor just set above always
+        // satisfies firstClaimedEpoch >= epochRemoved). The Dispenser no longer relies on the upstream Vote
+        // Weighting enforcing "remove is final"
         delete mapRemovedNomineeEpochs[nomineeHash];
 
         emit AddNomineeHash(nomineeHash);
@@ -923,20 +937,27 @@ contract Dispenser {
     /// @param chainId Chain Id.
     /// @param stakingTarget Staking target corresponding to the chain Id.
     /// @param bridgingDecimals Number of supported token decimals able to be transferred across the bridge.
+    /// @notice View only: the caller must checkpoint the nominee in Vote Weighting BEFORE calling this, and
+    ///         mark every returned zero-weight epoch as refunded (see `zeroWeightEpochs`). Because the
+    ///         function mutates nothing, a standalone call can never strand a zero-weight epoch's inflation.
     /// @return totalStakingIncentive Total staking incentive across all the claimed epochs.
-    /// @return totalReturnAmount Total return amount across all the claimed epochs.
+    /// @return totalReturnAmount Total return amount across all the claimed epochs (includes zero-weight epochs).
     /// @return lastClaimedEpoch Last claimed epoch number (not included in claiming).
     /// @return nomineeHash Hash of a Nominee(stakingTarget, chainId).
+    /// @return zeroWeightEpochs Sparse array over [firstClaimedEpoch, lastClaimedEpoch): a non-zero slot is a
+    ///         zero-total-weight epoch number the caller must set in `mapZeroWeightEpochRefunded` (epoch is
+    ///         never zero, so a zero slot means "not a zero-weight epoch"). Its incentive is in totalReturnAmount.
     function calculateStakingIncentives(
         uint256 numClaimedEpochs,
         uint256 chainId,
         bytes32 stakingTarget,
         uint256 bridgingDecimals
-    ) public returns (
+    ) public view returns (
         uint256 totalStakingIncentive,
         uint256 totalReturnAmount,
         uint256 lastClaimedEpoch,
-        bytes32 nomineeHash
+        bytes32 nomineeHash,
+        uint256[] memory zeroWeightEpochs
     ) {
         // Check for the correct chain Id
         if (chainId == 0) {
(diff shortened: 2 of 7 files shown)


77ph (dismissed):
Reviewed as releasing auditor — **no blocking finding; a net security improvement.** Full write-up folded into the internal19 doc (#316, PART B).

- **Genuinely `view`, compiler-enforced.** `IVoteWeighting.nomineeRelativeWeight` is `external view` (`Dispenser.sol:189`), the state-mutating `checkpointNominee` is hoisted out to both claim paths, and the function compiles as `public view` — so no SSTORE / state-mutating call remains inside.
- **Exactly-once refund preserved.** The loop keeps its head guard `if (mapZeroWeightEpochRefunded[j]) continue;`. Because the batch sets the flag **before the next target's calc**, a zero-weight epoch shared across targets lands in exactly one target's `returnAmount` (later targets skip it), and a second claim tx skips an already-refunded epoch too. The batch refund aggregates via `totalAmounts[2] += returnAmount` → a single `refundFromStaking`; no per-epoch refund calls, no double refund, no stranded inflation.
- **Encoding collision-free.** `zeroWeightEpochs` is sized `lastClaimedEpoch − firstClaimedEpoch`; epochs are always ≥ 1, so the `0 = not-a-zero-weight-epoch` sentinel can't collide with a real epoch.
- **Item-12 hazard gone by construction.** A standalone external (view) call now mutates nothing, so it can neither mark-refunded-without-refunding nor strand inflation, and it's directly `staticcall`-able for estimation. Effect-correctness is confined to the two in-contract claim paths, both of which set-flag-and-refund together (verified). Strictly safer than the previous atomic form.
- **`checkpointNominee` hoist is production-behavior-neutral.** On a non-existent nominee the real Vote Weighting `checkpointNominee` no-ops (Curve-style zero-init weight points), so the authoritative revert stays the Dispenser's `mapLastClaimedStakingEpochs` / `ZeroValue` gate — a non-existent claim still reverts. The removed `MockVoteWeighting` revert was unfaithful to the real contract; the change only aligns the test mock (test-only).
- **`forge test --mc Dispenser` → 19/19, re-run firsthand**, including `test_fix12_viewCalculate_zeroWeight_refundsOnceViaClaim`, `test_fix12_batchZeroWeight_dedupRefundsOnce`, and `test_fix12_separateClaims_zeroWeight_noDoubleRefund` — the three tests that directly exercise the exactly-once analysis. (Fork test not re-run in-environment — needs an archive mainnet RPC.)

Approving. No new finding; OBS-1 from #314's review (document the cross-contract non-zero-min/max invariant) still applies to this shared claim loop.

DavidMinarsch:
## Review: `calculateStakingIncentives` → view, effects on the claim path

Stack-level findings for the whole series are on #309. This is the strongest PR in the stack — moving the effects up and making the calculation `view` is the right shape for item 12, and it removes the hazard by construction rather than by gating.

### Verified

- Sparse-array indexing is in bounds. `_getClaimedEpochCounters` guarantees `lastClaimedEpoch > firstClaimedEpoch` (the `firstClaimedEpoch == eCounter` revert plus the removal-epoch bound both leave at least one epoch), so `new uint256[](lastClaimedEpoch - firstClaimedEpoch)` cannot underflow and `j - firstClaimedEpoch` stays within range.
- The zero sentinel is sound: `firstClaimedEpoch == 0` reverts `ZeroValue`, so every recorded `j >= 1` and a zero slot is unambiguous.
- The batch dedupe works as described. The flag is written before the next target's `calculateStakingIncentives` call, and the loop's `mapZeroWeightEpochRefunded[j]` skip (`:991`) then excludes the epoch from that target's `totalReturnAmount` — so a zero-weight epoch shared across targets is refunded exactly once, both within a batch and across separate transactions.
- forge `--mc Dispenser` 19/19.

### [Medium] The `MockVoteWeighting` change is justified on a false premise

The new comment states the mock now mirrors "the real Curve-style gauge controller (it operates on zero-initialized weight points and does not revert)". The real contract does revert. `VoteWeighting.checkpointNominee` (`autonolas-governance/contracts/VoteWeighting.sol:403`) calls `_getWeight`, which begins:

```solidity
if (mapRemovedNominees[nomineeHash] == 0 && mapNomineeIds[nomineeHash] == 0) {
    revert NomineeDoesNotExist(account, chainId);
}
```

Three consequences:

1. The mock was made **less** faithful to the real contract, not more.
2. Because this PR hoists `checkpointNominee` ahead of the `_getClaimedEpochCounters` cursor check, a claim for a never-added nominee now reverts `NomineeDoesNotExist` from VoteWeighting — **not** `ZeroValue` from the Dispenser, as the PR body asserts. That is a changed error surface on a public entry point.
3. The suite can no longer catch it, since the mock is exactly the component that would have.

Either restore the mock's revert and update the expected error in the affected tests, or keep the no-op mock and correct both the code comment and the body claim. Worth resolving before merge given the body currently documents the opposite behaviour.

### [Low] `calculateStakingIncentives` ABI break

The added fifth return value (`zeroWeightEpochs`) changes the return tuple of a `public` function used for off-chain estimation. Fine on a redeploy, but it belongs on the ABI-change list alongside the constructor change from #314 and the `Migrated` signature change from #311 — see (E) on #309.

### [Low] Undisclosed scope

`test/StakingL1ProcessorForkETH.t.sol` (284 lines) and `test/StakingL2MigrateForkOP.t.sol` (120 lines) are added here but appear nowhere in the PR body, which describes only the reworked item-12 test and two additions. They look like the migration-leg coverage implied by #313 and are welcome — but 400 lines of new fork tests arriving unannounced in a security stack is worth a body edit. Note also that neither runs in CI: see (D) on #309.

### Naming

The `_checkpointNomineeAndGetClaimedEpochCounters` → `_getClaimedEpochCounters` rename is a genuine correctness fix to the name — the function never checkpointed anything. Good catch.

kupermind:
Fixed in `7935707` — thanks, this was a real one (and it was my no-op mock change that introduced it).

Verified you're right: `checkpointNominee → _getWeight` reverts `NomineeDoesNotExist` for a never-added nominee, so the real VoteWeighting is not a Curve-style no-op. Rather than accept the changed error surface, I reordered to match `retain()` (which already checks the cursor before checkpointing):

- New `_requireClaimableNominee(stakingTarget, chainId)` reverts `ZeroValue` when the Dispenser has no claim cursor, called **before** `checkpointNominee` in both claim paths. So `ZeroValue` stays the authoritative not-claimable error and `checkpointNominee` is only ever reached for a live nominee.
- `MockVoteWeighting.checkpointNominee` restored to faithfully revert `NomineeDoesNotExist`; the guard means it's only reachable if Dispenser and VW ever disagree.
- Added `test_claim_neverAddedNominee_revertsZeroValueBeforeCheckpoint` (asserts the mock reverts when checkpointed directly, yet the claim reverts `ZeroValue`); the existing reverse-order batch `ZeroValue` test now also catches a reorder regression, since the mock reverts again.

Body corrected, and the two fork-test files are now disclosed there too. Also noted to @77ph, whose approval had repeated the no-op premise.

forge Dispenser 20/20, `StakingClaimForkETH` + `StakingL1ProcessorForkETH` fork 6/6, hardhat 192 passing.

DavidMinarsch:
## Follow-up review — `8e7ce41` + `8071297`

The mock finding is fixed the right way round: rather than editing the body to match a weakened mock, you made the body's claim true. Verified: forge `--mc Dispenser` **20/20** on this head.

### Resolved

`MockVoteWeighting.checkpointNominee` now reverts `NomineeDoesNotExist` for an unregistered nominee, faithfully mirroring the real `VoteWeighting._getWeight`. `_requireClaimableNominee` then guards the call in both claim paths so the Dispenser's own `ZeroValue` stays the authoritative not-claimable error — which is exactly what the PR body originally asserted.

Two things I checked rather than assumed:
- **Removed nominees still checkpoint.** `_requireClaimableNominee` passes for a removed nominee (its cursor is non-zero), and the real `_getWeight` short-circuits on `mapRemovedNominees[hash] != 0`, so it does not revert. Claiming residual pre-removal epochs still works.
- **`retain()` needs no guard.** It already calls `_getClaimedEpochCounters` (same `firstClaimedEpoch == 0` check) before `checkpointNominee`, so the ordering the new comment describes was already correct there.

`8071297` also fixes the #311-style nit here: with balance deliberately exceeding withheld, the restore assertion now genuinely proves the emitted value is used.

### New, medium — this branch is not rebased onto the updated #314

`cf881af` (allow zero `voteWeighting` at `initialize`) is not an ancestor of this branch, so neither head carries both changes:

| branch head | `initialize` rejects zero VW | has `_requireClaimableNominee` |
|---|---|---|
| `refactor/dispenser-drop-default-staking-params` (#314) | no | no |
| `refactor/dispenser-view-calculate-staking-incentives` (#315) | **yes** | yes |

A GitHub merge after #314 lands will resolve cleanly — the two changes touch different lines — so this will not ship broken. The problem is that **no branch currently under test contains the combination that will actually ship**, and the deploy scripts in #314 already assume the zero-VW `initialize` that this branch still rejects.

Worth a rebase plus one test, because the combination is where the two changes interact *well*: on a freshly initialized proxy with `voteWeighting == address(0)`, `_requireClaimableNominee` fires before anything touches the zero address, so the failure is a clean `ZeroValue` rather than an opaque revert on a call to `address(0)`. That is a genuinely nicer property of the merged state than of either branch alone, and right now nothing demonstrates it.

### Unchanged from the last round

The `calculateStakingIncentives` fifth-return-value ABI break and the two undisclosed fork suites (`StakingL1ProcessorForkETH`, `StakingL2MigrateForkOP`) still warrant a body edit — the latter now also carries the improved `balance != withheld` assertion, which is worth mentioning since it is the fix to a review item.

kupermind:
Docs done in `0396533` (OBS-1 invariant, greenfield-cursor at `_addNominee`, and the corrected item-8 `changeManagers` wording). Thanks for checking the removed-nominee and `retain()` paths.

Good catch on the rebase-onto-#314: you're right that no branch currently carries both the zero-VW `initialize` (`cf881af`) and `_requireClaimableNominee`, and that combination is exactly where they interact well — on a fresh proxy with `voteWeighting == 0`, `_requireClaimableNominee` fires a clean `ZeroValue` before anything touches the zero address. Evaluating the rebase + a test that demonstrates it, plus the body edit (the `calculateStakingIncentives` 5th-return ABI break and the two fork suites, one of which now also carries the `balance != withheld` fix). Will follow up.

kupermind:
Done — this branch is now rebased onto the updated #314 (`cfceafb`), so the shipping combination is under test on one branch (commit hashes shifted). Added `test_setPauseState_requiresVoteWeighting_beforeGoLive`: with the new go-live guard, a zero-VW proxy is actually stopped at the *unpause* step (barrier 1), and `_requireClaimableNominee` → `ZeroValue` on a claim remains barrier 2 — an even cleaner merged-state property than either branch alone. Body updated with the `calculateStakingIncentives` fifth-return ABI note and the two fork suites (one now carrying `balance != withheld`). forge Dispenser 22/22, ETH fork 6/6, OP fork 2/2, hardhat 192.

DavidMinarsch (dismissed):
Approving — the strongest PR in the stack, and every item I raised is closed. Verified locally on this head: **22/22** (Dispenser 3, DispenserProxy 11, DispenserFixes 8).

**The `view` + caller-owns-effects design is the right shape for item 12.** It removes the hazard by construction rather than by gating, and I checked the mechanics rather than the description:
- sparse-array indexing is in bounds — `_getClaimedEpochCounters` guarantees `lastClaimedEpoch > firstClaimedEpoch` (the `firstClaimedEpoch == eCounter` revert and the removal-epoch bound each leave at least one epoch), so `new uint256[](last - first)` can't underflow and `j - firstClaimedEpoch` stays in range;
- the zero sentinel is sound — `firstClaimedEpoch == 0` reverts `ZeroValue`, so every recorded `j >= 1`;
- the batch dedupe works as claimed — the flag is written before the next target's call and the loop's `mapZeroWeightEpochRefunded[j]` skip then excludes that epoch from the next target's `totalReturnAmount`, so a shared zero-weight epoch is refunded exactly once, within a batch and across transactions.

**The mock finding was fixed the right way round.** Rather than editing the body to match a weakened mock, you made the body's claim true: `checkpointNominee` now reverts `NomineeDoesNotExist` faithfully, and `_requireClaimableNominee` keeps `ZeroValue` as the authoritative not-claimable error. Two things I checked rather than assumed — removed nominees still checkpoint fine (the real `_getWeight` short-circuits on `mapRemovedNominees[hash] != 0`, so pre-removal epochs remain claimable), and `retain()` needed no guard because it already calls `_getClaimedEpochCounters` before `checkpointNominee`.

**The rebase closed the merge-hygiene gap.** The shipping combination — zero-VW `initialize` plus `_requireClaimableNominee` — is now on one branch and under test. `test_setPauseState_requiresVoteWeighting_beforeGoLive` covers what I asked for and a bit more: both staking-active states rejected, re-pause transitions still allowed at zero VW, then wire-and-go-live. Pairing it with the barrier-2 claim test is the right split.

**The `_checkpointNomineeAndGetClaimedEpochCounters` → `_getClaimedEpochCounters` rename** is a genuine correctness fix to the name — the function never checkpointed anything.

Two carry-overs for the stack rather than this PR: the ABI change list (this PR's fifth return value, #314's 3-arg constructor, #311's 4-arg `Migrated`) is tracked as a release-artifact step on #309, and the three fork suites — including the two added here — run only manually. Both are recorded as accepted risk on #309, not oversights.

PR #316, opened 2026-08-04
audit(internal19): Dispenser default-staking-param cleanup (PR #314) — GREEN-LIGHT

77ph:
Standalone internal-audit record for **#314** (`refactor(Dispenser): drop default staking-param fallback + obsolete changeStakingParams`). Follow-up cleanup to the Dispenser-proxy rework (internal18), addressing the #309 review comments. Adds `audits/internal19/README.md`.

**Verdict: GREEN-LIGHT — no blocking finding.** Reviewed by hand + grounded on live mainnet.

- **Removed fallback is provably dead code on the live Tokenomics** — verified 3 ways: `Tokenomics.changeStakingParams` zero-reject (`:767`) + unconditional epoch carry-forward of `min`/`max` (`:1236–1237`) + on-chain reads of the live proxy `0xc096362fa6f4A4B1a9ea68b1043416f3381ce300` (epochs 44–46: `stakingFraction=75`, `maxStakingIncentive=60000e18`, `minStakingWeight=50`, all non-zero). Even if the invariant broke, the blast radius is benign (`min` is a threshold not a divisor; `max=0` caps incentives at 0) — no div-by-zero, DoS, or over-payment.
- **`Dispenser.changeStakingParams` has no live caller** (the two proposal references target the retained `Tokenomics.changeStakingParams`); `maxNumClaimingEpochs`/`maxNumStakingTargets` are set at `initialize()` (`:389–390`), set-once, recoverable via the proxy upgrade path.
- **Storage-layout safe:** only immutables + a function + an event were removed — no storage slot changed; `DispenserProxy` is pragma + co-author only (`PROXY_DISPENSER` slot + delegatecall untouched); `deploy_07a` updated to the 3-arg `abi-encode`.
- **`forge test --mc Dispenser` → 17/17 re-run firsthand.** The `StakingClaimForkETH` mainnet-fork test is the author's evidence; not re-run here (needs an archive mainnet RPC unavailable in-environment) — the live-state claim it checks is instead grounded by the on-chain reads above.

One Informational observation (**OBS-1**): the removal makes an implicit cross-contract invariant load-bearing — recommend documenting it at the claim site / in `Vulnerabilities_list_tokenomics.md`. Full detail in `audits/internal19/README.md`.

### DIFF
--- a/audits/internal19/README.md
+++ b/audits/internal19/README.md
@@ -6,10 +6,30 @@ head `refactor/dispenser-drop-default-staking-params` @ `86b35bf`, base
 `test/dispenser-deploy-forktest` @ `4fb597b`. Follow-up cleanup to the Dispenser-proxy rework
 (internal18) addressing the #309 review comments. Reviewed by hand + grounded on live mainnet.
 
+> **Amended 2026-08-13 by D. Minarsch** (original auditor no longer at the company). Two changes:
+> **§B.3 was factually wrong and is corrected in place** (see the correction block there), and the
+> coverage boundary below is now stated explicitly. Parts A and B are otherwise as originally written.
+>
+> **Coverage boundary.** Part A is pinned to `86b35bf` and Part B to `5a01465`. Both predate the
+> review-response commits that actually shipped, so the following are **covered by no audit**:
+>
+> | Commit | Change | Why it matters |
+> |---|---|---|
+> | `cf881af` | `initialize` accepts a zero `voteWeighting` | breaks the Dispenser ↔ VoteWeighting circular deploy; creates a live window where `voteWeighting == address(0)` |
+> | `cfceafb` | `setPauseState` rejects go-live while `voteWeighting == address(0)` | the guard that makes that window safe by construction rather than emergently |
+> | `7935707` | `_requireClaimableNominee` + `MockVoteWeighting` revert restored | the §B.3 correction |
+> | `0396533` | claim-loop invariant comments | closes OBS-1 |
+>
+> `cf881af` and `cfceafb` change contract behaviour and are material to deploy safety. They should be
+> picked up by the next audit round rather than assumed covered by this one.
+
 **Verdict: GREEN-LIGHT — no blocking finding.** The removals are safe cleanup: the deleted
 fallback is provably dead code on the live system, and no storage slot, live caller, or
 custody path is affected. One Informational observation (make the now-implicit cross-contract
-invariant explicit).
+invariant explicit) — **OBS-1, closed 2026-08-13 by `0396533`**.
+
+*Amendment note: the §B.3 conclusion below stands only as corrected in place; the reasoning
+originally given for it was wrong. See the correction block in §B.3.*
 
 ## Scope
 
@@ -54,6 +74,20 @@ not a divisor, and `maxStakingIncentive` is a *cap* (`:1028`). So a hypothetical
 `availableStakingAmount` at 0 → **zero incentives distributed** — no division-by-zero, no DoS, no
 over-payment. The removed net therefore only ever guarded a *benign* degradation, so removing it
 carries negligible risk even under invariant violation.
+
+> **Amended 2026-08-13 — a fourth, stronger reason the removal is safe.** The three checks above
+> establish the invariant *going forward*, and the on-chain reads cover epochs 44/45/46; epochs 1–43
+> are not inspected. That gap does not matter, because of a property not stated in the original
+> review: **a fresh Dispenser's claim cursors all start at the current epoch.** `_addNominee` sets
+> `mapLastClaimedStakingEpochs[nomineeHash] = ITokenomics(tokenomics).epochCounter()`, and
+> `_getClaimedEpochCounters` reverts `ZeroValue` on a zero cursor — so on this deployment no claim
+> path can ever traverse a pre-deployment epoch, whatever its staking point contains. The same holds
+> for the retainer, which is added through the identical path. `retain()` is unaffected either way,
+> as it reads neither `minStakingWeight` nor `maxStakingIncentive`.
+>
+> This is what makes the argument *complete* rather than merely well-evidenced: the fallback was not
+> unreachable code, it was reachable code made unreachable by a deployment property. Worth keeping in
+> mind if a future implementation ever seeds cursors from an earlier epoch.
 
 ## 2. Remove `Dispenser.changeStakingParams` + `StakingParamsUpdated` — SAFE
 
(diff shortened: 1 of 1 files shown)


77ph:
@kupermind — **pre-deploy checklist for the Dispenser rework**

To turn the two conditional GREEN-LIGHTs (internal18 → #309/#310/#311, internal19 here → #314/#315) into a clean pre-deploy sign-off, here is exactly what I need. It's short and I can close same-day, so the review doesn't become the thing holding the deploy. **Three blocking items, then release hygiene.**

### Blocking — needed before the deploy tx

**1. An archive mainnet RPC (or the CI fork-run log) for the #313 `StakingClaimForkETH` test.**
This is the single check I could not run myself. The three money-path tests — `test_proxiedClaim_realTreasuryMint_realInflationRefund`, `test_proxiedClaim_withheldNetting_refundsReusedInflation`, `test_migrationWiring_realContractsRepointed` — currently rest on your reported 3/3 plus my independent on-chain grounding of the invariants, not a firsthand run (my usual archive endpoint is unavailable right now). Give me a `FORK_ETH_NODE_URL` (archive node) and I'll run `forge test -f $FORK_ETH_NODE_URL --mc StakingClaimForkETH -vvv` directly; or, if it already ran in CI, point me at the run log.

**2. The exact deploy config for the immutables — because a wrong value here is unfixable except by a full redeploy.**
`deploy_07a_dispenser` bakes 5 bytecode immutables, and (cross-repo) the redeployed `VoteWeighting` now holds `dispenser` as an immutable. Before the tx I need to check, against the live addresses:
  - the three `Dispenser` constructor args — OLAS / Tokenomics proxy / retainer `bytes32`;
  - the `DispenserProxy.initialize()` args — `maxNumClaimingEpochs`, `maxNumStakingTargets` (set-once at init, no runtime setter after the #314 cleanup);
  - the deploy order + repoint plan (Tokenomics proxy → Dispenser impl+proxy → repoint) with the concrete target addresses, and confirmation the new `DispenserProxy` is the address `VoteWeighting.dispenser` is constructed against — i.e. the governance VoteWeighting redeploy and this Dispenser redeploy are sequenced so neither is left pointing at `AddressZero` or a stale address.

**3. The single commit/branch that will actually be compiled and deployed.**
The stack (#309 → #310 → #311 → #313 → #314 → #315) is all still open. Tell me the merged head (squash or a release branch) that goes to the compiler, so my sign-off maps to the exact bytecode that ships. I'll do a short delta re-verify on that head — a rebase or squash can move things, and I don't want the audited SHA and the deployed SHA to diverge.

### Release hygiene (part of the procedure, but not blocking my sign-off)
  - Regenerated `abis/**/Dispenser.json` for the 3-arg constructor, and `docs/configuration.json`.
  - `Vulnerabilities_list_tokenomics.md` items 8 / 9 / 10 / 12 / 25 flipped to closed **only after** the on-chain rewiring, per #313's own note.
  - OBS-1 from internal19 (one-line NatSpec of the "Tokenomics never emits `stakingFraction > 0` with zero min/max" invariant at the claim site) — optional hardening.

Send me items 1–3 and I'll close the pre-deploy review the same day.

77ph:
@kupermind — **blocking item 1 (fork run) is cleared: I ran `StakingClaimForkETH` firsthand.**

Ran it against an archive Ethereum-mainnet fork, and — importantly — on the **tip of the stack** (`refactor/dispenser-view-calculate-staking-incentives` @ `5a01465`, i.e. #309+#310+#311+#313+#314+#315 together), so this covers the **final code that ships**, not just #313 before the #314/#315 changes.

```
forge test -f <archive-mainnet> --mc StakingClaimForkETH -vv
[PASS] test_migrationWiring_realContractsRepointed()
[PASS] test_proxiedClaim_realTreasuryMint_realInflationRefund()
[PASS] test_proxiedClaim_withheldNetting_refundsReusedInflation()
3 passed; 0 failed
```

The money path is real, not a hollow pass — the trace on live contracts shows:
- settled epoch **46**, `StakingPoint{ stakingIncentive ≈ 947,986 OLAS, maxStakingIncentive = 60,000 OLAS, minStakingWeight = 50, stakingFraction = 75 }` — the full epoch incentive far exceeds the per-epoch cap, so the **cap binds**;
- **60,000 OLAS** minted by the live Treasury and delivered to the deposit processor — the test asserts `totalSupply` grew by exactly that;
- **`StakingRefunded(epoch 47, 887,986 OLAS)`** — the capped remainder refunded to live staking inflation, matching `epochIncentive − cap` exactly.

So internal19's one conditional item is closed firsthand, and the money-path holds with the `calculateStakingIncentives → view` refactor (#315) and the dropped default-staking-param fallback (#314) in place.

**One non-blocking hardening note:** the test forks at **latest** (`-f`, no pinned block). It passed cleanly now, but `assertGt(epochIncentive, 0)` makes it sensitive to the exact block/epoch state, so CI could flake at an inconvenient moment. Recommend pinning a fork block (`vm.createSelectFork(url, BLOCK)`) so the run is deterministic — same reasoning as the unpinned-fork drift we hit on the governance side recently. Not a code issue.

**Still needed from you before deploy (from the checklist above):**
- **2 — the exact deploy config for the immutables** (Dispenser 3 constructor args + `DispenserProxy.initialize()` args + deploy order/repoint with concrete addresses, incl. the VoteWeighting `dispenser`-immutable sequencing).
- **3 — the single commit/branch that will be compiled & deployed**, so my sign-off maps to the shipped bytecode (I'll delta-verify that head).

Send those two and the pre-deploy review closes.

DavidMinarsch (changes requested):
## Review: internal19 — Dispenser cleanup (PR #314 + #315)

**Part A is exemplary.** The dead-code argument is verified three independent ways, and grounding it on live reads of the Tokenomics proxy at `epochCounter = 46` (epochs 44/45/46 all carrying non-zero `maxStakingIncentive` / `minStakingWeight`) is exactly the right move — it converts an argument into evidence. The blast-radius analysis in §1 matches my own independent trace: a hypothetical `min == 0 && max == 0` would clamp the incentive to zero and disable the threshold, i.e. zero distribution, never over-payment. §B.2's encoding and dedup analysis also matches what I checked by hand (sparse-array bounds, the `≥ 1` epoch sentinel, and the flag-before-next-target ordering that makes the batch dedup exact).

That makes the one error below worth fixing rather than glossing.

### 1. §B.3 is factually wrong, and green-lights a design that has since been reverted

> On a non-existent nominee the real Vote Weighting `checkpointNominee` **no-ops** (Curve-style zero-initialized weight points — it does not revert) … The `MockVoteWeighting` revert that was removed was unfaithful to the real contract (and was previously unreachable via the Dispenser); the change only makes the test mock match production. **No production path is weakened.**

The real contract reverts. `VoteWeighting.checkpointNominee` (`autonolas-governance/contracts/VoteWeighting.sol:403`) calls `_getWeight`, which opens with:

```solidity
if (mapRemovedNominees[nomineeHash] == 0 && mapNomineeIds[nomineeHash] == 0) {
    revert NomineeDoesNotExist(account, chainId);
}
```

So the mock was made *less* faithful, not more — and because #315 hoisted `checkpointNominee` ahead of the Dispenser's own cursor check, a claim for a never-added nominee would have reverted `NomineeDoesNotExist` from Vote Weighting rather than `ZeroValue` from the Dispenser.

This was raised on #315, the author agreed, and it was fixed in `7935707`: the mock's revert is restored and `_requireClaimableNominee` was added so `ZeroValue` remains authoritative. **The audit therefore certifies, under "verified firsthand", a design that was subsequently reverted because it was wrong.** §B.3 needs rewriting to the shipped design, otherwise this document records a false negative on the one item in #315 that actually needed changing.

The root cause is worth noting for future rounds: the claim is about a contract in a *different repository*, and it was carried over from the PR description rather than checked against `autonolas-governance`.

### 2. The title claims #315 coverage that the pinned commits predate

Part A is pinned to `86b35bf` and Part B to `5a01465`. Both predate the review-response commits that actually shipped:

- `cf881af` — `initialize` accepts a zero `voteWeighting` (breaks the Dispenser ↔ VoteWeighting deploy cycle)
- `cfceafb` — `setPauseState` rejects going live while `voteWeighting == address(0)`
- `7935707` — `_requireClaimableNominee` + the mock fix above
- the rebase of #315 onto #314

So **no audit covers the zero-VW `initialize` or the go-live guard**, both of which are material to deploy safety and both of which changed contract behaviour. Either re-pin to the merged heads or state the coverage boundary explicitly in Scope.

### 3. Smaller

- **§B.4 "19/19"** — merged `main` is 22 (Dispenser 3, DispenserProxy 11, DispenserFixes 8).
- **OBS-1 is closed, not outstanding.** `0396533` (#315) added the invariant at `Dispenser.sol:1034-1038` on `main`, and the wording is accurate — it names the two Tokenomics mechanisms that uphold it and the benign failure mode. The verdict's "Address OBS-1 as hardening" should record it as done.

### 4. A strengthening for §1, not a correction

The dead-code argument rests on the forward invariant plus on-chain reads of epochs 44–46, which leaves epochs 1–43 unverified. That does not change the verdict, but only because of a property the document doesn't state: a fresh Dispenser's claim cursors all start at the current epoch, since `_addNominee` sets `mapLastClaimedStakingEpochs[nomineeHash] = ITokenomics(tokenomics).epochCounter()`. Historical epochs are therefore unreachable on this deployment regardless of what their staking points contain. Worth adding — it is what makes the argument complete rather than merely well-evidenced, and it is the reason the removal is safe even if some pre-configuration epoch does carry `stakingFraction > 0` with zero params.

DavidMinarsch:
Amended directly on this branch (`ca2243d`) rather than waiting on the original auditor, who has left. No contract change. Dismissing my request-changes accordingly — the substance below is what I asked for, applied.

**§B.3 corrected in place, not footnoted.** This section was wrong *at its own pin*, not merely outdated, and it carried a "verified firsthand" certification of a claim that was reverted in code before merge. A superseded-header treatment (as used on internal18) would have left the false statement standing as the section's conclusion, so the analysis is replaced with the shipped behaviour and the original wording is quoted inside a dated correction block so the audit trail survives. The corrected section records what the shipped code does — mock revert restored, `_requireClaimableNominee` guarding both claim paths so `ZeroValue` stays authoritative — plus two properties I verified against merged `main`: removed nominees still checkpoint (the real `_getWeight` short-circuits on `mapRemovedNominees != 0`), and `retain()` needed no guard because it already runs the cursor check first.

I also recorded the root cause, since it is the more useful lesson: the claim concerned a contract in **a different repository** and was carried over from the PR description rather than checked against `autonolas-governance`. That is the same failure mode as internal21's zero-`voteWeighting`-init claim on the governance side. Worth a standing rule for future rounds — cross-repo assertions get verified in the other repo or they do not go in the document.

**Coverage boundary now explicit.** Part A is pinned to `86b35bf`, Part B to `5a01465`, so a table in the header names what no audit covers: `cf881af` (zero-`voteWeighting` `initialize`) and `cfceafb` (`setPauseState` rejecting go-live while `voteWeighting == address(0)`). Both change contract behaviour and are material to deploy safety. Flagged for the next round rather than quietly assumed in scope.

**Smaller amendments.** OBS-1 recorded as closed by `0396533` (`Dispenser.sol:1034-1038` on `main`, wording verified accurate). B.4's 19/19 noted as correct for the pinned commit, with merged `main` at 22/22.

**One strengthening to Part A.** The dead-code argument rested on the forward invariant plus on-chain reads of epochs 44–46, leaving 1–43 uninspected. I added the property that actually closes that gap: a fresh Dispenser's cursors all start at `epochCounter` via `_addNominee`, so no claim path can traverse a pre-deployment epoch regardless of its staking point. The fallback was not unreachable code — it was reachable code made unreachable by a deployment property, which is worth knowing if a future implementation ever seeds cursors from an earlier epoch.

Part A's live-mainnet grounding and Part B's dedup/encoding analysis are unchanged and stand — both match what I verified independently.

DavidMinarsch (approved):
Superseded by `ca2243d` on this branch — the requested corrections are applied (see the comment above). Approving so the request-changes no longer blocks.

Note the branch is `BEHIND` `main` (the Dispenser stack merged this morning); it needs an update before merge. Only `audits/internal19/README.md` is touched, so no conflicts expected.

PR #318, opened 2026-08-13
fix(pol): internal20 R6 — tighten MAX_ALLOWED_DEVIATION 10% -> 2% (+ live-mainnet analysis)

DavidMinarsch:
Closes the open instruction in internal20 **R6** — *"review the 10% value against the shallowest POL pool's single-block manipulability"* — by actually running that review and recording the numbers in the audit. Documentation only; no contract or test change.

## Why the question is answerable in closed form

Read from Ethereum mainnet at block **25,747,913**:

- The POL to migrate is the Treasury's V2 OLAS/WETH LP in `0x09D1d767eDF8Fa23A64C51fa559E0688E526812F` — **99.986%** of the pair, i.e. **23,383,332 OLAS + 309.01 WETH ≈ 618 WETH**, spot `1.3215e-5` WETH/OLAS.
- The 0.3% V3 OLAS/WETH pool that `convertToV3` targets **does not exist yet**, and the only OLAS/WETH V3 pool today (1% tier, `0x18f7…f391`) has active `liquidity()` of **0**.

So the pool the guard protects will be seeded by this POL and effectively nothing else — **its depth is our own position**. "The shallowest POL pool" is therefore not a rule-of-thumb question.

## Cost to obtain the maximum deviation the gate permits

| deviation | capital in | round-trip fee @ 0.3% |
|---|---|---|
| 2% | 4.3 WETH | 0.03 WETH |
| **10% (gate limit)** | **21.1 WETH** | **0.13 WETH** |

The capital is flash-loanable and returns on the reversal, so buying the maximum permitted manipulation really costs **~0.13 WETH — about 0.02% of the position it attacks**.

## What it extracts

LM loss for an entry minted at a manipulated `slot0` (position unwound at the *true* price, plus leftovers swept to treasury), against the honest baseline of 618.02 WETH:

| slot0 deviation | LM loss | bps | attacker fee |
|---|---|---|---|
| ±1% | 0.01 WETH | 0.2 | 0.01 |
| ±2% | 0.04 WETH | 0.7 | 0.03 |
| ±5% | 0.25–0.30 WETH | 4–5 | 0.07 |
| **±10%** | **0.92–1.30 WETH** | **15–21** | 0.14 |

## Recommendation

**Tighten `MAX_ALLOWED_DEVIATION` to 2–3% before any POL is seeded.**

10% is too loose, though the exposure is smaller than the cost ratio alone suggests: an attacker nets ~0.8–1.2 WETH per entry for ~0.13 WETH of fees — profitable, but a few thousand dollars, and only against owner/DAO-gated operations (`convertToV3`, `increaseLiquidity`, `changeRanges`), so it is per-operation rather than continuous.

Loss scales roughly with `dev²`, so 10% → 2% cuts extraction **~30×** (21 bps → 0.7 bps) while the attacker's cost barely moves. The liveness price is near zero: on a pool that is ~100% our own liquidity, `slot0`-vs-30-minute-TWAP should sit well under 1% outside genuine volatility, and the failure mode is the self-healing revert-and-retry the audit already accepts as a residual.

Two reasons to decide this **now** rather than post-deploy:

1. **Loss scales linearly with POL size** — the table is a floor, not a ceiling.
2. **The constant lives in implementation bytecode**, so changing it costs an implementation deploy plus `changeImplementation`. Cheap today, before seeding; materially less so once the proxies are custodial.

## Note

This PR only records the analysis and the recommendation — it deliberately does **not** change the constant. That is a security-parameter decision for the DAO/exec, and R6's own framing ("treat any future change to it as a security change, not a tuning knob") argues for making it explicitly rather than folding it into a docs PR. Happy to follow up with the 2–3% change if that is the call.


kupermind:
The analysis answers R6 exactly, and the live-state framing (the target pool is ~100% our own POL, so "the shallowest POL pool" is literally this position) makes it a closed-form question rather than a rule of thumb. Endorse the direction.

One point that **strengthens** the case: post-#306.1 the entry `amount{0,1}Min` is anchored to `slot0` and is vacuous by construction (desired amounts and floor computed at the same `slot0` in the executing tx). So there is no amount-space backstop behind the deviation gate — `MAX_ALLOWED_DEVIATION` really is the *sole* entry manipulation defence, which is the whole premise of R6. The `dev²` loss scaling + the "it's a `constant`, so tightening costs a redeploy" caveat both argue for doing it **before** the proxies go custodial.

Two follow-on notes for whoever implements the tightening:

1. **Coupling to `liquidityManagerMaxSlippage`.** The POL deploy README ties the deploy-time `maxSlippage` to `MAX_ALLOWED_DEVIATION` (both 10%) so the amount-space floor doesn't reject mints the gate accepted. Post-#306.1 that coupling is cosmetic (the floor is vacuous), but if the gate drops to 2–3% the README note should be refreshed so the two don't read as contradictory.
2. **Scope.** This is a `LiquidityManagerCore` constant change — cleanest as its own small PR (tighten + README refresh + a gate-boundary regression test), independent of the LM mixin/rename refactor, which doesn't touch `Core` or the constant.

Happy to take the implementation PR (2% vs 3% is the one call worth pinning down first — 2% gives ~30× headroom over 10% and still sits well clear of the sub-1% slot0-vs-TWAP drift expected on a pool that is ~100% our liquidity).

kupermind:
Folded the R6 **fix** into this PR so the analysis and the remediation land together (title updated from `docs:` to reflect it now carries code).

**Change:** `MAX_ALLOWED_DEVIATION` 10% → **2%** (`2e16`) in `LiquidityManagerCore`.

**Feasibility — verified on every LM fork suite at 2%.** The worry was whether our shallow pools can hold a tight gate; they can, because the pre-warm keeps `slot0` well within 2% of the TWAP, so the happy path is unchanged from the 10% baseline:

- ETH (UniV2→UniV3) 24/25 — the one failure is the pre-existing unpinned-fork flake (reproduces at 10%)
- Base (Balancer→Slipstream) 9/9; Base (Balancer→UniV3) 1/1 (the latter on the sibling refactor branch)
- dead-band / fail-open / exit-sandwich, and the non-fork price-guard suites 18/18

The boundary tests were hard-coding the old 10% line — they now **read `MAX_ALLOWED_DEVIATION` from the contract**, so they hold at 2% or 3% without edits. (The one scary result mid-way — the dead-band sweep "reappearing" — was exactly that hard-coded 10% in the test's `withinGate`, not a real regression.)

**Deliberately NOT changed: the `liquidityManagerMaxSlippage` deploy param.** Post-#306.1 the entry `amount{0,1}Min` is slot0-anchored (vacuous), so `maxSlippage` no longer backstops mints — its only live role is the **source-side** `removeLiquidity`/`exitPool` floor. Reflexively tightening it to "match" the gate would risk reverting legit exits on a shallow source pool. Left the three globals at 10% and instead corrected the now-dead "set them equal" note in `LiquidityManagerCore` and the deploy README.

**Scope:** kept off the LM mixin/rename PR (#319) on purpose — that one is a pure behavior-preserving refactor, and this is a deliberate behavior change. It only touches `Core` + the boundary tests, which are independent of the rename, so the two don't conflict. `2%` vs `3%` is now a one-line change since the tests read the constant.

kupermind:
@DavidMinarsch — one call left that's yours as the R6 author: **2% vs 3%** for `MAX_ALLOWED_DEVIATION`.

Shipped **2%** here, on the reasoning in your addendum (loss scales ~`dev²`, so 10%→2% is ~30× less extraction while attacker cost barely moves). It passed every LM fork suite with the happy path unchanged from the 10% baseline, so 2% is not too tight for our shallow pools in the tested scenarios. **3%** buys a bit more liveness headroom if you're worried about legitimate `slot0`-vs-TWAP drift during real volatility on a thin pool.

Either way it's a one-line change now — the boundary tests read `MAX_ALLOWED_DEVIATION` from the contract rather than hard-coding it. Let me know which you'd like locked in.

DavidMinarsch:
## The 2% vs 3% call: **3%** — and a correction to the premise it rests on

Thanks for folding the fix in, and for making the boundary tests read `MAX_ALLOWED_DEVIATION` from the contract — that is what makes this a one-line decision instead of a test rewrite.

### First, a scope correction

It was put to me that the gate matters only for the one-time deployment conversion, so "as long as `convertToV3` goes through we're fine". That is not what the code does, and it changes the calculus. `MAX_ALLOWED_DEVIATION` is a **standing runtime parameter** on five paths:

| consumer | when | who |
|---|---|---|
| `convertToV3` (`:765`) | one-time migration | owner/DAO |
| `changeRanges` (`:840`) | ongoing repositioning | owner/DAO |
| `increaseLiquidity` (`:1058`) | ongoing top-ups | owner/DAO |
| **`BuyBackBurner.buyBack`** (`BuyBackBurner.sol:267`, `:476`) | **ongoing** | **permissionless** |
| `_getExitSqrtPrice` (`:1295`) | **withdrawals** | owner/DAO |

The last two are the ones that make this more than a deploy-day threshold. `buyBack` is permissionless and runs on its own cadence, and `_getExitSqrtPrice` enforces **the same constant**, so a tight gate also makes the audit's accepted residual — *"a genuine extreme move → exit temporarily reverts until re-convergence"* — materially more likely. Tightening buys security on entries and pays for it on exits and buy-backs.

### Why 3%

From the R6 addendum's model, the single trade that alone breaches each gate on the seeded pool, and the extraction ceiling if abused:

| gate | trade that breaches it | as % of POL | extraction ceiling |
|---|---|---|---|
| 2% | **4.3 WETH** | 0.70% | 0.037 WETH |
| 3% | **6.4 WETH** | 1.04% | 0.083 WETH |
| 10% (old) | 21.1 WETH | 3.42% | 0.920 WETH |

The security difference between 2% and 3% is **0.046 WETH** — noise. The liveness difference is 50% more headroom, and it lands on the paths above. A **4.3 WETH (~$15k) swap is an ordinary trade**, not a whale event: at 2%, one such trade blocks `buyBack`, `increaseLiquidity` and withdrawals for as long as the 30-minute TWAP takes to catch up. At 3% it takes half again as much flow to cause the same stall.

Both are ~25× and ~11× better than the 10% we are leaving, so the marginal security given up is small against the liveness bought. On a pool this thin — 618 WETH, ~100% our own liquidity — I would rather not have organic flow tripping a security gate on a permissionless path.

The asymmetry also improves with time: as POL grows, the flow needed to breach 3% grows proportionally (10× POL → 64 WETH), while extraction at 3% stays a fraction of a WETH. 3% ages better than 2% in the direction we expect to move.

**Locking in 3% (`3e16`).** Since the tests read the constant, that is the one-line change plus refreshing the two comments that quote 2%.

### On the rest of the change — agreed, and verified

Leaving `liquidityManagerMaxSlippage` at 10% is right, and the reasoning checks out: all three `maxSlippage` uses in `Core` (`:377-378` exit floor, `:419-420` `_slot0MinAmounts`, `:463-464` `_increaseLiquidity`) are anchored to `slot0` in the executing transaction, so every one is vacuous by construction. Its only live role is the source-side V2/Balancer exit floor, and reflexively tightening that to "match" a gate it no longer backstops would risk reverting legitimate exits on a shallow source pool. Correcting the stale "set them equal" note rather than propagating it is the right call.

I cannot approve this PR — I opened it — so it needs a second reviewer regardless of which value lands.

DavidMinarsch:
## Correction: **2% is right — my 3% call was wrong.** Lock in 2%, as shipped.

I argued for 3% on liveness grounds. Rather than keep arguing it, I measured the thing the argument rested on, and the data reverses it.

### What I measured

Sampled the live OLAS/WETH V2 pair hourly over ~2.5 days (59 points), computing spot vs its own 30-minute TWAP — the same quantity `_getPoolPriceFacts` compares. That pair *is* today's POL, so it is the same tokens and the same depth class rather than a proxy.

| statistic | deviation |
|---|---|
| median | 0.000% |
| p90 | 0.193% |
| p95 | 0.742% |
| p99 / max | **7.819%** |

Converting for depth: a constant-product pool behaves as V3 with `L = √(x·y)` = **85,004**, while the POL concentrated over `-25000/+15000` gives `L` = **119,141**. The V3 pool is **1.402×** deeper, so the same flow moves it `0.713×` as far, and a V3 gate of `G` corresponds to a measured V2 deviation of `G × 1.402`.

| V3 gate | equivalent V2 deviation | samples blocked |
|---|---|---|
| 2% | > 2.80% | 1 / 59 (1.7%) |
| 3% | > 4.21% | 1 / 59 (1.7%) |

### Why that settles it

**2% and 3% block identically.** The distribution is bimodal — 95% of observations under 0.74%, then a single excursion at 7.819% that breaches both. There are **zero** observations in the band where 3% would help and 2% would not; the largest sample below the 2% line is 2.153% (V2), i.e. 1.53% in V3 terms.

So 3% buys **no** liveness on this data while permitting ~2.2× the extraction (0.083 vs 0.037 WETH). My "a 4.3 WETH trade blocks buy-backs" point was arithmetically fine but practically empty: flow of that size against this pair simply is not landing in the window between the two gates.

And the one blocking sample is a **7.8% deviation** — precisely the condition the gate exists to refuse. That block is the gate working, not a false positive, and it is blocked at 2%, 3% and 5% alike.

Two things I got wrong that are worth naming: I treated the deviation distribution as roughly continuous when it is bimodal, and I leaned on a liveness cost that lands entirely on retryable operations — `buyBack` re-runs freely, and VL#26 already accepts re-submitting a decrease proposal that lands in an extreme window.

**Ship 2%.** No code change needed — it is what you already have.

### Recorded in the audit

Pushed `ee99e10` to this branch adding the measurement, the depth conversion, and the conclusion under the R6 addendum, with the sample caveats stated: 59 hourly points over ~2.5 days, and hourly snapshots undersample short excursions, so the absolute blocked-fraction is a floor for every gate. That undersampling is symmetric between the two candidates, and the finding that decides it is the *shape* — no mass between them — not the absolute rate. Worth re-running over a longer window if the value is ever revisited.

Still needs a second reviewer since I opened the PR.

kupermind (approved):
Confirmed the branch tip: `MAX_ALLOWED_DEVIATION == 2e16`, and `ee99e10` is doc-only (the measurement addendum).

The empirical reversal is the right way to settle 2% vs 3% — measuring the live pair rather than arguing the model. The finding that decides it is the **shape**: a bimodal distribution with no mass between the depth-adjusted gates means 3% buys zero liveness for ~2.2× the extraction, and the single 7.8% excursion is the gate correctly refusing an extreme move, not a false positive. And the liveness cost lands on retryable paths (`buyBack` re-runs; VL#26 already accepts re-submitting a decrease), so it's not load-bearing.

Agreed on leaving `liquidityManagerMaxSlippage` at 10% — all three `maxSlippage` uses in `Core` are `slot0`-anchored in-tx (vacuous), so it only governs the source-side exit floor and must not track the gate.

Approving as the second reviewer — **2% as shipped**, no code change needed.

PR #317, opened 2026-08-11
docs: external-audit candidate contracts

kupermind:
Stacked on #315. Compiles the not-externally-audited Solidity changed across the open PRs into a single external-audit scope document (`docs/external_audit_candidates.md`), in the style of the previous "not externally audited smart contracts" review.

## Scope — 4 contracts, 1817 SLoC (full SLoC + delta per contract)

**Governance** (`v1.3.0-pre-external-audit`):
- `VoteWeighting.sol` — 437 SLoC — +63/−39 (~102) — `removeNominee` accounting reconciliation / checkpoint-DoS fix (governance PR #215).

**Tokenomics** (`1.5.0-pre-external-audit`):
- `Dispenser.sol` — 705 SLoC — +159/−88 (~247) — proxy conversion + vuln items #8/#9/#12/#25 + `calculateStakingIncentives` → `view` (#309/#310/#314/#315).
- `DispenserProxy.sol` — 34 SLoC — new file (#309/#314).
- `LiquidityManagerCore.sol` — 641 SLoC — +158/−82 (~240) — fail-closed price guard + soft-priced exit floor (#306).

Both figures (whole-file SLoC and the delta vs the previously audited version) are listed per contract so the review can be scoped either way. `DefaultTargetDispenserL2` (#311) is intentionally excluded — a two-field event addition, not audit-worthy.

The two `*-pre-external-audit` tags are referenced as the scope anchors; they are to be cut at the merged heads of the respective stacks.

Docs-only change; no contract or test code touched.

PR #319, opened 2026-08-14
refactor(pol): source/target/burn LiquidityManager mixins + rename + Balancer->UniV3 manager

kupermind:
> **Note:** #318 (R6 gate tightening, `MAX_ALLOWED_DEVIATION` 10%→2%) is merged into this branch so the suite verifies under the final gate. The refactor commits below remain behavior-preserving; the tightening rides along via a merge of #318 (merge-commits dedupe on landing — merge #318 first). See the pinned comment for the 2% suite results.

## Summary

Decomposes the two `LiquidityManager` implementations into composable **source / target / burn** mixins over `LiquidityManagerCore`, renames the leaves by **capability** rather than chain, and adds the third combination — **Balancer V2 → Uniswap V3** — as pure composition with no new logic.

The leaves only ever varied along three orthogonal axes: which DEX the POL is withdrawn from, which concentrated-liquidity DEX it is minted into, and how OLAS is burned. Factoring those makes the missing combination free.

```
LiquidityManagerCore  (unchanged)
 ├ source:  SourceUniV2 | SourceBalancer   (shared TWAP fair-min math in SourceBase)
 ├ target:  TargetUniV3 | TargetSlipstream
 └ burn:    BurnViaBridge (L2); L1 direct burn inlined in the UniV2->UniV3 leaf

UniV2UniV3          = SourceUniV2    + TargetUniV3       + (L1 burn)     // was LiquidityManagerETH
BalancerSlipstream  = SourceBalancer + TargetSlipstream  + BurnViaBridge // was LiquidityManagerOptimism
BalancerUniV3 (new) = SourceBalancer + TargetUniV3       + BurnViaBridge // Polygon / Arbitrum / Base-UniV3
```

## Why decompose (rather than add a third leaf)

The immediate trigger was needing the **Balancer V2 → Uniswap V3** combination — and it is **not optional**. On L2s that hold their POL in **Balancer** and have **Uniswap V3** but **no Slipstream** (Velodrome/Aerodrome) — e.g. Polygon and Arbitrum — POL migration has *no other path*: `BalancerSlipstream` cannot run there (no Slipstream target), and `UniV2UniV3` needs a Uniswap-V2 POL source that those chains don't have. So `LiquidityManagerBalancerUniV3` is a **hard requirement** for liquidity migration on every such chain, not a convenience.

The naive way to build it — copy `LiquidityManagerOptimism`, swap its Slipstream mint for `LiquidityManagerETH`'s Uniswap V3 mint — would have — copy `LiquidityManagerOptimism`, swap its Slipstream mint for `LiquidityManagerETH`'s Uniswap V3 mint — would have **triplicated** code that was already duplicated across the two existing leaves. The status quo had:

- **A security-critical routine copy-pasted verbatim.** The TWAP-anchored fair-reserve `minAmountsOut` math (the overflow-safe `sqrt(k·twap)` that guards every POL withdrawal) was byte-identical in both `LiquidityManagerETH` and `LiquidityManagerOptimism`. A fix to it — e.g. the `mulDivDown` overflow guard — had to be made in *N* places, and a missed mirror is a silent divergence on the withdrawal path.
- **No way to scale.** The real axes of variation are **(source DEX) × (target DEX) × (burn mode)**, not chains. Chain-named leaves mean every new chain/combination is another full copy of ~300 lines, most of it shared.
- **Misleading names.** `LiquidityManagerETH` is really "UniV2 → UniV3" (valid on any such chain); `LiquidityManagerOptimism` is "Balancer → Slipstream" (Optimism *and* Base). The "Arbitrum or Polygon?" naming question for the new contract has no good answer under the chain scheme — one contract serves both.

Factoring those axes into mixins fixes all three:

- **Each concern lives in exactly one place.** A fix to the Balancer exit, the UniV3 mint, or the shared fair-min math (`SourceBase`) now applies to *every* manager that composes it, atomically — no mirror-and-drift on the security-critical path.
- **The Nth combination is free.** `LiquidityManagerBalancerUniV3` is pure composition — `SourceBalancer + TargetUniV3 + BurnViaBridge`, zero new logic. That is exactly why the new leaf in this PR carries no bespoke code to audit beyond its constructor wiring.
- **Smaller audit surface.** The shared source-side math is reviewed once (`SourceBase`), and each mixin is independently auditable — the external-audit-candidates doc lists them with that framing.

It is safe to do now because it is a **pure extraction** (bodies moved verbatim, behavior verified identical by the fork suites) with all storage still in `Core`, so `changeImplementation` upgrades are unaffected.

## Rename (capability, not chain)

Chain names were misleading — the UniV2→UniV3 manager is not ETH-specific, and one Balancer→UniV3 manager serves Polygon, Arbitrum and Base:

| was | now |
|---|---|
| `LiquidityManagerETH` | `LiquidityManagerUniV2UniV3` |
| `LiquidityManagerOptimism` | `LiquidityManagerBalancerSlipstream` |
| — | `LiquidityManagerBalancerUniV3` (new) |

Propagated to deploy scripts (+ a deploy script for the new leaf), the audit-chains verifier, ABIs, the fork script, and the POL docs / BBB chain-DEX matrix. Fork-test files keep their chain names (they are pinned to one chain's real pools).

## Behavior-preserving

The mixin split is a **pure extraction — no logic change**. Verified by the existing fork suites against the renamed contracts:

- **ETH** (UniV2→UniV3): 24/25 — the one failure is pre-existing unpinned-fork deviation-gate flakiness that reproduces identically on pre-refactor `main`.
- **Base** (Balancer→Slipstream): 9/9.
- The #306 proofs still green: dead-band (0 in-gate cells), fail-open (both triggers), exit-sandwich.

## New contract test

`LiquidityManagerBalancerUniV3` is composition of already-fork-proven halves (Balancer source on Base, UniV3 target on ETH); the only untested surface is the two in one transaction. `test/LiquidityManagerBalancerUniV3Base.t.sol` proves exactly that on a **Base fork**: withdraw the real OLAS/WETH Balancer POL and mint it into a canonical Uniswap V3 position via `convertToV3`, asserting the BPT is fully exited and one UniV3 position with real liquidity is held. **Passes.**

## Notes

- Storage layout is unchanged (Core owns all state; mixins add only immutables) — `changeImplementation` upgrades stay safe.
- No change to `LiquidityManagerCore`, the proxy, or the scanner.
- Pre-push checks run locally: `forge build`, `yarn compile` (156 files), the non-fork LM unit suites (18/18), solhint (0 errors), eslint (clean), gitleaks (no leaks).


kupermind:
## Merged in #318 (R6 gate tightening) — full suite verifies under the 2% gate

Pulled #318's `MAX_ALLOWED_DEVIATION` 10% → 2% tightening (+ its gate-agnostic boundary-test changes) into this branch so the whole suite runs under the **final** gate rather than the old 10%. All three managers pass at 2%:

| suite | result @ 2% |
|---|---|
| ETH (UniV2→UniV3) | 24/25 (the 1 is the pre-existing unpinned-fork flake) |
| Base (Balancer→Slipstream) | 9/9 |
| Base (Balancer→UniV3, new) | 1/1 |
| dead-band / fail-open / exit-sandwich | pass |
| non-fork price-guard suites | 18/18 |

The refactor commits themselves remain behavior-preserving; the tightening rides along via a **merge** of #318. This repo uses merge-commits (not squash), so the shared commits dedupe by SHA when both PRs land — no duplication. Cleanest merge order is **#318 first**, then this one.

DavidMinarsch (changes requested):
## Review

The "pure extraction, behaviour-preserving" claim is the whole PR, so I checked it with the compiler rather than by reading diffs. It holds, on every axis I could test mechanically.

**Storage layout is identical.** `forge inspect … storageLayout` on both trees:

| contract | layout |
|---|---|
| `LiquidityManagerETH` (main) | `owner`@0.0, `maxSlippage`@0.20, `_locked`@0.22, `mapPoolAddressPositionIds`@1 |
| `LiquidityManagerOptimism` (main) | identical |
| `LiquidityManagerUniV2UniV3` (PR) | identical |
| `LiquidityManagerBalancerSlipstream` (PR) | identical |
| `LiquidityManagerBalancerUniV3` (PR) | identical |

This was the risk I went looking for — mixin extraction changes C3 linearization, which is exactly how a refactor silently reorders slots under a live proxy. All six mixins declare **zero** state variables (immutables only, which live in bytecode), so `changeImplementation` is genuinely unaffected. The claim in the body is correct.

**External surface is unchanged.** Selector sets are identical across the renames: `LiquidityManagerETH` → `UniV2UniV3` **30/30**, `LiquidityManagerOptimism` → `BalancerSlipstream` **31/31**.

**The extracted security-critical math is faithful.** `SourceBase._fairMinAmountsOut` is line-for-line the old inline body — same `sqrt(mulDivDown(k, twap, 1e18))` pair, same branch, same `(liquidity · fairReserve · (MAX_BPS − maxSlippage)) / (totalSupply · MAX_BPS)`. The only change is parameterisation, which is where this kind of extraction usually breaks — so I checked both call sites, and each passes `tokens[0] == olas`, matching the original inline predicate. `SourceBalancer` also keeps `k = amounts[0] * amounts[1]`, identical to the old `LiquidityManagerOptimism`.

**The new leaf really is free.** `LiquidityManagerBalancerUniV3` contains a constructor and nothing else — no bespoke logic to audit, exactly as advertised.

Also: POL unit suites **18/18**; runtime sizes 20,959–21,286 bytes, comfortably under EIP-170 with ~3.3–3.6 KB of margin; and the ABI artifacts are properly renamed (old deleted, new added) rather than left to rot, which is a welcome change from the pattern on the Dispenser stack.

### 1. [Medium — process] #318 rides along, and the merge order is load-bearing but unenforced

This branch has #318 merged into it: `MAX_ALLOWED_DEVIATION = 2e16` is present here, and the diff against `main` includes internal20's R6 addendum. The body flags it —

> merge-commits dedupe on landing — **merge #318 first**

— but nothing enforces that, and #319 targets `main` directly. If #319 lands first, a change to **the sole entry manipulation defence** reaches `main` through a refactor PR, with #318's review thread bypassed entirely. #318 currently has **zero approvals** and is the PR where that change is supposed to be scrutinised.

The refactor itself is fine; the packaging is the problem. Cleanest fix is to rebase #319 off the merge so it is purely the mixin work, and let #318 carry the constant on its own merits. If you would rather keep the merge for suite-verification reasons, then #319 should not be merged until #318 is approved and landed — worth stating as a blocking condition rather than a note, since it is invisible to anyone reading the file diff.

### 2. [Low] The primary test harness keeps its chain name

`test/LiquidityManagerETH.t.sol` still carries the old name and is imported by three fork suites (`ExitFailOpenForkETH`, `ExitSandwichFork`, `IncreaseDeadBandForkETH`). In a PR whose thesis is *capability, not chain*, leaving the main harness chain-named is the one place the rename does not land — and it is the file someone will look for once the contract is `LiquidityManagerUniV2UniV3`.

### 3. [Low] `changeImplementation` cannot tell the three leaves apart

All three `deploy_02_*` scripts write the same `liquidityManagerAddress` globals key, and `script_05` points the proxy at whatever it holds. Since globals are per-network this is unambiguous in normal use, and the README does document the mapping ("`balancer_slipstream` (e.g. Optimism/Base), or `balancer_univ3` … (e.g. Polygon/Arbitrum)") — so this is a residual, not a gap.

But the capability rename makes running the wrong `deploy_02_*` for a chain a more plausible slip than it was under chain names, and `changeImplementation` accepts any address (the deliberate no-codesize-check convention), so a wrong-variant impl is accepted on-chain and fails at runtime instead of at deploy. `script_05` already reads back the `PROXY_LIQUIDITY_MANAGER` slot, which proves it pointed where globals said — not that globals held the right *variant*. A cheap addition: after the swap, `cast call` a variant-distinguishing immutable (`routerV2` for UniV2 source, the Balancer `vault` for the others) and assert it matches the chain's expectation.

Nothing in 2 or 3 blocks merge. Item 1 does, until the ordering is settled.

DavidMinarsch (dismissed):
Blocking item resolved by #318 landing — approving.

`#318` merged at 08:33 (`607d884`), and the merge deduped exactly as you said it would: `git diff origin/main...HEAD` for `contracts/pol/LiquidityManagerCore.sol`, `audits/` and `.github/` is now **empty**. This branch carries only the mixin work — the constant change reached `main` through its own PR with its own review, which was the whole concern.

I re-ran the mechanical checks against the **current** `main` (i.e. with the 2% gate in place on both sides, so the comparison has no confound):

```
storage   ETH      -> UniV2UniV3         : IDENTICAL
storage   Optimism -> BalancerSlipstream : IDENTICAL
selectors ETH      -> UniV2UniV3         : IDENTICAL
selectors Optimism -> BalancerSlipstream : IDENTICAL
POL unit suites                          : 18/18
```

So the behaviour-preservation claim stands on the clean comparison, not just the earlier cross-version one: no slot moved, no selector changed, and the extracted `SourceBase._fairMinAmountsOut` is line-for-line the old inline body with both call sites passing the same `tokens[0] == olas` predicate.

The two non-blocking items from my previous review carry over, neither worth holding the PR for:

- `test/LiquidityManagerETH.t.sol` keeps its chain name while the contract it exercises is now `LiquidityManagerUniV2UniV3` — the one spot the capability rename doesn't reach, and it's the file someone will go looking for.
- `script_05` verifies the proxy points where globals said, not that globals held the right *variant*. With three leaves behind one `liquidityManagerAddress` key and `changeImplementation` accepting any address by convention, a wrong-variant deploy is accepted on-chain and surfaces at runtime. A `cast call` on a variant-distinguishing immutable after the swap would catch it.

Worth flagging for whoever sequences the deployment: this PR renames the deploy scripts and the ABI artifacts, and #320's runbook plus the external-audit scope doc in #317 both reference the contracts. Once this lands, those two should be checked for the old `LiquidityManagerETH` / `LiquidityManagerOptimism` names so the rename doesn't strand them — the same way the `dispenserAddress` rename stranded its consumers on the Dispenser stack.

kupermind:
Thanks for the storage/selector verification against clean `main` — that's the comparison that matters, and I'm glad the extraction held slot-for-slot.

Both non-blocking items done (`005f6a0`), and the branch is resynced off `main` (your `ee99e10` + the workflow tweak), so the three-dot diff is purely the mixin work again.

**Item 2 — the rename now lands everywhere.** `test/LiquidityManagerETH.t.sol` → `LiquidityManagerUniV2UniV3.t.sol` (the `BaseSetup` harness + the `LiquidityManagerUniV2UniV3Test` suite), the three importing fork suites, and the `CLAUDE.md` fork-test command + architecture/overflow references. The concern-specific fork tests (`ExitFailOpenForkETH`, etc.) keep their `…ForkETH` names — those are scoped to a chain's real pools, which is the one place a chain name is the right identifier. Renamed ETH suite 24/25 (the 1 is the pre-existing flake) + the three dependents pass.

**Item 3 — variant guard in `script_05`.** After the swap it reads a source-distinguishing immutable via the proxy — `routerV2` for a UniV2 source, `balancerVault` for a Balancer source — and asserts it matches the chain's globals, so a wrong `deploy_02_*` variant is caught on-chain instead of at runtime. As you noted, this covers the **source** DEX; it doesn't distinguish Slipstream vs UniV3 target (both Balancer-source) — called out in the comment.

The only thing left is the formal approval — dismissing the changes-request cleared the block but left no approval on record, so it still shows `REVIEW_REQUIRED`.

DavidMinarsch:
Re-checked at `005f6a0`. Both non-blocking items from my approval are addressed, and one **new** issue has appeared that is partly my fault.

**Test harness rename — done.** `LiquidityManagerETH.t.sol` → `LiquidityManagerUniV2UniV3.t.sol`, with all three importers updated (`ExitFailOpenForkETH`, `ExitSandwichFork`, `IncreaseDeadBandForkETH`). The capability rename now reaches every file.

**Variant guard — good, and honestly scoped.** Reading a source-distinguishing immutable through the proxy and asserting it against globals is the right shape, and the inline note that it "covers the source DEX only; it does not distinguish a Slipstream vs UniV3 target" is exactly the kind of limitation worth writing down rather than papering over.

For what it's worth, I went looking for a way to close that target-side gap and do not think there is a clean one. Both Balancer leaves expose an identical getter set (`balancerVault`, `factoryV3`, `positionManagerV3`, `neighborhoodScanner`), so selector presence doesn't separate them — and they take the *same* globals-sourced constructor args, so a value comparison doesn't either: running the wrong `deploy_02_*` passes the correct chain NPM address to the wrong contract type, and `positionManagerV3()` still matches. A codehash check doesn't work either, since immutables are spliced into deployed bytecode. The difference lives only in the mint path (tickSpacing vs feeTier), which you cannot probe without executing. Documenting it is the right call.

### 1. [Medium — new] #319 and #320 now conflict, and the resolution is a trap

Test-merging both into `main`:

```
319 into main : clean
then 320      : CONFLICT (content) in docs/liquidity_migration_runbook.md
```

Both PRs edit the same guard table:

- **#319** renames contracts in the **V2 exit** row (`LiquidityManagerETH.sol` / `LiquidityManagerOptimism.sol` → the new names).
- **#320** de-hardcodes the gate value in the adjacent **V3 mint** row and rewrites the paragraph below it, which asserted the `maxSlippage` ↔ `MAX_ALLOWED_DEVIATION` mirroring that #318 removed.

The edits are **complementary, not contradictory** — the correct resolution is to take both sides. The trap is that taking either side wholesale looks clean and silently loses something: resolving toward #319 restores `MAX_ALLOWED_DEVIATION (10%)` and the mirroring guidance (including the release-checklist line telling an operator to match a gate that is now 2%); resolving toward #320 reverts the contract rename in that table.

I caused half of this by pushing the doc fix onto #320 rather than a standalone PR, so I'm happy to do the resolution myself on whichever lands second — say the word and I'll push it.

### 2. [Low] The variant guard degrades open

The fallback branch prints an error and then continues:

```bash
else
  echo "${red}!!! Cannot determine expected source variant (neither routerV2Address nor balancerVaultAddress set) — skipping variant check${reset}"
fi
```

Red `!!!` reads as a failure, but execution proceeds to the upgrade. Given this guard exists precisely *because* `changeImplementation` accepts any address by convention, an unset globals key is the case where you most want to stop — you have lost the only check that distinguishes a right- from a wrong-variant implementation on a single-signer EOA path. Either `exit 1`, or drop it to a non-red warning so the output doesn't claim more than it does. Every other assertion in this script and in `deploy_23` fails closed; this is the one that doesn't.

My approval stands — neither item is a reason to hold the refactor.

PR #320, opened 2026-08-14
docs: publish public Dispenser migration runbook + close audit R3

kupermind:
## Summary

Two doc-only leftovers from the #306–#316 cycle:

1. **Publish the public Dispenser migration runbook** (`docs/dispenser_migration_runbook_public.md`) — the sanitized, public-ready procedure for redeploying `VoteWeighting` + `Dispenser` and cutting the whole cross-chain staking stack over to them (deploy order, per-chain L2 migrate, withheld re-sync, verification). Links only public source registries; no exposure material. The internal working draft stays untracked.
2. **Close audit item R3** (internal20): the `Vulnerabilities_list_tokenomics.md` item #26 snippet typed `checkPoolAndGetCenterPrice` as `internal` — it is `public` (`BuyBackBurner` calls it cross-contract). One-word fix.

## Explicitly *not* included

- **The #15 / #16 / VL#26 vulnerabilities-list removals.** These are #306 *code fixes* (fail-open guard, collectFees front-run, spot-derived slippage) that are merged to `main` but **not yet redeployed** — the live `LiquidityManager` proxy still runs the old impl, so the entries must stay until `changeImplementation` deploys the fix. They belong in the deployment PR, not here.
- **R2 / R4 / R5** — already implemented in the shipped code: the skipped `collectFees` fork test is restored, the `applyDeviationGate` flag is gone from `_decreaseLiquidity`, and `checkPoolAndGetCenterPrice` / `_getExitSqrtPrice` already share `_getPoolPriceFacts` (and `_manageCollectedAmounts` is deduped into `_manageUtilityAmounts`). **R6** landed in #318.

Doc-only; no code, no conflict with the in-flight POL PRs (#317/#318/#319).


DavidMinarsch (changes requested):
## Review

This is the document I could not find when reviewing #309, where findings **(B)** (the one-time bridging-stack cascade) and **(E)** (stale ABIs) were closed against "the migration runbook". It delivers on both: §2 and §5 spell out the cascade — new Dispenser drags in every L1 processor, which drags in every L2 target dispenser, all `immutable` — and §7 carries a non-optional regeneration list naming `abis/<ver>/Dispenser.json`, `DefaultTargetDispenserL2.json` and `docs/configuration.json`. Both closures are now backed by a reviewable artifact.

Three things it does that are worth calling out, because they are the parts that usually get lost:

- It documents `cf881af` (zero `voteWeighting` at `initialize`) and `cfceafb` (the `setPauseState` go-live guard). Per internal19's coverage boundary those two commits are covered by **no internal audit**, so this runbook is currently the only artifact describing either. Step 8's "initialized with a **zero** `voteWeighting` … breaks the otherwise-circular dependency" and step 21's "the Dispenser rejects going live while `voteWeighting == address(0)`" are both accurate against the merged code.
- Step 20 re-nominates **the retainer**, and §7 verifies `retain()` succeeds. That is exactly the step that gets skipped — a fresh Dispenser's `mapLastClaimedStakingEpochs[retainerHash]` is zero until `addNominee` seeds it, and `retain()` reverts `ZeroValue` until then.
- §9 carries the corrected `checkpointNominee` fact ("reverts `NomineeDoesNotExist` for a never-added nominee; the Dispenser guards this"), rather than the version internal19 §B.3 originally got wrong.

### 1. [Medium] Step 20's cross-reference points at a bug that was fixed

> 20. … Do **not** route any target through `removeNominee` first (permanent brick — §2).

The **advice** is right, but §2 is the wrong reason, and it is a reason the code no longer has. §2 describes the brick correctly *for the existing Dispenser*:

> It hits a **permanent brick** if any nominee is `removeNominee`-then-re-added: the `firstClaimedEpoch >= epochRemoved` guard reverts (tokenomics known-issue on `mapRemovedNomineeEpochs`).

Scoped to the old contract that is accurate. But step 20 is about the **new** stack, and #310 closed vuln-list item **#25** precisely here — `addNominee` now does `delete mapRemovedNomineeEpochs[nomineeHash]`, so the new Dispenser survives a remove-then-re-add.

What still makes the advice correct is a *different* mechanism in a *different* repo: `VoteWeighting._addNominee` reverts `NomineeRemoved` while `mapRemovedNominees[hash] > 0`, which is set on removal and never cleared — removal is terminal on the VoteWeighting side. So the operator guidance stands, but a reader following the `§2` pointer concludes the new Dispenser still carries an item-25 bug it does not have.

Suggest re-pointing step 20 at the real cause: *"removal is terminal in VoteWeighting (`_addNominee` reverts `NomineeRemoved`); the Dispenser-side `mapRemovedNomineeEpochs` brick described in §2 is fixed on the new Dispenser (#310, item #25) but the VoteWeighting-side one is not."*

### 2. [Low] R3 is half-closed

The snippet is corrected, but the sentence introducing it is not — three lines above the fixed code block:

> The following **internal helper** is implemented in the LiquidityManagerCore contract

Also, the corrected signature drops a modifier that matters for this entry:

```solidity
// doc:  function checkPoolAndGetCenterPrice(address pool) public returns (uint160 sqrtP)
// code: function checkPoolAndGetCenterPrice(address pool) public view returns (uint160 centerSqrtPriceX96)
```

`view` is not cosmetic in an entry about a fail-open price read — it tells a reader the function is `staticcall`-able and state-free, which is part of why `BuyBackBurner` can call it on the hot path. If R3 is being closed, close it fully: the prose, `view`, and the return name.

### 3. [Low] VL#26's status line is now inaccurate in the other direction

The entry still reads **"Severity: Low — acknowledged; code fix planned"**. The fix is merged on `main` (#306), it is the *deployment* that is pending. Deferring the entry's **removal** to the deployment PR is the right call and I am not asking to change that — but "planned" understates the current state. "Code fix merged (#306); pending redeploy" would be accurate without prematurely claiming closure.

### Note

The branch is `BEHIND` `main`. Worth updating before merge so the runbook lands against the tree that includes #317 and the R6 constant, both of which it references.

kupermind:
All three addressed (`acf2268`), and the branch is updated off `main`.

**1. [Medium] step 20 cause.** You're right — pointing at §2 tells a reader the new Dispenser still carries item #25, which #310 fixed. Re-pointed step 20 to the real standing cause: removal is terminal in **VoteWeighting** (`_addNominee` reverts `NomineeRemoved`; `mapRemovedNominees[hash]` set on removal, never cleared), with an explicit note that the Dispenser-side `mapRemovedNomineeEpochs` brick from §2 **is** fixed on the new build (#310, item #25) but the VoteWeighting-side one is not. §2 itself stays as-is — it's scoped to the *existing* Dispenser, and it already says a fresh one avoids it.

**2. [Low] R3 fully closed.** Fixed the prose ("public helper") and matched the snippet to the code exactly — `public view returns (uint160 centerSqrtPriceX96)`. Agreed `view` isn't cosmetic here: it's what tells a reader the fail-open read is `staticcall`-able on the permissionless `BuyBackBurner` path.

**3. [Low] VL#26 status.** Changed "acknowledged; code fix planned" → **"code fix merged (#306); pending redeploy"**. Entry stays (removal waits for the redeploy, as you agreed) — this just stops "planned" from understating it.

**Branch.** Merged `main` in; it now carries #317's candidates doc. (The R6 constant is still on #318, not yet merged — the runbook doesn't reference it, so no dependency there.)

DavidMinarsch (dismissed):
All three verified against the code — approving, with one residual that arrived by timing rather than oversight.

**1. Step 20 cause — fixed, and better than what I asked for.** It now names the standing cause (removal is terminal in VoteWeighting, `_addNominee` reverts `NomineeRemoved`) *and* states explicitly that the Dispenser-side `mapRemovedNomineeEpochs` brick from §2 is fixed on the new build via #310 item #25. I only asked for the first half; adding the second prevents the inverse misreading too — someone re-reading §2 later won't conclude item #25 is still open. Agreed that §2 itself should stay as written, since it is correctly scoped to the existing Dispenser.

**2. R3 fully closed.** Prose reads "public helper", and the snippet now matches `LiquidityManagerCore.sol:1187` exactly:

```solidity
function checkPoolAndGetCenterPrice(address pool) public view returns (uint160 centerSqrtPriceX96)
```

**3. VL#26 status.** "Low — code fix merged (#306); pending redeploy". Accurate in both directions now.

### One residual — a stale gate value, in this PR's own file

`Vulnerabilities_list_tokenomics.md` §26 still reads:

> `decreaseLiquidity` reverts (`Overflow`) while `slot0` sits > `MAX_ALLOWED_DEVIATION` **(10%)** from the lagging 30-minute TWAP

This branch is current with `main` (0 behind, 4 ahead), so it carries `MAX_ALLOWED_DEVIATION = 2e16` — **2%** — at `LiquidityManagerCore.sol:128`. The document contradicts the constant sitting in the same tree.

Pure timing: #318 merged at 08:33, two minutes after your comment, so this was correct when written. Worth fixing before merge precisely because §26 is the public governance artifact, and "docs quoting a value the code no longer has" is the exact failure mode this cycle kept surfacing in internal18, internal19 and internal20.

Cheapest durable fix is to drop the parenthetical entirely and let the constant name carry it — the rest of §26 is already version-agnostic (it references `MAX_ALLOWED_DEVIATION` by name everywhere else), and the runbook has no gate-value references at all. That way the next tightening doesn't strand it again.

Not blocking the approval — but I would land the one-line change before merging rather than after.

DavidMinarsch:
Pushed `b944f87`. The stale `10%` turned out to be wider than the one line I flagged — and one instance was live operational guidance, not a gloss.

**Four documents described the gate as 10%**, and my first sweep only caught one because the other three describe the *behaviour* without naming the constant:

- `Vulnerabilities_list_tokenomics.md` §26 — the exit-blocked residual (the one I flagged).
- `liquidity_migration_runbook.md` — the guard table row, the seed-time description ("a seed whose `slot0` is within 10% of the TWAP mints at the TWAP price; a >10% deviation reverts"), and the L2 residual bound ("a bounded ≤10%+`maxSlippage` slip").

Everywhere a document was glossing the constant with a number, I dropped the number and let `MAX_ALLOWED_DEVIATION` carry it, so the next tightening doesn't strand it again.

**Three documents still recommended coupling `maxSlippage` to the gate** — the thing #318 deliberately removed. #318 corrected the note in `LiquidityManagerCore` and the POL deploy README; these were missed:

- The runbook's *"Both use a 10% bound, mirrored … so the V3 NPM's `amount{0,1}Min` check doesn't reject a mint the deviation check already accepted"* paragraph.
- The **release-checklist item**: `- [ ] maxSlippage set to 10% (matches MAX_ALLOWED_DEVIATION) on the proxy.` This is the one worth calling out — an executable instruction telling the operator to match a gate that is now 2%, in a checklist for the deployment this whole cycle is building toward. Following it literally would have re-created the coupling. The value stays 10%; I removed the false justification and noted it is sized for the source-side V2/Balancer exit floor, which is its only live role now.
- `bbb_update_options.md`'s "coherence heuristic" paragraph, replaced with a superseded note.

That last one is worth reading, because it nearly caught the dead band a month early. Its own premise was exactly right —

> a 10% price-space tolerance does not produce exactly 10% amount-space drift, it can be more or less depending on tick width and where slot0 sits

— and then it concluded that mirroring the two was therefore sound. The observation implies the opposite: because the mapping is non-linear in tick width, a matched percentage *cannot* guarantee the amount check accepts what the price check accepted, which is precisely the in-gate dead band internal20 #306.1 later found empirically (from ~0.5% deviation on a narrow range). #306 then re-anchored `amountMin` to `slot0`, making the amount-space check vacuous by construction on entries — so the heuristic is now moot as well as unsound. I recorded both halves rather than just deleting it.

Also fixed three stale `LiquidityManagerCore.sol` line citations (`:122` → `:128`, `:374-375` → `:377-378`, and the `changeMaxSlippage` reference).

Final sweep: the only remaining `10%` in the runbook and vulnerabilities list is the `maxSlippage` checklist value itself, which is correct and intentional. My approval stands.

kupermind:
Thanks — and that sweep is more thorough than the one line I'd have fixed. Verified `b944f87`:

- **§26** drops the `(10%)` parenthetical, letting `MAX_ALLOWED_DEVIATION` carry it.
- **`liquidity_migration_runbook.md`** — the guard-table row, the seed-time "within 10%" description, and the L2 residual bound all now name the constant instead of `10%`; the "both use a 10% bound, mirrored" paragraph is replaced with the decoupled explanation + "do not re-couple them."
- **`bbb_update_options.md`** — the "coherence heuristic" paragraph becomes a proper *superseded* note (its own non-linearity premise implied the opposite conclusion — the in-gate dead band — which #306 then closed by re-anchoring `amountMin` to `slot0`), the release-checklist item keeps `10%` but loses the false "matches `MAX_ALLOWED_DEVIATION`" justification, and the stale `:122`/`:374-375` citations are corrected to `:128`/`:377-378`.

All correct, and the "let the constant name carry the value" approach is the right durable fix. Approval noted.

**One heads-up for merge sequencing:** `b944f87` edits `liquidity_migration_runbook.md` and `bbb_update_options.md`, which **#319 also modifies** (it renames the guard-table's V2-exit contract refs to the capability names). They touch adjacent lines, so whichever of #319/#320 lands second will hit a trivial doc conflict — resolved by keeping this PR's gate-value sweep **and** #319's contract renames. I'll take that resolution on the second merge; flagging so it isn't a surprise.

PR #321, opened 2026-08-14
fix(config): point configuration.json at the renamed LiquidityManager artifacts

DavidMinarsch:
Follow-up to #319, which renamed the LiquidityManager leaves by capability and deleted the old chain-named ABI artifacts. `docs/configuration.json` was the one consumer the rename did not reach.

## Two live breakages on `main`

**1. Dangling artifact paths.** All three entries point at files #319 deleted:

```
abis/0.8.30/LiquidityManagerETH.json         MISSING
abis/0.8.30/LiquidityManagerOptimism.json    MISSING
```

**2. The on-chain audit script is broken — this is the one that matters.** #319 also updated `scripts/audit_chains/audit_contracts_setup.js` to look these contracts up by their **new** names:

```js
await checkLiquidityManagerImpl(providers[0], mDg.pol, configs[0]["contracts"], "LiquidityManagerUniV2UniV3", log);
await checkLiquidityManagerImpl(providers[chainNumber], dg.pol, configs[chainNumber]["contracts"], "LiquidityManagerBalancerSlipstream", log);
```

…while `configuration.json` still supplied `LiquidityManagerETH` / `LiquidityManagerOptimism`. `findContractInstance` returns `undefined` on a name miss, so `impl.address` throws — the audit crashes at its first LiquidityManager check, on every configured chain. It fails loudly rather than passing vacuously, which is the better failure mode, but it means the chain-audit is currently unusable.

## The change

Names and artifact paths only, following #319's own capability table:

| chain | was | now |
|---|---|---|
| mainnet | `LiquidityManagerETH` | `LiquidityManagerUniV2UniV3` |
| optimism | `LiquidityManagerOptimism` | `LiquidityManagerBalancerSlipstream` |
| base | `LiquidityManagerOptimism` | `LiquidityManagerBalancerSlipstream` |

**Addresses are untouched.** The implementations deployed behind those proxies are unchanged and still predate the POL redeploy — this corrects how the config *names* them, not what it points at on-chain.

## Verification

- `configuration.json` parses.
- It references **zero** missing artifacts (was 3).
- Every `LiquidityManager*` name it declares is one `audit_contracts_setup.js` actually looks up — checked both directions.

## Note

This is the third instance this cycle of a rename reaching the contracts, scripts and ABIs but not a consumer that keys off names: the `dispenserAddress` → `dispenserProxyAddress` split (caught in review on #309, swept in #314), the `internal18` audit-number collision (#306), and now this. Worth a grep for the old identifier across `docs/`, `scripts/` and `abis/` as a standing step whenever a contract is renamed — all three were mechanical to find once someone looked.

### DIFF
--- a/docs/configuration.json
+++ b/docs/configuration.json
@@ -103,8 +103,8 @@
         "address": "0x8C31cd6d5d90BB536e045A59C71b85154De1C190"
       },
       {
-        "name": "LiquidityManagerETH",
-        "artifact": "abis/0.8.30/LiquidityManagerETH.json",
+        "name": "LiquidityManagerUniV2UniV3",
+        "artifact": "abis/0.8.30/LiquidityManagerUniV2UniV3.json",
         "address": "0x0171D7178fbf3B48E75675DEE2EA3EfecEf775C1"
       },
       {
@@ -242,8 +242,8 @@
         "address": "0xDC175E77d11246c79B23D7088750eb59160DD6b7"
       },
       {
-        "name": "LiquidityManagerOptimism",
-        "artifact": "abis/0.8.30/LiquidityManagerOptimism.json",
+        "name": "LiquidityManagerBalancerSlipstream",
+        "artifact": "abis/0.8.30/LiquidityManagerBalancerSlipstream.json",
         "address": "0xCf9B4710fe450Dca52374f28e4917FDCd44F9487"
       },
       {
@@ -288,8 +288,8 @@
         "address": "0x96a475Cf364999dE5317d13f66E8e3c072bA1c0a"
       },
       {
-        "name": "LiquidityManagerOptimism",
-        "artifact": "abis/0.8.30/LiquidityManagerOptimism.json",
+        "name": "LiquidityManagerBalancerSlipstream",
+        "artifact": "abis/0.8.30/LiquidityManagerBalancerSlipstream.json",
         "address": "0x1D64516E7654F23dB7cE2ddaD00FFb3E765cf5ce"
       },
       {


kupermind (approved):
Correct and necessary — #319 renamed/deleted the old ABI artifacts, so `configuration.json`'s `abis/0.8.30/LiquidityManagerETH.json` / `LiquidityManagerOptimism.json` paths were dangling and would have broken `audit_contracts_setup.js` on load.

Verified against current `main`:

- **Artifacts resolve:** `abis/0.8.30/LiquidityManagerUniV2UniV3.json` and `LiquidityManagerBalancerSlipstream.json` both exist.
- **Complete:** all three deployed LM entries are updated (ETH→UniV2UniV3, Optimism + Base→BalancerSlipstream); `configuration.json` was the *only* remaining stranded consumer — `audit_contracts_setup.js` already carries the new names, and `external_audit_candidates.md`'s old-name mentions are intentional provenance ("extracted from / replaces"). No `BalancerUniV3` entry is needed since it isn't deployed yet.
- **Addresses correctly unchanged** — this is a source rename (ABI/selectors identical, verified on #319), not a redeploy.

One non-blocking note for whoever runs the audit script before the migration: the on-chain impls at these addresses were compiled under the old contract names, so their bytecode-metadata differs from the renamed artifacts. That's inherent to renaming-before-redeploy and reconciles when the fixed impls are deployed per the migration runbook; the name/artifact remap here is the right interim state for tooling.

Approving.

DavidMinarsch:
Thanks for the approval. Your closing note prompted me to check the bytecode question on-chain rather than reason about it, and the result refines it in two ways worth having on record before the migration.

Reading the mainnet implementation at `0x0171D7178fbf3B48E75675DEE2EA3EfecEf775C1` against the artifact `configuration.json` now points at:

```
on-chain deployed size : 21,044 B
artifact deployed size : 20,963 B   (Δ 81 bytes)
```

**1. It's a length mismatch, not metadata drift.** `checkBytecode`'s Tier 2 metadata-trailer warning only applies when lengths match; these differ by 81 bytes, so it lands in **Tier 1**. That path prints `bytecode length mismatch: artifact=… onchain=…` and `return`s — it does not throw. So the audit runs to completion and simply skips the bytecode assurance for these three contracts.

**2. The rename is not the main cause.** The deployed code does not contain the `NotEnoughHistory` selector at all, so that implementation **predates the whole #306 fail-closed remediation** — the fail-open price guard is still the live code on mainnet. The rename contributes metadata differences on top, but the 81-byte delta and the missing error are the substantive part. This is direct on-chain confirmation of what VL#26 records as *"code fix merged (#306); pending redeploy"*, and it reconciles exactly when `script_05` rolls the fixed implementation onto the proxy — so your conclusion holds, just for a bigger reason than metadata.

### One thing this surfaced, pre-existing and not from this PR

`checkBytecode`'s severity ordering is inverted: **Tier 1 (lengths differ — the strongest signal that deployed code is not the code in the repo) prints and `return`s, while Tier 2 (same length, metadata drift — the weakest signal) also only warns.** Nothing in the function fails the audit. So the one check that would catch "the wrong implementation is deployed" is the one that cannot fail it.

That is arguably right *today*, precisely because we are knowingly in a pre-redeploy state where a mismatch is expected — a hard failure would make the audit permanently red for a known reason. But it means the check provides no assurance at exactly the moment it matters most: after the migration, when a length mismatch would indicate a genuinely wrong or stale implementation behind a live proxy.

Worth considering as a post-migration follow-up: make Tier 1 a hard failure once the fixed implementations are deployed, or gate it on an expected-mismatch allowlist that the migration empties. Happy to open that as its own PR if useful — it belongs with the deployment work rather than here.

PR #323, opened 2026-08-14
feat(pol): LiquidityManagerUniV2UniV3Bridge (UniV2 → UniV3, L2 bridge burn — Celo)

kupermind:
## Summary

Adds the one (source, target, burn) combination the mixin decomposition left uncovered: **Uniswap V2 source → Uniswap V3 target → L2 bridge burn**.

It's needed for **Celo**, whose POL sits on **Ubeswap** (a Uniswap V2 fork) and which — being an L2 — must **bridge** OLAS to L1 for burning rather than burn locally. The existing `LiquidityManagerUniV2UniV3` leaf hard-codes L1 `OLAS.burn`, so it can't serve Celo.

```
LiquidityManagerUniV2UniV3Bridge = SourceUniV2 + TargetUniV3 + BurnViaBridge
```

Pure composition of the already-audited mixins — the leaf declares **no logic of its own** (constructor only; 21,102-byte runtime, ~3.4 KB EIP-170 margin).

## What's included
- **`contracts/pol/LiquidityManagerUniV2UniV3Bridge.sol`** — the leaf.
- **`scripts/deployment/pol/deploy_02_liquidity_manager_univ2univ3_bridge.sh`** — UniV2 source args (`routerV2` + `uniswapPriceOracle`) + L2 treasury + `bridge2Burner`.
- **`test/LiquidityManagerUniV2UniV3BridgeForkETH.t.sol`** — ETH-fork proof of the composition end-to-end: withdraw the real OLAS/WETH Uniswap V2 POL, mint a Uniswap V3 position via `convertToV3`, and route a 10% OLAS slice to the bridge burner (asserted) instead of `OLAS.burn()`. **Passes.** (ETH fork is a convenient stand-in with a real UniV2 OLAS pool; the composition is chain-agnostic — the eventual target is Celo/Ubeswap.)
- ABI, the BBB chain/DEX matrix (Celo now has a variant), the external-audit candidates (13th contract), and CLAUDE.md.

## Behavior risk
Both halves are already fork-proven — UniV2→UniV3 on ETH, `BurnViaBridge` on Base (via Slipstream) — so the only new surface is the three-mixin composition (constructor wiring + diamond linearization), which `forge build` + this fork test confirm.


DavidMinarsch (changes requested):
## Review

The composition claim holds, and I checked it the same way as #319 rather than by reading:

- **Constructor-only.** The leaf declares no functions of its own — pure composition, as advertised.
- **Storage layout identical** to all three existing leaves (`owner`@0.0, `maxSlippage`@0.20, `_locked`@0.22, `mapPoolAddressPositionIds`@1), so nothing about proxy upgradeability changes.
- **External surface is exactly what composition should produce**: `LiquidityManagerUniV2UniV3`'s selector set **plus `bridge2Burner()`** and nothing else — and that getter is the same one `LiquidityManagerBalancerSlipstream` exposes. The L1-burn leaf loses nothing, consistent with its direct burn being inlined rather than a public entry point.
- POL unit suites green; runtime 21,102 B, ~3.4 KB EIP-170 margin — matches the body.

This is the case #319's decomposition was justified on, and it delivers: a fourth combination for the cost of a constructor.

### 1. [Medium] This PR invalidates the `script_05` variant guard — and unlike the target-side gap, this one is closable

The guard added in #319 checks a **source-distinguishing** immutable, with a documented limitation that it "covers the source DEX only". That was acceptable because burn mode *correlated* with source: UniV2 → L1 burn, Balancer → bridge. Checking `routerV2()` implicitly pinned the burn mode too.

This leaf breaks the correlation. There are now two UniV2-source leaves differing **only** in burn mode, and the guard cannot tell them apart — it reads `routerV2()`, prints `source variant -> UniV2`, and passes for either. Concretely: deploying `LiquidityManagerUniV2UniV3` on Celo clears the guard, and OLAS burns then attempt a local `OLAS.burn` on an L2 where the bridge path is required.

When I looked at the Slipstream-vs-UniV3 *target* gap on #319 I concluded it was genuinely unclosable — identical getter sets, same globals-sourced constructor args, so neither selector presence nor value comparison separates them. **This one is different and cheap to close:**

- the Bridge leaf exposes `bridge2Burner()`; the L1-burn leaf does not;
- `globals_celo_mainnet.json` (added here) already carries `bridge2BurnerAddress`.

So: where `bridge2BurnerAddress` is set, assert `bridge2Burner()` equals it; where it is absent (ETH mainnet, L1 burn), assert the call reverts. That pins burn mode as well as source, and closes the one axis that this PR makes ambiguous. Worth doing here rather than as a follow-up, since this PR is what creates the ambiguity.

While that file is open — the guard's fallback branch still prints a red `!!!` and then continues to the upgrade (raised on #319, not yet changed). On a single-signer EOA path where this is the only check separating a right from a wrong implementation, an unset globals key is the case where it should stop.

### 2. [Low] The POL README doesn't know about the new variant

`scripts/deployment/pol/README.md` enumerates the choices —

> `univ2univ3` (Uniswap V2 → V3, e.g. ETH), `balancer_slipstream` (…e.g. Optimism/Base), or `balancer_univ3` (Balancer → Uniswap V3, e.g. Polygon/Arbitrum)

— and neither the new variant nor `deploy_02_liquidity_manager_univ2univ3_bridge.sh` appears anywhere in it. An operator using the README to pick the script for Celo will not find it, and the three variants listed are all wrong for that chain.

This is the fourth time this cycle an addition or rename has reached the contracts, scripts and ABIs but not a consumer keyed by name (`dispenserAddress` → `dispenserProxyAddress`, the `internal18` collision, `configuration.json` in #321, now this). Each was mechanical to spot with one grep for the identifier across `docs/` and `scripts/`.

### 3. [Low] The fork test proves the composition, not the target chain

Using ETH as a stand-in is reasonable and clearly disclosed, and it does prove the three-mixin wiring end-to-end. But Celo's POL sits on **Ubeswap**, a Uniswap V2 *fork*, and the eventual deployment is the one thing the test does not exercise. The repo already forks Celo elsewhere (`LPSwapCeloForkTest`), so a Celo-fork run against the real Ubeswap router before deploying there looks cheap — worth it given the source-side `removeLiquidity` path is where the fair-min math lands.

Nothing here questions the composition itself, which checks out.

DavidMinarsch:
Expanding on the last point in my review, since it is now a pattern rather than a one-off and this PR is the fourth instance.

Each time, a change reached the contracts, the scripts and the ABIs — but not a consumer that keys off a **name**:

| # | change | stranded consumer | how it was caught |
|---|---|---|---|
| 1 | `dispenserAddress` → `dispenserProxyAddress` | `deploy_08_09_change_managers.js`, all eight L1 processor deploy + verify scripts, `audit_contracts_setup.js`, both proposals | review of #309, swept in #314 |
| 2 | LM audit numbering | `audits/internal18/` collided with the already-merged Dispenser audit | review of #306 |
| 3 | LM capability rename | `docs/configuration.json` — dangling artifact paths **and** names the audit script no longer looked up | #321 |
| 4 | new LM variant (this PR) | `scripts/deployment/pol/README.md` still lists three variants; the new deploy script appears nowhere | this review |

Instance 3 is the one worth remembering: it did not merely leave a stale path. #319 updated `audit_contracts_setup.js` to the new names while `configuration.json` kept the old ones, so `findContractInstance` returned `undefined` and the chain audit crashed at its first LiquidityManager check — a rename splitting a script from its config.

All four were mechanical to find with a single grep for the identifier across `docs/`, `scripts/` and `abis/`. Two options, either cheap:

- a checklist line in the deploy/rename docs — *"grep the old and new contract name across `docs/`, `scripts/`, `abis/` before opening the PR"*; or
- a CI check that every contract in `contracts/pol/` with a deploy script also appears in the POL README's variant list, and that every `artifact` path in `docs/configuration.json` resolves. The second half would have caught instance 3 outright.

Not asking for either in this PR — flagging because the rate is now four in one cycle, and the next one lands during a migration rather than during review.

For clarity on what I would like changed **here**: item 1 of my review (extend the `script_05` guard to assert `bridge2Burner()`, which pins burn mode as well as source) and item 2 (the README variant list). Item 3 (a Celo/Ubeswap fork run) is pre-deployment rather than pre-merge, and the composition itself checks out — constructor-only, layout-identical, surface exactly `UniV2UniV3` + `bridge2Burner()`.

kupermind:
Flagged the source-side `removeLiquidity` TWAP-oracle min-out as a removal candidate in `5c1c2dd` — tracked as #324 (a proportional removeLiquidity can't be sandwiched for a value loss, so the oracle + `maxSlippage` may be droppable; the issue lays out the reasoning and what to analyze first). Comment-only here; no behavior change.

kupermind:
Both requested items done (`7ab75f0`).

**1. [Medium] burn-mode axis — closed.** You're right that this PR breaks the source↔burn-mode correlation the #319 guard leaned on. `script_05` now checks burn mode as a second axis: where `bridge2BurnerAddress` is set it asserts `bridge2Burner()` on the proxy equals it; where it's absent (the L1-burn variant — ETH globals carry no such key) it asserts the getter **reverts**. So the L1-burn `UniV2UniV3` deployed on Celo — the exact failure you described — is now caught before the first burn, not at runtime. Verified: ETH global has no `bridge2BurnerAddress`, so it takes the "assert reverts" branch.

And the fallback: the unset-source-key `else` now **aborts (`exit 1`)** instead of printing red and continuing — agreed that on a single-signer EOA path with no proposal in front of it, a missing distinguishing key is the case to stop on.

**2. [Low] README variant list — added.** It now lists `univ2univ3_bridge` (Uniswap V2 → Uniswap V3, L2 bridge burn, Celo/Ubeswap) alongside the other three, and I annotated each with its burn mode since that's now a distinguishing axis.

**3. [Low] Celo/Ubeswap fork run — agreed, pre-deployment.** The ETH stand-in proves the three-mixin wiring; a Celo-fork run against the real Ubeswap router (reusing the `LPSwapCeloForkTest` fork setup) is the right pre-deploy gate for the source-side `removeLiquidity` path specifically. I'll run it before Celo deployment rather than block merge on it — happy to add it to this PR instead if you'd prefer it in-tree now.

On the **pattern** (4th instance): fair, and this one's on me. I ran the grep this round — `LiquidityManagerUniV2UniV3Bridge` / `univ2univ3_bridge` now resolve across `docs/`, `scripts/`, `abis/` and the test, so no keyed consumer is stranded here. The CI check you sketched (every `contracts/pol/` leaf with a deploy script appears in the README list; every `configuration.json` `artifact` path resolves) is the durable fix — worth its own issue given it would have caught instance 3 outright.

DavidMinarsch (approved):
Approving. Both requested items are properly done, and the burn-mode axis is closed the way I hoped.

**Burn mode (item 1).** Verified both branches: where `bridge2BurnerAddress` is set it asserts `bridge2Burner()` matches; where absent it asserts the getter is *not* exposed. That catches the exact failure I described — the L1-burn `UniV2UniV3` deployed on Celo — and it closes the one axis this PR made ambiguous, while correctly leaving the Slipstream-vs-UniV3 target note as "genuinely unclosable" rather than pretending otherwise.

**Fallback (item 1b).** The unset-source-key `else` now `exit 1`s. That was outstanding from #319; good to have it consistent with `deploy_23` and the rest of the deploy path.

**README (item 2).** All four variants listed, and annotating each with its **burn mode** is a better fix than what I asked for — burn mode is now a selection axis, so it belongs in the table an operator reads to choose a script.

**The `#324` note in `SourceBase`** is well calibrated: *"may be droppable… Read #324 before changing."* That's the right strength — the convexity result is solid, but the `olasBurnRate` interaction I posted on #324 means "safe for value, not for intent", and pointing readers at the thread rather than summarising a conclusion keeps that intact.

---

Two things about the guard itself, neither blocking the leaf, which is what this PR is for.

### 1. The guard is detective, not preventive

It runs **after** the upgrade — `cast send … changeImplementation` at `:82`, checks from `:115`. So a wrong-variant implementation is live on the proxy from the moment the tx lands until an operator reads the abort. Harmless while POL is unseeded; not after.

The implementations are plain contracts, so `routerV2()` / `balancerVault()` / `bridge2Burner()` are all readable **directly on `$liquidityManagerAddress`** before the swap. Running both axes against the impl address ahead of the `cast send` turns this into a pre-flight check at no cost, and the existing post-swap slot read-back still confirms the swap itself landed.

### 2. The negative burn-mode branch reads an RPC failure as success

```bash
if cast call $liquidityManagerProxyAddress "bridge2Burner()(address)" … >/dev/null 2>&1; then
  # fail: getter exists but globals has no bridge2BurnerAddress
fi
echo "  burn mode -> L1 direct (bridge2Burner() absent, as expected)"
```

Any non-zero exit is taken as "the getter is absent". A flaky RPC, a wrong `networkURL`, or a missing API key produces the same non-zero exit as a missing selector — so the script prints *"bridge2Burner() absent, as expected"* and proceeds, having checked nothing.

The positive branch is immune (a failed call yields an empty string that cannot match the expected address), which is why this only bites on the L1 path — the one where the check is the sole thing standing between a bridge-burn variant and a chain that cannot bridge.

Cheap fix: in the `else`, first confirm a getter that *must* exist responds — `olas()` or `owner()` — and only then treat a `bridge2Burner()` failure as absence. If the control call also fails, that is infrastructure, not a variant signal, and should abort.

Worth landing before the script is next run rather than strictly before merge.

Issue #324, opened 2026-08-14
Re-tune the V2↔V3 ratio tolerance against the live basis once the target V3 pool is seeded


kupermind:
## Summary

The LiquidityManager's source-side withdrawal (`_checkTokensAndRemoveLiquidityV2` in `LiquidityManagerSourceUniV2` / `LiquidityManagerSourceBalancer`) derives a manipulation-resistant `minAmountsOut` from a TWAP oracle (`LiquidityManagerSourceBase._fairMinAmountsOut` → `IOracle(oracleV2).getTWAP()`). This is the **sole** purpose of the per-chain `oracleV2` (`UniswapPriceOracle` / `BalancerPriceOracle`), and post-#306.1 it is also the **only live use** of the `liquidityManagerMaxSlippage` deploy parameter.

This issue proposes evaluating whether that min-out — and therefore the oracle dependency — can be dropped.

## Why we want to do it

1. **A proportional `removeLiquidity` cannot be sandwiched for a value loss.** Burning `L` LP at manipulated reserves `(RA', RB')` with `RA'·RB' = k` yields a basket worth `L/TS · RB · (m + 1/m)` at the *true* price, and `m + 1/m ≥ 2` for any manipulation `m` — i.e. **≥ the fair (balanced) value**. A skewed pool hands the remover *more* value at the true price, never less. This is the same constant-product convexity proved for the V3 exit in `audits/internal20` R6 / vulnerabilities-list #26 (`dV/dPm = (L/2√Pm)(1 − Pt/Pm)`). So the min-out is **not** protecting against value extraction.
2. **The withdrawals are owner-gated.** `convertToV3` / `changeRanges` / `increaseLiquidity` are `onlyOwner` (DAO timelock) and pre-flight-checked per the migration runbook. An attacker cannot trigger the removal; the DAO does, at a chosen time, after confirming the pool. The manipulation window is a governance-timed op, not permissionless.
3. **It removes real deployment + maintenance surface.** Dropping it obsoletes the per-chain price-oracle deployments *and* the `liquidityManagerMaxSlippage` parameter (whose only live role is this floor — its V3 `amount{0,1}Min` uses are vacuous post-#306.1). That directly reduces the "which oracles are current/outdated?" operational burden.

## What must be analyzed before doing it

[...]

PR #325, opened 2026-08-14
Prototype: V2-removed-ratio vs V3-slot0 manipulation cross-check (oracle-drop analysis, #324)

kupermind:
## What this is

The **design + measurement half of #324** — evaluating whether the source-side TWAP oracle (`oracleV2`) on the V2/Balancer `removeLiquidity` path can be dropped. No production wiring yet: this adds a **fork prototype**, a **design note**, and **NatSpec** capturing the answer, so a tolerance can be committed on evidence rather than a guess. Follow-up PR would wire the check into `convertToV3` + retire `oracleV2`/`_fairMinAmountsOut`.

## The question behind #324

The oracle's only job is a manipulation-resistant `minAmountsOut` floor on the withdrawal (`_fairMinAmountsOut`). But a proportional `removeLiquidity` **cannot be sandwiched for a value loss** — burning `L` LP at manipulated reserves yields a basket worth `L/TS·RB·(m + 1/m)` at the true price, and `m + 1/m ≥ 2` always (constant-product convexity; same result as internal20 R6 / vuln-list #26). So the floor isn't protecting value, and the oracle is pure deployment + "is this oracle still current?" maintenance surface.

## The residual concern — and why the V3-side checks don't cover it

`convertToV3` removes from V2 and mints into V3 in the **same tx**. With the oracle gone, a **V2 pool manipulated to skew `(A, B)`** is *not* caught by the V3-side defences, because those guard the V3 **pool**, not the incoming basket:

- the 2% `MAX_ALLOWED_DEVIATION` gate (`checkPoolAndGetCenterPrice`) verifies the V3 pool vs its own TWAP — it never inspects the `(A, B)` ratio;
- the mint's `amount{0,1}Min` are slot0-anchored post-#306.1 (vacuous).

A skewed removal mints what the V3 ratio accepts and **sweeps the excess to the treasury** (`_manageUtilityAmounts(tokens, MAX_BPS, false)` — leftovers are *transferred*, not burned). Net: **no value loss** (convexity + leftover retained), but a **lopsided, inefficient conversion** — POL parked idle in treasury instead of placed in V3.

## Proposed replacement — reuse the price we already trust

`convertToV3` already computes a gate-verified `sqrtP = checkPoolAndGetCenterPrice(v3Pool)` (within 2% of the V3 TWAP) and doesn't reuse it. The V2-removed amounts are reserve-proportional, so their ratio `A:B` **is** the V2 spot. Two pools tracking the same OLAS price arbitrage together, so:

```
assert | ratio(A:B) − price(V3 slot0) | / price(V3 slot0) ≤ tolerance
```

This is today's "V2 spot vs a trusted reference" check with the reference swapped from an **external oracle** to the **V3 slot0 already on hand** — dropping the per-chain oracle while **keeping a manipulation gate** and **re-tasking `maxSlippage`** as the V2↔V3 tolerance.

- **Ratio, not "small leftover":** leftover size after a concentrated mint depends on the scanner's tick range, not just price — it conflates a narrow range with manipulation. The `A:B`-vs-slot0 ratio is range-independent.
- **Trust precondition, already enforced:** the reference is only meaningful on a *warmed* pool. An empty pool's slot0 is stale — exactly what the 2% gate + runbook pre-warm reject before any mint. The two checks **compose**: the gate makes the reference trustworthy; the cross-check makes the V2 input match it.

## Fork evidence (`test/LiquidityManagerV2V3CrossCheckForkETH.t.sol`)

| Scenario | Measured | Meaning |
| --- | --- | --- |
| Deep pair WETH/USDC, V2 vs V3 (0.05% / 0.30%) | **20 / 0 bps** | honest V2↔V3 basis is tiny → tolerance floor |
| Honest OLAS removal vs V3 seeded at true price | **0 bps** | honest flow untouched |
| **Real 50-WETH sandwich on ~272-WETH OLAS/WETH V2** | **~4007 bps (~40%)** | **cross-check reverts** |
| Unseeded OLAS/WETH V3 slot0 vs live V2 | **~9361 bps** | untrustworthy reference — 2% gate rejects it first |

Honest 0–20 bps vs manipulated ~40% — any low-single-digit-percent tolerance separates them cleanly.

## The one open sizing question

The deep-pair basis (≤20 bps) is a *lower bound*; thinner OLAS pools sit above it. Before shipping the production tolerance, **measure the real OLAS V2↔V3 basis once the target V3 pool is seeded** (post pre-warm) and set `maxSlippage` above that honest basis, below the manipulation regime — the same tuning the current `maxSlippage` already needed.

## Files
- `test/LiquidityManagerV2V3CrossCheckForkETH.t.sol` — ETH-fork prototype (long-form rationale in header)
- `docs/lm_source_crosscheck_design.md` — design note
- `contracts/pol/LiquidityManagerCore.sol`, `LiquidityManagerSourceBase.sol` — NatSpec pointers (no logic change)
- `CLAUDE.md` — fork-test command

Run: `forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerV2V3CrossCheckForkETH -vv`

Refs #324.

---

### Also in this PR — production-comment hygiene (commit 2)

Deployed contracts must be self-contained: they can't resolve GitHub issues/PRs/audit findings on-chain, so those references don't belong in NatSpec/inline comments (tests and `docs/` are fine). Rewrote the LiquidityManager comments that carried such tags — `#306.1`, `VL#15`, `internal20 R6`, `#324`, "audited", and a test-file path — into self-contained descriptions of the actual mechanism. Comment-only, no bytecode change; verified the contract diff touches no executable lines.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,7 @@ forge test --mc LPSwapCeloSlippageTest -vvv
 ```bash
 forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerUniV2UniV3 -vvv  # ETH mainnet (UniV2 -> UniV3)
 forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerUniV2UniV3BridgeForkETH -vvv  # ETH fork of the UniV2 -> UniV3 L2-burn manager (Celo)
+forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerV2V3CrossCheckForkETH -vvv  # #324 prototype: V2-removed ratio vs V3-slot0 manipulation cross-check (oracle-drop analysis)
 forge test -f $FORK_ETH_NODE_URL --mc LiquidityManagerObservationCardinalityGasETH -vvv
 forge test -f $FORK_ETH_NODE_URL --mc UniswapPriceOracleETH -vvv
 forge test -f $FORK_ETH_NODE_URL --mc BuyBackBurnerUniswapETH -vvv
--- a/contracts/pol/LiquidityManagerBalancerUniV3.sol
+++ b/contracts/pol/LiquidityManagerBalancerUniV3.sol
@@ -11,7 +11,7 @@ import {LiquidityManagerBurnViaBridge} from "./LiquidityManagerBurnViaBridge.sol
 /// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
 /// @author Mariapia Moscatiello - <mariapia.moscatiello@valory.xyz>
 /// @dev Converts protocol-owned Balancer V2 liquidity into a canonical Uniswap V3 concentrated position, for L2s
-///      that run Balancer as the source DEX and Uniswap V3 as the target. Pure composition of the audited
+///      that run Balancer as the source DEX and Uniswap V3 as the target. Pure composition of the existing
 ///      Balancer-source, UniV3-target and L2-burn mixins — it declares no logic of its own.
 contract LiquidityManagerBalancerUniV3 is
     LiquidityManagerSourceBalancer,
--- a/contracts/pol/LiquidityManagerCore.sol
+++ b/contracts/pol/LiquidityManagerCore.sol
@@ -121,10 +121,10 @@ abstract contract LiquidityManagerCore is ERC721TokenReceiver {
         0xf7d1f641b01c7d29322d281367bfc337651cbfb5a9b1c387d2132d8792d212cd;
     // Max allowed slot0-vs-TWAP price deviation (2%) in 1e18 format. Pre-flight anti-manipulation gate on
     // entries/trades: rejects an operation when slot0 has been pushed more than this far from the TWAP.
-    // Post-#306.1 the entry `amount{0,1}Min` is anchored to slot0 (vacuous by construction), so this gate is
-    // the SOLE entry manipulation defence — tightened from 10% to 2% per internal20 R6 (on a shallow,
-    // ~fully-POL pool a 10% push is buyable for ~0.1% of the position). Treat any change as a security
-    // change, not a tuning knob.
+    // The entry `amount{0,1}Min` is anchored to slot0 (vacuous by construction), so this gate is the SOLE
+    // entry manipulation defence — set to 2% because on a shallow, ~fully-POL pool a wider (e.g. 10%) push is
+    // buyable for a tiny fraction (~0.1%) of the position. Treat any change as a security change, not a
+    // tuning knob.
     uint256 public constant MAX_ALLOWED_DEVIATION = 2e16;
     // Seconds ago to look back for TWAP pool values
     uint32 public constant SECONDS_AGO = 1800;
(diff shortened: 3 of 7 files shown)


DavidMinarsch (changes requested):
## Review

Verified the framing claims rather than taking them from the description:

- **"No production wiring yet" holds** — the four `contracts/pol/` files in the diff are **comment-only**; stripping comment lines from the contract diff leaves nothing. Good discipline for an analysis PR.
- The evidence is the valuable part, and it separates cleanly: 0–20 bps honest versus ~4007 bps for a real 50-WETH sandwich is three orders of magnitude, so the tolerance choice isn't delicate.
- The **unseeded-V3 row (~9361 bps)** is the strongest thing in the note. It shows the reference is only meaningful on a warmed pool *and* that the 2% gate rejects that state first — so the two checks genuinely compose rather than just coexisting. That's the argument that makes reusing `slot0` defensible instead of merely convenient.
- Preferring the input **ratio** over a "leftover must be small" test is right for the reason given: leftover size after a concentrated mint is a function of the scanner's tick range, so it would conflate a narrow range with manipulation.

### One fix — the disposal enumeration is incomplete, and this note is the record

Lines 40–41:

> …**transferred to the treasury** (`convertToV3`'s trailing `_manageUtilityAmounts(tokens, MAX_BPS, false)` — leftovers are *transferred* to treasury, not burned/lost). Net result: **no value loss** (convexity + leftover retained)

The enumeration walks removal → mint → leftovers, and misses the step in between. `convertToV3` also runs, *before* the mint:

```solidity
if (olasBurnRate > 0) {
    amounts = _manageUtilityAmounts(tokens, olasBurnRate, true);   // true -> _burn()
}
```

That is a **rate on the realized withdrawal**, and it burns — irreversibly — where leftovers are merely parked. So the "not burned/lost" clause is accurate for leftovers but not for the whole path, and `olasBurnRate` appears nowhere in the note.

**The conclusion still holds** once the cross-check exists, for a reason worth stating explicitly rather than leaving implicit: the check reverts the whole transaction, which unwinds the burn atomically, so placing it after the burn is safe. And within tolerance the burned amount inherits the skew — at 500 bps a manipulator can move it by ~5% at swap-fee cost, bounded and value-safe by convexity, but real, and more relevant if the migration passes a non-zero `olasBurnRate`.

Since this note is the artifact of record for why a security-relevant guard was removed, I'd rather the enumeration be complete than have a future reader find the burn step unconsidered. It's one paragraph: name the burn step, state that atomic revert covers it, and record the within-tolerance skew as an accepted residual.

I've blocked #326 (the behaviour change) on three stale doc lines; this is the only thing I'd change here. Everything else in the analysis stands up.

kupermind:
Enumeration completed (`05da0c6`).

The note now names the step it was skipping: when `convertToV3` runs with a non-zero `olasBurnRate`, `_manageUtilityAmounts(tokens, olasBurnRate, true)` **burns** that rate of the removed OLAS *before* the mint — irreversibly, where leftovers are merely parked. And it states the reason the conclusion survives that step explicitly rather than implicitly: once the cross-check exists it reverts the whole transaction, so the burn unwinds atomically and a post-burn gate is safe. The within-tolerance skew (at 5% a manipulator can move the burned OLAS by up to ~5% at swap-fee cost — bounded and value-safe by convexity, but real, and larger the bigger the migration's `olasBurnRate`) is recorded as an **accepted residual** so the artifact of record leaves no step unconsidered.

Agree on everything else in your review — the unseeded-V3 row being the argument that makes reusing `slot0` defensible is exactly why it's in there.

DavidMinarsch (approved):
Approving. The enumeration now names the step it was skipping, states why the conclusion survives it, and records the residual — so the artifact of record leaves nothing unconsidered:

- the burn (`_manageUtilityAmounts(tokens, olasBurnRate, true)`) is named as running *before* the mint and as irreversible where leftovers are parked;
- the reason the conclusion holds is explicit — the cross-check reverts the whole transaction, unwinding the burn atomically, so a post-burn gate is safe;
- the within-tolerance skew is logged as an **accepted residual**, scaled to the migration's `olasBurnRate`.

That was the only thing I wanted changed here. Verified the rest already: contract changes are comment-only, and the unseeded-V3 row remains the argument that makes reusing `slot0` defensible rather than merely convenient.

Issue #322, opened 2026-08-14
audit_contracts_setup: checkBytecode cannot fail the audit (make Tier 1 blocking after the redeploy)


DavidMinarsch:
## Summary

`checkBytecode` in `scripts/audit_chains/audit_contracts_setup.js` cannot fail the audit. Its two tiers are severity-inverted: the check that would catch *"the deployed implementation is not the code in this repo"* is the one that only prints a line.

This is fine **today**, and deliberately so — we are knowingly in a pre-redeploy state where a mismatch is expected. It stops being fine the moment the POL and Dispenser migrations complete, which is exactly when the check would start carrying real assurance value.

## Current behaviour

```js
// Tier 1 (failure): on-chain code length must match the artifact's deployedBytecode length.
if (onChainCode.length !== bytecode.length) {
    console.log(tag + ", bytecode length mismatch: artifact=… onchain=…");
    console.log("\n");
    return;                      // <-- prints and returns; does not throw
}

// Tier 2 (warning): same length, trailing CBOR metadata differs
if (artifactTail !== onchainTail) {
    console.log(tag + ", WARN: metadata-trailer drift …");
}
```

- **Tier 1** — lengths differ. This is the *strongest* signal that the deployed code differs from the repo. It prints and returns.
- **Tier 2** — same length, metadata trailer differs. The *weakest* signal (solc patch version, optimizer settings, source-tree state). It warns.

Neither throws, so a chain audit passes with a wrong implementation deployed. The comment labels Tier 1 "(failure)", which does not match what the code does.

## Evidence that it is currently live

Mainnet `LiquidityManagerUniV2UniV3` implementation `0x0171D7178fbf3B48E75675DEE2EA3EfecEf775C1`, against the artifact `docs/configuration.json` points at after #321:

```
on-chain deployed size : 21,044 B
artifact deployed size : 20,963 B   (Δ 81 bytes → Tier 1)
```

The deployed code contains no `NotEnoughHistory` selector, i.e. it predates the #306 fail-closed price-guard remediation entirely — the fail-open guard is still the live code. That is expected and is what `Vulnerabilities_list_tokenomics.md` #26 records as *"code fix merged (#306); pending redeploy"*. The point is that the audit reports it as a printed line among many, not as a failure.

## Why it matters after the migration

[...]

PR #326, opened 2026-08-14
Implement the V2-removed-ratio vs V3-slot0 cross-check; drop the source-side TWAP oracle (#324)

kupermind:
Stacked on #325 (base = `feat/lm-324-crosscheck-prototype`). Implements the design #325 proposed and validated.

## What changes
Drops the per-chain source-side TWAP oracle (`oracleV2` + `_fairMinAmountsOut`) and gates source-pool manipulation **inside `convertToV3`** instead: after the gate-verified V3 price is computed, it cross-checks the ratio of the V2/Balancer-removed tokens against it and reverts `RatioDeviation` when they diverge by more than `maxSlippage`.

- A proportional `removeLiquidity`/`exitPool` **cannot lose value** (constant-product convexity), so the removal passes **zero per-token floors** to the router/vault.
- The residual risk the oracle used to cover — a manipulated source pool yielding a lopsided basket that converts inefficiently — is now caught by the ratio check, using a price we **already compute** (`checkPoolAndGetCenterPrice`). No external oracle.

## Why the V3-side checks didn't already cover this
The 2% `MAX_ALLOWED_DEVIATION` gate guards the V3 *pool*, not the incoming basket; the mint's `amountMin` is slot0-anchored (vacuous). So the ratio cross-check is the sole source-side manipulation gate — and it composes with the gate: the gate makes the V3 slot0 a trustworthy reference; the cross-check makes the removed basket match it.

## Tolerance: `maxSlippage = 500` bps (5%)
No live OLAS V2↔V3 basis exists to measure (OLAS isn't seeded on a V3 pool — that's what migration creates), so it's set from first principles and **fork-demonstrated on a seeded pool**. Honest gap ceiling ≈ 3% (2% V3 gate band + ~1% V2 arb/fee band); 5% clears it with margin and rejects meaningful manipulation. Sub-5% residual is value-safe by convexity anyway.

Measured (`test/LiquidityManagerV2V3RatioCheckForkETH.t.sol`, ETH-fork-seeded OLAS/WETH V3, ~272-WETH V2):

| WETH into V2 | removed-ratio vs V3 | rejected by 500 bps? |
| --- | --- | --- |
| honest | 0 bps | — |
| 5 | 371 bps | no |
| 10 | 748 bps | **yes** |
| 20 | 1523 bps | **yes** |
| 40 | 3154 bps | **yes** |

## Surface
- **Contracts**: `LiquidityManagerCore` (`_checkRemovedRatioAgainstV3` + `RatioDeviation` + `Q96`); `SourceBase` (oracle/fair-min removed → thin shared base); `SourceUniV2`/`SourceBalancer` (drop oracle arg, zero floors); four leaves (drop `_oracleV2` ctor arg).
- **Deployment**: `deploy_02_*` scripts drop the oracle ctor arg; the 7 pol globals set `liquidityManagerMaxSlippage=500` and drop the now-unused `*PriceOracleAddress` field; README prerequisite/env-table/`maxSlippage` guidance updated; 4 LM ABIs regenerated.
- **Docs**: `lm_source_crosscheck_design.md` (status → implemented, tolerance table); `external_audit_candidates.md` (SLoCs recalculated: 2191 total).

## Tests (all green on public-RPC forks)
- New integration test: honest convert passes, real V2 sandwich reverts `RatioDeviation`, tolerance boundary measured.
- Existing suites updated + passing: UniV2UniV3 (ETH), UniV2UniV3Bridge (ETH), Balancer→Slipstream (Base), Balancer→UniV3 (Base). One pre-existing UniV2UniV3 ATH/ATL flake fails identically on `main` (live-block price sensitivity) — unrelated.

Refs #324.


DavidMinarsch (changes requested):
## Review (#324 → #325 → #326 as one chain)

The design is good, and the cross-check is a genuinely better answer than "drop the oracle": reusing the gate-verified `sqrtP` that `convertToV3` already computes turns an external per-chain dependency into an internal invariant, and the range-independence argument for using the **ratio** rather than leftover size is right — leftover magnitude conflates a narrow scanner range with manipulation.

I verified the mechanics on `feat/lm-324-crosscheck-impl` rather than from the description.

### First, correcting my own emphasis on #324

I raised that `olasBurnRate` applies to the **realized** withdrawal (`_manageUtilityAmounts(tokens, olasBurnRate, true)` → `_burn()`), so a skewed removal burns proportionally more OLAS. The reply on #324 answered a different thing — that post-mint leftovers go to treasury via `_manageUtilityAmounts(tokens, MAX_BPS, false)` and are therefore retained. That argument is correct but doesn't reach the burn, because burning is irreversible where treasury-parking is not.

**The cross-check resolves it anyway, for a reason worth recording explicitly.** The ordering is:

```
:748  _checkTokensAndRemoveLiquidityV2       ← removal
:776  if (olasBurnRate > 0) … _burn()        ← irreversible
:782  sqrtP = checkPoolAndGetCenterPrice()
:789  _checkRemovedRatioAgainstV3(...)       ← reverts here
```

The check runs *after* the burn, which looks wrong at first glance — but `convertToV3` is one transaction, so a `RatioDeviation` revert unwinds the burn atomically. Placement is fine. Worth a one-line comment saying so, because the next reader will have the same reaction and might "fix" it by hoisting the gate call.

**Residual worth documenting:** within tolerance the burn still inherits the skew. At 500 bps a manipulator can shift the burned OLAS by up to ~5% at swap-fee cost — bounded and value-safe by convexity, but it is a real property of a rate-on-realized-amounts design, and it matters more if the migration passes a non-zero `olasBurnRate`. Currently unstated in the design note.

### 1. [Low] The 5% derivation conflates two different bands

> Honest gap ceiling ≈ 3% (2% V3 gate band + ~1% V2 arb/fee band)

The check compares **V2 removed ratio vs V3 slot0**. The honest divergence between those is the cross-pool arbitrage band. The 2% `MAX_ALLOWED_DEVIATION` bounds **V3 slot0 vs the V3 TWAP** — a different quantity that doesn't simply add to it. The conclusion (5%) is well supported by the measured table; it's the derivation that's a heuristic dressed as arithmetic. Say "5% sits above the measured honest basis and below the manipulation regime, with the fork table as evidence" and it's unarguable.

Also worth stating plainly in the table's own terms: **5 WETH into a ~272-WETH pool produces 371 bps and is not rejected.** That's a cheap manipulation passing the gate. Value-safe by convexity, so it's an acceptable residual — but it belongs in the design note as an accepted residual rather than implied by an unrejected row.

### 2. [Medium] Sixth instance of the pattern — and this time it's the docs correcting the last instance

#326 removes the source-side exit floor and re-tasks `maxSlippage` as the V2↔V3 tolerance. The README table is updated properly. Two statements on `main` are not, and both become false on merge:

- `docs/liquidity_migration_runbook.md:123` — *"Its only live role is the source-side V2/Balancer exit floor, so tightening it to 'match' the gate would risk reverting legitimate exits on a shallow source pool."*
- `docs/bbb_update_options.md:167` — *"`liquidityManagerMaxSlippage`'s only live role is therefore the **source-side** V2/Balancer exit floor."*

And the runbook's release checklist at `:272` still reads *"`maxSlippage` set to **10%** on the proxy"*, while #326 sets the globals to `500` (5%). That's an executable line telling the operator the wrong value — the same class of defect as the `matches MAX_ALLOWED_DEVIATION` line I removed two rounds ago, in the same file.

These are the lines I wrote in #320 to fix the *previous* instance, which is the sharpest illustration yet that this needs the mechanical check rather than reviewer attention: the grep for `maxSlippage` across `docs/` finds all three in one pass.

### 3. Confirmed sound

- Zero floors passed to the router (`removeLiquidity(..., 0, 0, ...)`) — consistent with the convexity result, and the floors were the only thing the oracle fed.
- `maxSlippage`'s other consumers are unaffected by 1000 → 500: the V3 `amountMin` paths are slot0-anchored and vacuous post-#306.1, as is `_decreaseLiquidity`'s. So the re-tasking doesn't quietly tighten a live control elsewhere.
- The convexity result itself I re-derived on #324 and it holds, including that manipulation fees only add to `k`.

Happy to approve once the three doc lines are updated — the contract change itself looks right.

kupermind:
All three addressed (`e7a8898`), and the mechanical sweep caught more than the three lines — details below.

**Placement comment.** Added at the call site: the check runs after the `olasBurnRate` burn on purpose, because `convertToV3` is one transaction so a `RatioDeviation` revert unwinds the burn atomically — with an explicit "do not hoist it before the burn" so the next reader doesn't try.

**1 [Low] — derivation reframed.** You're right that the "2% + ~1% = ~3%" read conflated two quantities: the check measures V2-removed-ratio vs V3-slot0 (the cross-pool arb basis), while `MAX_ALLOWED_DEVIATION` bounds V3-slot0 vs V3-TWAP (a different thing that guarantees the *reference* is trustworthy, and does not add to the arb basis). The note now anchors 5% to the fork table — "above the measured honest basis, below the manipulation regime" — and calls out the 2% as orthogonal, not a term in the budget. I also recorded the **371 bps unrejected 5-WETH row** explicitly as an accepted residual (value-safe by convexity), rather than leaving it implied by an unrejected table row.

**2 [Medium] — stale doc lines, and the sweep found the rest.** Ran `grep -rniE 'maxslippage|oracle' docs/` as you suggested. It caught your three (`runbook:123`/`:272`, `bbb:167`) **and** the fuller cascade this PR makes false — the source oracle is *removed*, so the runbook's "deploy the V2 oracle" steps (§1.1/§1.2), the "keep the Balancer oracle warm or the tx reverts" section (§4), the oracle-steerability residual (§3.2), the §5 cross-reference, and two checklist items all went stale too. Fixed all of them (the exit now passes zero floors; the source-side gate is the ratio cross-check), plus the `1000`→`500` value in `bbb_update_options` and the `10%`→`5%` checklist line. So the whole `maxSlippage`/source-oracle surface across `docs/` is consistent with the contract now, not just the three lines.

**On the pattern (6th instance):** fair, and this round is the proof — one grep surfaced a cascade a line-by-line read would have missed. The CI check you sketched (every doc `maxSlippage`/oracle claim reconciled against the contract, every deploy-script arg against the constructor) is the durable fix; worth its own issue.

**3 — confirmed-sound items:** noted, thanks for re-deriving the convexity and checking the other `maxSlippage` consumers independently.

DavidMinarsch (dismissed):
Approving. All three items resolved, and the extra commit (`8b02f4e`, removing the now-empty `LiquidityManagerSourceBase`) is the one I most wanted to check independently — removing a contract from the inheritance graph of four proxy implementations is exactly how C3 linearization silently reorders slots.

**Verified with the compiler, on a clean tree** (my first attempt produced a false pass — `forge inspect` was erroring and I was comparing empty strings against each other, so I rebuilt from a fresh clone):

| leaf | storage layout, main → PR | selectors |
|---|---|---|
| `UniV2UniV3` | **IDENTICAL** | `−oracleV2()` `+Q96()` |
| `UniV2UniV3Bridge` | **IDENTICAL** | `−oracleV2()` `+Q96()` |
| `BalancerSlipstream` | **IDENTICAL** | `−oracleV2()` `+Q96()` |
| `BalancerUniV3` | **IDENTICAL** | `−oracleV2()` `+Q96()` |

All four still `owner`@0.0 / `maxSlippage`@0.20 / `_locked`@0.22 / `mapPoolAddressPositionIds`@1, so dropping `SourceBase` is layout-safe and `changeImplementation` is unaffected. The selector delta is exactly what the change implies and nothing else. POL unit suites 18/18.

**Placement comment** — present at the call site, including the "Do not hoist it before the burn" instruction. That's the part that stops a future reader from "fixing" it.

**Guard condition** — I checked that `if (v2Pool != 0)` wrapping `_checkRemovedRatioAgainstV3` mirrors the identical condition on `_checkTokensAndRemoveLiquidityV2` exactly, so the check applies precisely when there is a removal to check, and is skipped only when there is nothing removed.

**Tolerance reframing** — the note now anchors 5% to the fork table and treats the 2% gate as orthogonal (it makes the *reference* trustworthy rather than contributing a term to the basis budget). That's the correct relationship. Recording the unrejected 371 bps row as an accepted residual rather than leaving it implied is the right call.

**Doc sweep** — confirmed, and it did go beyond my three lines. `grep -rn "only live role" docs/` is now empty, the checklist reads `5% (500 bps)` with the ratio-cross-check rationale, and the surviving oracle mentions are historical by design (*"Earlier revisions warmed a per-chain source oracle…"*, *"doesn't need … anymore"*) rather than stale claims. That is the right way to retire them — the reader learns why the oracle is gone instead of finding no trace of it.

You're right that this round is the argument for the mechanical check: one grep surfaced a cascade — the §1.1/§1.2 deploy steps, the §4 warm-up section, the §3.2 steerability residual, the §5 cross-reference and two checklist items — that a line-by-line read of the diff would not have reached, because none of it is in the diff.

### One nit, non-blocking

`Q96()` is now a public getter on all four leaves. It's a constant used only by the ratio math, so `internal`/`private` would keep it off the ABI. Four wasted selectors is nothing functionally; it's just surface that a future ABI consumer may read meaning into.

kupermind:
Thanks for the independent storage-layout verification — that's exactly the failure mode I was most wary of when pulling `SourceBase` out of four proxy implementations, so having the `main → PR` layout table confirmed identical (owner@0.0 / maxSlippage@0.20 / _locked@0.22 / map@1) is the reassurance that matters.

**Nit addressed** (`dfa0c7b`): `Q96` is now `internal` — it's pure ratio-math with no external meaning, so it's off the ABI. Selector delta on all four leaves is now just `−oracleV2()` (the `+Q96()` is gone). ABIs regenerated to match.

Also, separately: I'd missed updating **`docs/external_audit_candidates.md`** for this change — the intro and the `LiquidityManagerCore` scope section described the price-guard/decomposition work but not the source-side oracle removal + cross-check. Same commit fills that in (the new bullet + `_checkRemovedRatioAgainstV3`, the errors count, Core delta/SLoC → 673, and #326 added to the tag-diff + Ref-PRs).

Two follow-ups worth their own issues, per your reviews: the **CI reconciliation check** (doc `maxSlippage`/oracle claims vs the contract; deploy-script args vs constructors) that would have caught the cascade mechanically, and — from #324 — deferring the production tolerance re-tune against the real OLAS V2↔V3 basis once a target V3 pool is actually seeded. I'll open both unless you'd rather fold them elsewhere.

DavidMinarsch (dismissed):
Re-approving — my previous approval was auto-dismissed by the three new commits (`dismiss_stale_reviews` is on). All three verified.

**`dfa0c7b` — `Q96` nit done.** Selector delta on all four leaves is now cleanly `−oracleV2()` with nothing added; the `+Q96()` is gone. Re-ran the storage check after the new commits and all four leaves are still **IDENTICAL** to `main` (`owner`@0.0 / `maxSlippage`@0.20 / `_locked`@0.22 / `map`@1), so nothing in this round disturbed the layout.

**`70b1dfa` — the deterministic tests are the most valuable commit here.** 6/6 pass, and the coverage is the right shape: exact match, tolerance edge in **both** directions, just-over in both directions, zero-tolerance, and a high-price overflow case. Until now the cross-check was only exercised by fork tests, which need an RPC and never run automatically — so this is the first coverage that can actually protect it in CI.

**`cc7e3b5` — checked the dropped globals keys rather than assuming.** `gasPriceInGwei` and `providerName` are referenced in 32 and 100 files repo-wide, which looked alarming, but none of those are under `scripts/deployment/pol/` — the pol scripts never read either. Correct cleanup.

(While there I also confirmed the remaining per-chain key asymmetry is deliberate: `routerV2Address` is absent from Balancer-source chains and `balancerVaultAddress` from UniV2-source ones, which is exactly what `script_05`'s variant guard branches on. Worth knowing that the guard *depends* on that asymmetry, so a well-meaning "fill in all the keys everywhere" cleanup would silently disarm it.)

### One thing to land before merge — the new tests don't run in CI

`.github/workflows/workflow.yaml` uses a hardcoded `--mc` allowlist. The three existing LM suites are on it:

```
… LiquidityManagerCorePriceGuard  LiquidityManagerCoreCollectDecrease  LiquidityManagerPriceGuardRegression
```

`LiquidityManagerRatioCheckUnit` is **not**. So the suite added specifically to protect the cross-check is dead in CI — it passes locally and will silently rot exactly like the fork suites do. One entry appended to that line fixes it, and it's worth doing in this PR rather than later, since running is the entire point of making those tests deterministic.

### On the two follow-up issues

Both worth opening, with one adjustment:

- **CI reconciliation check** — yes, and keep it separate from #322 (that one is `checkBytecode`'s inverted severity; this is doc/contract and script/constructor reconciliation). This is the round that earned it: one grep found a cascade across six sections that no diff review would have reached.
- **Tolerance re-tune once a target V3 pool is seeded** — I'd leave this on **#324** rather than opening a new issue. #324 is still open, it is the parent of this whole line of work, and the re-tune is the last unfinished item in its own scope. A separate issue would orphan it from the analysis that justifies it.

Note the branch is `BEHIND` `main` now that #325 has merged — worth updating before merge.

kupermind:
Thanks for the re-review — all landed.

**CI allowlist — the one pre-merge item — done (`8ba755a`).** You're right that the deterministic suite was dead in CI, which defeats the point of making it deterministic. Appended `LiquidityManagerRatioCheckUnit` to the workflow's `--mc` line next to the other `LiquidityManager*` unit suites; confirmed it's discoverable there (6/6) and the YAML still parses.

**Branch updated.** Merged `main` in now that #325 has landed — the branch is no longer behind.

**Deliberate key asymmetry — agreed, and flagged.** I hit the same thing while cleaning the globals: the `routerV2Address`/`balancerVaultAddress` presence is exactly what `script_05`'s source-variant guard branches on, so a "fill in every key everywhere" pass would silently disarm it. I called that out explicitly in the reconciliation issue so the check treats variant-specific keys as expected-absent rather than orphans.

**Follow-ups, per your adjustments:**
- **CI doc/config reconciliation → opened as #327**, kept separate from #322 (that one is `checkBytecode` inverted severity; this is static doc/config ↔ source). It covers deploy-script args vs constructor arity, leaf ↔ README variant list, orphan config keys (with the asymmetry caveat), and a stretch doc-constant reconciliation.
- **Tolerance re-tune → left on #324**, not a new issue — you're right it's the last unfinished item in that scope and orphaning it from the analysis would lose the justification.

One thing worth surfacing that came out of the deployment review since your approval: the three live LM proxies read `maxSlippage = 1000` on-chain and `initialize` can't re-run, so a `changeImplementation`-only upgrade would leave the new ratio check at 10%. Added `script_06` (`changeMaxSlippage`, with read-back verification) for the in-place upgrade path; fresh proxies still `initialize` at 5%. Not part of the contract change, but it affects the rollout.

DavidMinarsch (approved):
Re-approving (dismissed again by the new pushes). All landed, and the `script_06` catch is the substantive one — I verified it independently rather than taking it on trust.

**CI allowlist (`8ba755a`)** — `LiquidityManagerRatioCheckUnit` is on the `--mc` line and the YAML still parses. The deterministic suite now actually runs.

**`script_06` — confirmed on-chain, and it matters more than "affects the rollout" suggests.** Reading the live mainnet proxy:

```
0x83Adc54B2828E99a6DA3A11263C75E38E0A1a215 → maxSlippage() = 1000
```

So a `changeImplementation`-only upgrade would indeed leave the new ratio cross-check running at **10%**, not 5%. Putting that against this PR's own fork table shows it is not a cosmetic gap:

| WETH into V2 | divergence | rejected at 500 | rejected at 1000 |
|---|---|---|---|
| 5 | 371 bps | no | no |
| **10** | **748 bps** | **yes** | **no** |
| 20 | 1523 bps | yes | yes |
| 40 | 3154 bps | yes | yes |

A 10-WETH manipulation on a ~272-WETH pool — the regime the 5% tolerance was specifically chosen to catch — passes at the stored value. So without `script_06` the in-place upgrade ships a gate about twice as loose as the one the fork evidence justifies, silently, because nothing in `changeImplementation` touches storage. Good catch, and the right place to catch it.

The script itself checks out: it rejects `0` and `> MAX_BPS` before sending rather than paying for a reverting tx, asserts the receipt status, and reads `maxSlippage()` back with a non-zero exit on mismatch. Same fail-closed shape as `script_05` and `deploy_23`.

**Worth adding to the runbook, not just the script.** The in-place upgrade path is now two transactions that must both land — `changeImplementation` then `changeMaxSlippage` — and the window between them is one where the new check is live at the old tolerance. Sequencing them adjacently in the deployment checklist (and stating that order) is cheap insurance against someone running the first and treating the upgrade as complete. It is exactly the class of "upgrade succeeded, configuration didn't follow" gap that `script_05`'s read-back was added for.

**Follow-ups** — #327 opened and kept separate from #322, tolerance re-tune left on #324. Both as discussed. Noting the reconciliation issue explicitly treats variant-specific keys as expected-absent rather than orphans, which is the detail that would otherwise have made the check disarm `script_05`'s guard.

Branch is up to date with `main`. Nothing outstanding from me.

DavidMinarsch (dismissed):
Re-approving (dismissed again by the new pushes). All landed, and the `script_06` catch is the substantive one — I verified it independently rather than taking it on trust.

**CI allowlist (`8ba755a`)** — `LiquidityManagerRatioCheckUnit` is on the `--mc` line and the YAML still parses. The deterministic suite now actually runs.

**`script_06` — confirmed on-chain, and it matters more than "affects the rollout" suggests.** Reading the live mainnet proxy:

```
0x83Adc54B2828E99a6DA3A11263C75E38E0A1a215 → maxSlippage() = 1000
```

So a `changeImplementation`-only upgrade would indeed leave the new ratio cross-check running at **10%**, not 5%. Putting that against this PR's own fork table shows it is not a cosmetic gap:

| WETH into V2 | divergence | rejected at 500 | rejected at 1000 |
|---|---|---|---|
| 5 | 371 bps | no | no |
| **10** | **748 bps** | **yes** | **no** |
| 20 | 1523 bps | yes | yes |
| 40 | 3154 bps | yes | yes |

A 10-WETH manipulation on a ~272-WETH pool — the regime the 5% tolerance was specifically chosen to catch — passes at the stored value. So without `script_06` the in-place upgrade ships a gate about twice as loose as the one the fork evidence justifies, silently, because nothing in `changeImplementation` touches storage. Good catch, and the right place to catch it.

The script itself checks out: it rejects `0` and `> MAX_BPS` before sending rather than paying for a reverting tx, asserts the receipt status, and reads `maxSlippage()` back with a non-zero exit on mismatch. Same fail-closed shape as `script_05` and `deploy_23`.

**Worth adding to the runbook, not just the script.** The in-place upgrade path is now two transactions that must both land — `changeImplementation` then `changeMaxSlippage` — and the window between them is one where the new check is live at the old tolerance. Sequencing them adjacently in the deployment checklist (and stating that order) is cheap insurance against someone running the first and treating the upgrade as complete. It is exactly the class of "upgrade succeeded, configuration didn't follow" gap that `script_05`'s read-back was added for.

**Follow-ups** — #327 opened and kept separate from #322, tolerance re-tune left on #324. Both as discussed. Noting the reconciliation issue explicitly treats variant-specific keys as expected-absent rather than orphans, which is the detail that would otherwise have made the check disarm `script_05`'s guard.

Branch is up to date with `main`. Nothing outstanding from me.

77ph:
**Independent security review — PASS.** Did a fresh firsthand pass (not leaning on the prior round), focused on the crux: is the ratio cross-check a sound replacement for the source-side TWAP oracle, and does it preserve manipulation-resistance? It does.

### Verified firsthand
- **`_checkRemovedRatioAgainstV3` math.** `referenceRatio = (sqrtP/2^96)^2 · 1e18` and `removedRatio = removed1/removed0 · 1e18` are **both raw token1-per-token0 ratios, so the token decimals cancel on both sides** — no scaling bug hiding in the Q96/1e18 mix. Staged `mulDiv` keeps intermediates 512-bit-correct (no overflow: `sqrtP^2/2^96 ≤ 2^224`, final `≈ 2^188`). The gate `diff > referenceRatio · maxSlippage / MAX_BPS` is a correct relative-deviation test.
- **Call-site.** The `amounts[0]==0 || amounts[1]==0 → revert` guard precedes `removed0/removed1`, so no div-by-zero. The ratio is captured **pre-burn** (the true source spot), the check is gated `if (v2Pool != 0)` (only the source-removal path — the direct-supply path has no manipulable source), and running it after the `olasBurnRate` burn is atomically safe (a `RatioDeviation` revert unwinds the burn in the same tx).
- **Zero floors are safe.** UniV2 `removeLiquidity(..., 0, 0, ...)` and Balancer `EXACT_BPT_IN_FOR_TOKENS_OUT` are both **proportional** withdrawals → the basket is convexity-protected (worth ≥ fair at the true price regardless of a source-pool push), so zero per-token floors cannot lose value on the removal; the residual "lopsided basket" risk is exactly what the ratio check catches.
- **Storage layout — compile-verified identical.** `forge inspect …:LiquidityManagerUniV2UniV3 storage-layout` on `origin/main` vs this branch is byte-identical (`owner@0.0 / maxSlippage@0.20 / _locked@0.22 / mapPoolAddressPositionIds@1`), and the removed `SourceBase` declared no storage state — so pulling it out of the four proxy implementations is layout-neutral and the `changeImplementation` upgrade is proxy-safe. (Corroborates the manual table earlier.)
- **The V3 reference is trustworthy:** `checkPoolAndGetCenterPrice` fail-closes (`NotEnoughHistory`) on an unverifiable pool and rejects a slot0 more than `MAX_ALLOWED_DEVIATION` off the TWAP; with the tightening below, that band is 2%. Fork table (honest 0 bps; 10 WETH → 748 bps rejected) confirms the gate behaves.

### Two notes tying prior findings
- **This preserves — and arguably sharpens — the anti-manipulation posture I signed off previously.** My earlier read leaned on the source-side fair-reserve valuation; this replaces it with an **explicit** revalidation of the removed basket against the gate-verified V3 price before the mint commits. Convert stays POL-only (no external mint keyed to a manipulable spot), and the removed basket is now cross-checked rather than implicitly floored — a cleaner, oracle-free guard.
- **`MAX_ALLOWED_DEVIATION 10% → 2%`** directly implements the R6 note from the price-guard review (on a shallow, near-fully-POL pool a 10% push is buyable for ~0.1% of the position). Good to see it as a security parameter with the "not a tuning knob" comment.

### Non-blocking observations
- The source-manipulation guard's reference has moved from the **source** pool's TWAP (always available) to the **target** V3 pool's TWAP — so the pre-warm prerequisite in the deployment runbook (#307) is now load-bearing for the ratio check too, not just the mint gate. It fails closed (`NotEnoughHistory`) if the V3 pool isn't warm, which is the safe direction.
- The 5% tolerance means a genuine, persistent V2↔V3 basis > 5% would revert a legitimate `convertToV3` (liveness, retry after convergence) — acceptable for a governance-timed POL op; the sub-5% residual is value-safe by convexity and yields no attacker profit (leftovers sweep to treasury).

No new issues. The prior round's operational catches (placement, storage layout, `Q96` internal, the dead-CI suite) all check out on my independent pass. Nice, oracle-free simplification.

— 77ph

DavidMinarsch (dismissed):
Re-approving (dismissed again by `73dbe77`). The two-tx window is now documented — README lists `script_06` in the sequence and says to run it *immediately after* `script_05`, with the reasoning attached. That was my last ask.

**On @77ph's independent pass** — it verified three things I had not, and they're the right three: that the decimals cancel on both sides of the ratio comparison (so there's no scaling bug hiding in the Q96/1e18 mix), that the ratio is captured **pre-burn**, and that Balancer's `EXACT_BPT_IN_FOR_TOKENS_OUT` is a *proportional* withdrawal — which is what lets the convexity argument carry over to the Balancer path at all. I had only established convexity for the constant-product case, so that closes a genuine gap in my own analysis.

Three small things left, none blocking.

### 1. The tolerance reframe landed in the design note but not in the guidance

`lm_source_crosscheck_design.md:112` now has it right:

> …so it does not add to the V2↔V3 arb basis. Its role is **orthogonal**…

But the README's globals table — the text an operator actually reads when setting the value — still carries the original conflation verbatim:

> Set it above the honest V2↔V3 basis (**bounded by the 2% V3 deviation gate + the source pool's arb/fee band, ~3% total**)

That is the additive reading the note now explicitly retires. Worth making the two agree, and the README is the one that matters more for the value someone types in.

### 2. Stale line citation, in the file that just had its citations fixed

`bbb_update_options.md:172` cites `LiquidityManagerCore.sol:624` for `changeMaxSlippage`; it is at **:690** on this branch. Same class as the `:122`→`:128` / `:374-375`→`:377-378` corrections made in that same file two rounds ago — line-number citations drift every time the contract moves, which is an argument for referring to the symbol rather than the line.

### 3. The 50/50 assumption is now load-bearing in a second, unstated way

@77ph established that the Balancer exit is proportional, so the basket is convexity-protected. Agreed. But the **ratio check** needs something stronger than proportionality: it treats the removed ratio as the source pool's spot price. For a Balancer weighted pool, spot is `(B1/w1)/(B0/w0)` while the removed ratio is `B1/B0` — those coincide **only when `w0 == w1`**.

The assumption does survive, in `LiquidityManagerSourceBalancer.sol:46` ("Balancer V2 (50/50 weighted) pool"), but as a description of the pool type rather than as a precondition of the check that now depends on it. Previously it was stated where it was used, next to `k = balance0 * balance1`, and that code is gone.

The failure mode is safe — a non-50/50 POL pool would make honest removals diverge from the V3 reference and revert, i.e. it fails closed, no exploit. So this is documentation, not a security gap. But it is worth a line at `_checkRemovedRatioAgainstV3` saying the equivalence assumes equal weights, so that a future non-50/50 POL pool surfaces as a known precondition rather than as an unexplained `RatioDeviation` during a migration.

Everything else from the prior rounds re-verified clean on this head: storage layout identical across all four leaves, selector delta `−oracleV2()` only, the deterministic suite in CI, and `script_06`'s read-back.

DavidMinarsch (approved):
Re-approving. All three landed exactly as intended, and the contract change is **comment-only** (stripping comment lines from the `LiquidityManagerCore` diff leaves nothing), so storage and selectors are untouched from the state I verified. Suites green: PriceGuard 11/11, RatioCheckUnit 6/6, PriceGuardRegression 5/5.

**1. Tolerance guidance de-conflated.** The README table now says the 2% is *"**orthogonal**, not an additive term: it bounds V3 `slot0` vs the V3 TWAP (making the reference this check compares *against* trustworthy), not the V2↔V3 basis"*. That's the right relationship stated where the value actually gets set, so the note and the guidance now agree.

**2. Line citation dropped in favour of the symbol.** The durable fix rather than bumping `:624` → `:690` — that citation would have drifted again on the next contract edit.

**3. The 50/50 precondition is recorded at the check that depends on it**, with the derivation (`(B1/w1)/(B0/w0)` vs `B1/B0`), the reason it holds today, and the fail-closed consequence if it ever doesn't. That's the form I was after: a future non-50/50 POL pool now surfaces as a named precondition rather than an unexplained `RatioDeviation` mid-migration.

Nothing outstanding from me on this PR. For the record, what I verified across the rounds, in case it's useful for the audit trail:

- storage layout **identical** to `main` on all four leaves, re-checked after `SourceBase` was removed from the inheritance graph — the change I most expected to reorder slots;
- selector delta exactly `−oracleV2()` and nothing else;
- the ratio-check guard condition mirrors the removal condition, so it applies precisely when there is something removed to check;
- `maxSlippage = 1000` on the live mainnet proxy, which is what makes `script_06` load-bearing rather than cosmetic — a `changeImplementation`-only upgrade would have run the new gate at twice its intended tolerance, and this PR's own fork table shows a 10-WETH manipulation passing at that value;
- the deterministic suite is in CI's `--mc` allowlist, so it will actually run.

One editorial nit if you touch the file again, not worth a round on its own: the README entry now says "below the manipulation regime" twice in the same sentence.

Issue #327, opened 2026-08-17
CI: reconcile docs + deploy-script args against the contracts (doc/config drift guard)


kupermind:
## Problem

Doc / deploy-config claims drift from the contracts without anything catching it. The source-side oracle removal (#326) made **six** doc sections false at once — the migration-runbook "deploy the V2 oracle" steps (§1.1/§1.2), the "keep the Balancer oracle warm or the tx reverts" section (§4), the oracle-steerability residual (§3.2), a §5 cross-reference, two checklist items, plus the `maxSlippage` "only live role" / `1000` / `10%` claims in `bbb_update_options` and the runbook — **none of which appear in the code diff**, so no amount of diff review reaches them. A single `grep` did. That's the fifth-plus instance of this class this cycle; it should be mechanical, not reviewer-dependent.

## Proposed CI check (its own job, distinct from #322)

A lightweight script (runnable in the existing workflow) that fails when docs/config no longer match the contracts:

1. **Deploy-script args vs constructor.** For each `scripts/deployment/pol/deploy_02_*.sh`, assert the `cast abi-encode "constructor(...)"` signature and the number of `constructorArgs` match the actual leaf constructor arity in `contracts/pol/`. (This is what would have caught a stale oracle arg.)
2. **Leaf ↔ README variant list.** Every `contracts/pol/LiquidityManager*.sol` leaf with a `deploy_02_*` script appears in the pol `README.md` variant table, and vice-versa.
3. **Config key ↔ script read.** Every key in `scripts/deployment/pol/globals_*.json` is either read by some pol script or is a known deploy-time output; flag orphans (would have flagged the dead `oracleV2`/`providerName`/`gasPriceInGwei` keys). NB: the per-chain `routerV2Address`/`balancerVaultAddress` asymmetry is **deliberate** — `script_05`'s variant guard branches on it — so the check must treat variant-specific keys as expected-absent, not as errors.
4. **(Stretch) Doc claim reconciliation.** A small allow-list of load-bearing claims (e.g. `maxSlippage` default, `MAX_ALLOWED_DEVIATION` value) checked against the contract constants, so a value change fails the doc that still quotes the old number.

## Why separate from #322

#322 is `checkBytecode`'s inverted severity (a runtime-verification concern). This is static doc/config ↔ source reconciliation — a different surface and a different failure mode. Keeping them separate avoids conflating "deployed bytecode matches source" with "docs/scripts match source".

Context: PR #326 review threads.

PR #328, opened 2026-08-18
chore(pol+staking): redeploy-prep — ^0.8.30 pragma + ABI refresh + audit-script fixes (#322)

kupermind:
Redeploy-prep + audit hygiene after #326. **Only touches contracts that are being redeployed / externally audited** (LMs, staking DepositProcessor/TargetDispenser); everything else is off-chain (ABIs of those contracts, docs, audit tooling). No other live contract source is changed.

## Contract pragma → `^0.8.30` (redeploy set only)
- **13 staking contracts** + **`IBridgeErrors`** (imported only by staking). These are being redeployed/audited, so the pragma is standardized to `^0.8.30` (metadata-honest — the only 0.8.x compiler configured is 0.8.30, so they already compiled with it). Wormhole* are bumped for source consistency though not slated for redeploy (live deployment unchanged).
- **Dispenser / DispenserProxy** already `^0.8.30` — untouched.
- **Nothing else** — Treasury/Depository/Tokenomics/interfaces/BuyBackBurner etc. are deliberately left at their pragmas (live, not being redeployed).

## ABIs
- Refreshed the **11 concrete staking-leaf ABIs** and the **4 LM leaf ABIs** to match source (all are being redeployed). Staking ABI diffs were **array-order only** (solc version), no interface change.
- Legacy `abis/0.8.25` / `abis/0.8.28` (matching live deployments) untouched.
- Dispenser/staking ABIs *not* being redeployed weren't altered.

## Audit script (`audit_contracts_setup.js`)
- **LMs:** dropped the removed `oracleV2()` check; capability-based checks for all four leaves.
- **#322 addressed:** a Tier-1 `deployedBytecode` length mismatch now flags the run to `exit(1)` — the audit can no longer green-light a wrong implementation. Metadata-trailer drift stays a Tier-2 warning. (Manual tool, not in CI.)
- **Dispenser proxy:** `findContractInstance` resolves `DispenserProxy` via the impl abi (like `TokenomicsProxy`); added `checkDispenserProxy`, wired in **guarded** on a config entry so it activates once the proxied Dispenser is redeployed. Typo fix.

## Runbook
Verify `maxSlippage` on-chain (in-place-upgrade `1000` trap → `script_06`); post-seed re-tune (→ #324); 50/50 precondition.

---
**#327** (doc/config reconciliation CI check) is intentionally **not** in this PR — it's a separate tooling build (new script + CI job) and belongs on its own branch; happy to do it next.

---
### Correction (per review) — the ABI refresh is NOT purely metadata/order

An earlier revision of this description said the staking ABI diffs were "array-order only, no interface change." That is wrong, and it matters. Comparing ABIs as **signature sets**:

- **10/15 are interface-identical** — all four LM leaves + the L1 DepositProcessors — metadata/ordering only, as before.
- **5 `TargetDispenserL2` ABIs gain the 4-arg `Migrated`** (Arbitrum/Gnosis/Optimism/Polygon/Wormhole): `Migrated(address,address,uint256)` → `Migrated(address,address,uint256,uint256)`. The refresh is **correct** — it catches these ABIs up with the 4-arg event from #311, which had never been regenerated.

**Consumer impact (the one thing to actually look at in this PR):** the added argument changes **topic0**, so any indexer / subgraph / migration tool keyed on the old 3-arg `Migrated` silently stops matching. Consumers must handle **both** signatures during the L2 cutover, because already-deployed dispensers keep emitting the 3-arg form until they are migrated. This is already documented in `docs/dispenser_migration_runbook.md` (the ABI-consumer note); flagging it here since this is the PR that ships the updated ABIs.


DavidMinarsch (changes requested):
## Review

The pragma scoping is disciplined — redeploy set only, live contracts (Treasury/Depository/Tokenomics/BuyBackBurner) explicitly left alone, and legacy `abis/0.8.25` / `0.8.28` untouched so they keep matching what is actually deployed. That is the right instinct.

Two things need correcting, one of them in the PR description rather than the code.

### 1. [Medium] "Staking ABI diffs were array-order only … no interface change" is not accurate

I compared every changed ABI as a **set of signatures**, so ordering is ignored. Ten of fifteen are genuinely interface-identical, including all four LM leaves. Five are not:

```
ArbitrumTargetDispenserL2.json    - event Migrated(address,address,uint256)
                                  + event Migrated(address,address,uint256,uint256)
GnosisTargetDispenserL2.json      (same)
OptimismTargetDispenserL2.json    (same)
PolygonTargetDispenserL2.json     (same)
WormholeTargetDispenserL2.json    (same)
```

**The refresh is correct** — it is catching the ABIs up with the 4-arg `Migrated` from #311, which had never been regenerated. So the change should land. The problem is that the description tells a reviewer there is nothing to look at, and this is the one change in the PR that off-chain consumers must act on: the added argument changes **topic0**, so any indexer, subgraph or migration tool keyed on the 3-arg signature silently stops matching. I raised that on #311 and it is called out in the migration runbook; it should be called out here too, since this is the PR that ships the ABI.

Worth saying explicitly in the body: *five `TargetDispenserL2` ABIs gain the 4-arg `Migrated`; consumers must handle both signatures during the L2 cutover, because already-deployed dispensers keep emitting the 3-arg form.*

### 2. [Medium] The #322 fix takes option 1 without option 2, and the tool exits 1 today

The mechanism is right — Tier 1 sets `bytecodeMismatchFound` and `main()` exits non-zero, Tier 2 stays a warning. That is what #322 asked for mechanically.

But #322 offered two routes and warned specifically about this one:

> **(1) Make Tier 1 throw** once the redeploy is complete … **(2) Gate it on an expected-mismatch allowlist** … (2) is more work but avoids a window where the audit is red for known reasons and people learn to ignore it — which is the failure mode that makes checks like this stop being read.

There is no allowlist (`grep -ciE "allowlist|expectedMismatch|skipBytecode|knownStale"` → 0), and the window is live now. Against the refreshed artifact:

```
mainnet LiquidityManagerUniV2UniV3 impl 0x0171D717…
  on-chain 21,044 B   artifact 20,518 B   → Tier 1 FAIL → exit(1)
```

The deployed implementation predates #306 entirely, so this is expected — but it means anyone running the audit today gets a red run, and the same holds for the Optimism and Base LM implementations.

Since the script is manual and not in CI, nothing is *blocked* by this. So it may well be the right call — a standing red as a reminder that the redeploy is outstanding is defensible. But it should be a **stated, dated decision** in the script or the runbook ("expected to fail on the LM implementations until the POL redeploy lands"), not an emergent property that the next person discovers and works around. Otherwise it is precisely the dynamic #322 described.

If you would rather it be green-until-genuinely-wrong, the allowlist is small: contract + address entries that the migration deletes as each implementation is redeployed.

### Confirmed

- All four LM ABIs are interface-identical — the LM half of the refresh really is metadata/ordering only, as described.
- Pragma changes touch only the redeploy set plus `IBridgeErrors` (staking-only import); no live non-redeployed contract source changed.
- The audit script's `oracleV2()` check is dropped, consistent with #326 having removed it.

Nothing here blocks the contract or ABI content — item 1 is a body edit and item 2 is a decision to record.

kupermind:
Both addressed.

**1 [Medium] — you're right, and I mischaracterized it.** The refresh is *not* purely metadata/order: the five `TargetDispenserL2` ABIs gain the 4-arg `Migrated` (the #311 event that was never regenerated). The ABI change is correct; my description told a reviewer there was nothing to look at, which was wrong for exactly the one change consumers must act on. Corrected the PR body to call it out — 10/15 interface-identical (incl. all four LM leaves), 5 gaining the 4-arg `Migrated`, and the topic0/consumer impact (indexers/subgraphs/migration tools must handle both signatures during the L2 cutover, since already-deployed dispensers keep emitting the 3-arg form). That impact is already documented in `dispenser_migration_runbook.md`.

**2 [Medium] — recorded as a stated, dated decision (`421924e`).** Went with the documented standing-red rather than the allowlist, and made it explicit at the check: a dated (2026-08) comment stating the run is *expected* to exit(1) until the POL/Dispenser redeploys land, naming the pre-#306 mainnet LM impl `0x0171D717…` and the Optimism/Base LM impls + proxied Dispenser as the intended mismatches, that each turns green as it's redeployed and `configuration.json` is repointed, and that this is deliberate — not an emergent property to work around. The allowlist is noted inline as the green-until-genuinely-wrong alternative if you'd prefer it later; the entries would be the same contract+address pairs, deleted per redeploy.

DavidMinarsch (approved):
Approving. Both items resolved, and `421924e` is **comment-only** in the audit script (9 added lines, no logic), with the ABI set untouched since my review — so nothing I verified has moved.

**1 — body corrected, and more precisely than I asked.** It now states 10/15 interface-identical, names the five `TargetDispenserL2` ABIs gaining the 4-arg `Migrated`, and adds a *"Consumer impact (the one thing to actually look at in this PR)"* section covering the topic0 change and the both-signatures requirement during cutover. That last framing is the useful part: it tells a reviewer where to spend attention rather than burying it in a list.

**2 — the standing-red is now a real decision rather than a side effect.** The comment is dated (2026-08), cites #322, names the specific intended mismatches with the byte counts (`0x0171D717…`, 21,044 on-chain vs ~20,518 artifact, pre-#306), says each clears as its implementation is redeployed and `configuration.json` is repointed, states plainly that the red is intentional and not a bug to work around, and records the allowlist as the alternative if we later want green-until-genuinely-wrong.

That is the right call, and the reason it works is the last clause: the failure mode #322 warned about is people learning that red is normal *without knowing why*. A dated comment naming exactly which contracts should fail and when each stops failing converts it into a checklist that shrinks as the migration proceeds. If it is still red after the redeploys, that is now a signal rather than noise.

One thing worth doing when the migration lands: the dated comment should be deleted in the same PR that repoints `configuration.json`, not left to age. Worth a line on the runbook's release-artifacts list next to the ABI regeneration, since that is the step where someone is already editing this surface.

Nothing outstanding from me.

PR #329, opened 2026-08-18
ci: POL deployment reconciliation check (closes #327)

### LINKED ISSUE #327: CI: reconcile docs + deploy-script args against the contracts (doc/config drift guard)

kupermind:
Closes #327.

Adds a static, read-only reconciliation check (`scripts/audit_chains/check_pol_reconciliation.js`, no RPC) wired into CI after the linters. It fails when the POL deployment surface drifts from the contracts — the drift class that never appears in a code diff and that the #326 review round kept surfacing by hand (a stale oracle arg, a missing README variant, dead globals keys).

**Three checks:**
1. **Constructor arity** — for each `deploy_02_liquidity_manager_*.sh`: `constructorArgs` count == the `cast abi-encode "constructor(...)"` signature == the contract's ABI constructor arity. (This is exactly what would have caught the stale `oracleV2` arg.)
2. **Variant list** — every `deploy_02` variant slug appears in the pol README variant list, and vice-versa.
3. **Orphan config keys** — every `globals_*.json` key is either read by some pol script or a known deploy-output / variant-specific key. Per the review note, the `routerV2Address`/`balancerVaultAddress` per-chain asymmetry (which `script_05`'s guard depends on) is treated as **expected-absent**, not an orphan — so the check can't disarm that guard.

**Validation:** passes on `main`; negative-tested by dropping one `address` from a deploy script's `abi-encode` signature → the check reports the mismatch and exits `1`; restores to `0`. eslint + gitleaks clean, YAML valid.

Scoped to POL (where the cascades happened); the same pattern can extend to staking deploy scripts later.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -39,6 +39,8 @@ jobs:
         run: ./node_modules/.bin/eslint . --ext .js,.jsx,.ts,.tsx
       - name: Run solhint
         run: ./node_modules/.bin/solhint contracts/interfaces/*.sol contracts/*.sol contracts/test/*.sol
+      - name: POL deployment reconciliation (#327)
+        run: node scripts/audit_chains/check_pol_reconciliation.js
 
       # Compile the code
       - name: Compile, adjust uniswap library hash and recompile again
--- a/scripts/audit_chains/check_pol_reconciliation.js
+++ b/scripts/audit_chains/check_pol_reconciliation.js
@@ -75,6 +75,12 @@ for (const scriptFile of deploy02Scripts) {
 // ============================================================================
 // Check 3 — every globals key is read by a pol script or is a known output/config key
 // ============================================================================
+//
+// DIRECTION IS DELIBERATE — do not invert it. This flags keys that are PRESENT but UNREAD (orphans); it does
+// NOT require that every script-read key be present in every globals file. That asymmetry is load-bearing:
+// routerV2Address is absent from Balancer-source chains and balancerVaultAddress from UniV2-source ones, and
+// script_05's variant guard branches on exactly that absence. A check demanding every read key be present
+// everywhere would push someone to fill them all in and silently disarm that guard.
 {
     // Keys read anywhere in the pol scripts (jq -r '.X' / jq -rc '.X' / // empty variants).
     const scriptText = fs.readdirSync(POL_DIR)
--- a/scripts/audit_chains/check_pol_reconciliation.js
+++ b/scripts/audit_chains/check_pol_reconciliation.js
@@ -1,0 +1,117 @@
+/*global process*/
+
+// Static reconciliation of the POL deployment surface against the contracts (issue #327).
+// Catches drift that never appears in a code diff: deploy-script constructor args vs the contract constructor,
+// deploy scripts vs the README variant list, and orphan globals keys. Exits non-zero on any mismatch, so it can
+// run in CI. Read-only; no RPC, no on-chain calls.
+
+const fs = require("fs");
+const path = require("path");
+
+const POL_DIR = "scripts/deployment/pol";
+const ABIS_DIR = "abis/0.8.30";
+const README = path.join(POL_DIR, "README.md");
+
+const failures = [];
+const fail = (m) => failures.push(m);
+const read = (p) => fs.readFileSync(p, "utf8");
+
+// ---- helpers ----
+const deploy02Scripts = fs.readdirSync(POL_DIR)
+    .filter((f) => /^deploy_02_liquidity_manager_.*\.sh$/.test(f));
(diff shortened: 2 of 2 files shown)


DavidMinarsch (dismissed):
Approving. This is the durable fix for the drift class that recurred six times across the POL/Dispenser work, and I checked it does more than pass.

**It runs and it's wired in.** `node scripts/audit_chains/check_pol_reconciliation.js` → *"POL reconciliation OK — 4 deploy variants, 7 globals checked"*, exit 0, and it's a CI step next to solhint. Read-only, no RPC, so it costs nothing per run.

**It actually bites.** A guard that passes proves little, so I introduced the two historical failures and confirmed it catches both:

```
# the #323 failure — new variant not added to the README list
POL reconciliation FAILED:
  - README: deploy variant `univ2univ3_bridge` (…_bridge.sh) is not listed
exit=1

# the configuration.json failure class — an orphan key
POL reconciliation FAILED:
  - globals_celo_mainnet.json: key "bogusOrphanAddress" is not read by any pol script
    and is not a known output/variant key (orphan)
exit=1
```

and returns to exit 0 once reverted. Those are precisely the two instances that got through review before.

**The direction of Check 3 is the right one, and worth stating so nobody "improves" it.** It flags keys that are *present but unread*, not keys that a script reads and a globals file lacks. That asymmetry is deliberate and load-bearing: `routerV2Address` is absent from Balancer-source chains and `balancerVaultAddress` from UniV2-source ones, and `script_05`'s variant guard branches on exactly that absence. A check that demanded every script-read key be present everywhere would push someone to fill them all in and silently disarm the guard. It passes on the current tree with those keys legitimately absent, so the behaviour is right — a comment saying *why* would stop a future contributor tightening it the wrong way.

Closes #327 as intended.

kupermind:
Thanks — and thanks for introducing the two historical failures to confirm it bites. Added the comment you asked for on Check 3 (`f0bd9c3`): it states the present-but-unread direction is deliberate and must not be inverted, because forcing every script-read key to be present everywhere would fill in the per-chain `routerV2`/`balancerVault` keys and silently disarm `script_05`'s variant guard.

DavidMinarsch (approved):
Approving. `f0bd9c3` says exactly what was needed — that the present-but-unread direction is deliberate, must not be inverted, and *why*: forcing every script-read key to exist everywhere would fill in the per-chain `routerV2Address` / `balancerVaultAddress` and silently disarm `script_05`'s variant guard. Naming the consequence rather than just "intentional" is what will actually stop someone tightening it.

Re-verified on this head: `POL reconciliation OK — 4 deploy variants, 7 globals checked` (exit 0), and it still bites — reintroducing the #323 drift (variant missing from the README list) exits 1, clean again on revert.

Closes #327.

PR #331, opened 2026-08-19
docs(audit): fix tag references now that v1.5.0-pre-external-audit is cut

kupermind:
Doc-only fix to `docs/external_audit_candidates.md` now that **`v1.5.0-pre-external-audit` has been cut** (on `main`, at the #330 merge).

Two issues, both in the "Before → after code state (tag diffs)" section:

1. **Stale "to be cut" language.** The doc said the tag *"is to be cut … until it is pushed, the same diff is …main"*. The tag exists now, so that fallback is dropped and it states the tag **is cut**.
2. **Dead compare link — missing `v`.** Four references wrote `1.5.0-pre-external-audit` without the `v`, so `compare/v1.4.3-post-external-audit...1.5.0-pre-external-audit` resolved the second ref to a non-existent tag. Added the `v` to all four (every other tag in the doc already has it).

Referenced by **tag name, not a commit hash**, so re-tagging after this merges keeps the doc self-consistent. Also notes the tag includes the CI/doc follow-ups #329/#330.

Verified: contract list + all 12 SLoCs still match `cloc` (SUM 1749); overall summary (13 contracts / 2186 SLoC) correct. No content beyond the tag references changed.

> Note: `v1.5.0-pre-external-audit` currently points at the #330 merge, which carries the *old* doc. After this merges, the tag should be moved to the new `main` HEAD so the corrected doc is inside it.

### DIFF
--- a/docs/external_audit_candidates.md
+++ b/docs/external_audit_candidates.md
@@ -28,9 +28,19 @@ externally-audited tag against the pre-external-audit snapshot of each repo:
 
 - **Tokenomics** — `v1.4.3-post-external-audit` → `v1.5.0-pre-external-audit`:
   https://github.com/valory-xyz/autonolas-tokenomics/compare/v1.4.3-post-external-audit...v1.5.0-pre-external-audit
-  The `v1.5.0-pre-external-audit` tag is cut on `main` at the merge of PRs
-  #306/#307/#309/#310/#311/#314/#315/#318/#319/#323/#326/#328 (the contract changes listed below), plus the
-  CI/doc follow-ups #329/#330.
+  The `v1.5.0-pre-external-audit` tag is cut on `main`. The **contract changes** are the PRs listed below —
+  #306/#307/#309/#310/#311/#314/#315/#318/#319/#323/#326/#328 — plus the **comment-only** NatSpec scrub in
+  #325 (it touches `contracts/pol/*.sol` incl. `LiquidityManagerCore.sol` but changes no non-comment line:
+  it only strips internal references such as `#306.1` / `VL#15` / `internal20 R6`). The full
+  `v1.4.3-post-external-audit…v1.5.0-pre-external-audit` range spans ~33 PRs (#287–#330); the remainder are
+  deployment, config, test, docs and audit-report changes, including the CI/doc follow-ups #329/#330.
+
+  > **Tag maintenance (decision):** `v1.5.0-pre-external-audit` is **moved** (force-updated) to the `main` HEAD
+  > that carries this doc correction — the tag name is stable, its target advances — so the snapshot the
+  > auditors receive has the corrected references above, not the pre-fix `870d46f` state. This assumes the tag
+  > has not yet been externally distributed; if it has, cut a suffixed tag (e.g. `-v2`, per the repo's
+  > `v1.4.2-post-external-audit-v3` precedent) instead of re-pointing a published ref.
+
 - **Governance** — last external-audit tag → `v1.3.0-pre-external-audit`:
   https://github.com/valory-xyz/autonolas-governance/compare/<last-audited-tag>...v1.3.0-pre-external-audit
   (per-contract diff is PR https://github.com/valory-xyz/autonolas-governance/pull/215).
--- a/docs/external_audit_candidates.md
+++ b/docs/external_audit_candidates.md
@@ -6,7 +6,7 @@ This document describes the parts of the protocol code that were not externally 
 an external audit can be performed. Repositories in scope:
 
 - https://github.com/valory-xyz/autonolas-governance/tree/v1.3.0-pre-external-audit
-- https://github.com/valory-xyz/autonolas-tokenomics/tree/1.5.0-pre-external-audit
+- https://github.com/valory-xyz/autonolas-tokenomics/tree/v1.5.0-pre-external-audit
 
 The changes below are a focused follow-up to the previously audited core protocol: a security
 hardening of the governance `VoteWeighting` gauge controller, a rework of the tokenomics `Dispenser`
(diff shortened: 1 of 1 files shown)


DavidMinarsch (commented):
Verified every claim against the GitHub API rather than reading the diff alone. **The fix is correct and needed** — and the caveat in your own PR body is, I think, the most consequential thing here and deserves more than a footnote.

## Verified

- **The dead link was genuinely dead.** `v1.5.0-pre-external-audit` resolves to `870d46f4e3`; the v-less `1.5.0-pre-external-audit` returns **404**, so `compare/v1.4.3-post-external-audit...1.5.0-pre-external-audit` could not have worked. All four `v`-additions are right.
- **Both compare links in that section now resolve**: `v1.4.3-post-external-audit` → `8d64a4420a`, governance `v1.3.0-pre-external-audit` → `886d350203`.
- **The unchanged SLoC section is internally consistent**: 717 + 34 + 673 + 325 = 1749; mixins 230 + leaves 95 = 325; 437 + 1749 = 2186; 12 tokenomics + 1 governance = 13.

## The tag still contains the broken doc

`v1.5.0-pre-external-audit` points at `870d46f4e3`, which is also current `main` HEAD (the #330 merge). I read `docs/external_audit_candidates.md` **at that tag** — it still has both defects this PR fixes:

```
:30  .../compare/v1.4.3-post-external-audit...1.5.0-pre-external-audit     ← the dead link
:31  The `1.5.0-pre-external-audit` tag is to be cut at the merge of PRs   ← the stale text
```

So after this merges, `main` is correct and **the snapshot the external auditors actually receive is not**. Your body says the tag "should be moved" — but once this PR merges, that intention exists only in a merged PR description, which is not where anyone will look.

**And moving a published audit tag is itself a hazard**, which is worth weighing rather than assuming. If auditors, CI, or other docs have already pinned `v1.5.0-pre-external-audit` at `870d46f4`, re-pointing it silently changes what that name denotes — for an audit-scope snapshot, immutability is usually worth more than doc self-consistency inside the tag.

This repo already has precedent for the alternative: **`v1.4.2-post-external-audit` (`33a2428c`) and `v1.4.2-post-external-audit-v3` (`62f5c6f9`) are distinct tags at distinct commits** — re-cut with a suffix rather than moved. Three options, worth picking explicitly in the PR rather than leaving as a note:

1. **Move the tag** — self-consistent, mutates a published ref.
2. **Leave it** — tag stays immutable, doc inside it stays stale, `main` is the source of truth.
3. **Cut `v1.5.0-pre-external-audit-v2`** — the `-v3` precedent; immutable and self-consistent, at the cost of a second tag name to communicate.

Whichever you pick, it wants a line *in the doc* ("the tag is cut at X; the corrected doc lives on `main` from Y"), so the decision survives the PR being merged.

## Minor — the enumeration now reads as exhaustive, and isn't

The new sentence is *"cut on `main` at the merge of PRs #306/…/#328 (the contract changes listed below), plus the CI/doc follow-ups #329/#330"*. The compare range actually spans **33 merged PRs (#287–#330)**, not 14. The old text's trailing "(currently `main`)" at least pointed at the real definition; the rewrite reads as a complete account of the tag's contents.

Of the 19 unlisted, 18 are deployment / config / docs / test / audit-report PRs — no issue with the "contract changes" framing. The exception is **#325 ("Prototype: V2-removed-ratio vs V3-slot0 manipulation cross-check", merged 2026-08-17)**, which modifies four files under `contracts/pol/`, including `LiquidityManagerCore.sol` (+20/-20).

I checked whether that matters: it does not, but only because **the changes are comment-only** — I filtered the patches for non-comment changed lines across all four files and got zero. They scrub internal references (`#306.1`, `VL#15`, `internal20 R6`) out of NatSpec, presumably deliberately pre-audit. So no logic entered the tag from an unlisted PR.

Still worth a clause, because a reviewer diffing the two tags *will* see `contracts/pol/*.sol` touched by a PR the doc does not mention, and will have to re-derive that it was comment-only. Something like: *"…plus the CI/doc follow-ups #329/#330 and the comment-only scrub in #325; the remaining PRs in the tag range are deployment, config, test and audit-report changes."*

Nothing blocking beyond deciding the tag question.

kupermind:
Both addressed.

**The tag decision now lives in the doc, not the PR body.** You're right that "should be moved" in a merged PR description is where no one looks. Recorded the choice — **move** `v1.5.0-pre-external-audit` to the corrected `main` HEAD — as a note in the tag-diff section, with the assumption it hasn't been externally distributed stated explicitly, and the suffixed `-v2` re-cut (the `v1.4.2-…-v3` precedent you pointed to) named as the fallback if it has. So whoever performs the tag step sees the decision at the tag, and the immutability trade-off is on the record rather than assumed.

**Enumeration fixed — it no longer reads as exhaustive.** The listed PRs are now framed as the *contract changes*, plus an explicit callout of **#325 as a comment-only NatSpec scrub** (touches `contracts/pol/*.sol` incl. `LiquidityManagerCore.sol`, zero non-comment lines — it strips `#306.1` / `VL#15` / `internal20 R6` from NatSpec), so a reviewer diffing the tags won't have to re-derive why an unlisted PR touched `contracts/pol/`. And it now states the full range is ~33 PRs (#287–#330), the remainder deployment/config/test/docs/audit-report.

Thanks for verifying the SLoC arithmetic and both compare links independently — good to have the 404 on the v-less ref confirmed from the API rather than the diff.

DavidMinarsch (commented):
Re-reviewed `4164cecf`. Both items are addressed, and addressed the right way — the tag decision is now **in the doc** rather than in a PR body that stops being visible once merged, with the `-v2` re-cut named as the fallback. The enumeration fix is accurate: I re-verified that #325 changes **no non-comment line** across the four `contracts/pol/*.sol` files, that the scrubbed references are exactly `#306.1` / `VL#15` / `internal20 R6`, and that the range is 33 PRs (#287–#330).

Two things the new wording creates, neither large.

## 1. The doc now asserts a tag state that is not yet true

> *"`v1.5.0-pre-external-audit` **is moved** (force-updated) to the `main` HEAD that carries this doc correction"*

Present tense, but at this moment the tag still dereferences to `870d46f4e3` — the pre-fix state the sentence contrasts itself against. It becomes true only if someone performs the move right after merge.

That is the same class of defect this PR set out to fix: a doc asserting a tag arrangement that does not hold. If the move is done promptly it is self-consistent afterwards (the tag would then contain the doc that describes the move), so the exposure is just the window — but it is worth closing deliberately, either by wording it as pending until executed:

> *"is **to be moved** on merge of this PR; until that is done the tag remains at `870d46f`"*

…or by treating the move as part of merging and confirming it on the PR. I would not leave it in the present tense unexecuted, since the failure mode is silent.

## 2. Executing the decision makes the range line stale

The doc says the range *"spans ~33 PRs (#287–#330)"*. Correct today. But moving the tag to the post-merge `main` HEAD pulls **#331 itself** into the range, making it #287–#331 (~34). Self-referential and unavoidable given the decision — just needs the endpoint written as `#331` so it is right after the move rather than before it.

## Minor — the tag is annotated, which matters for "force-updated"

`v1.5.0-pre-external-audit` is an **annotated** tag: the ref points to tag object `748f1b24…`, which dereferences to commit `870d46f4e3` (tagged 2026-08-18T19:20:33Z). So is the `v1.4.2-post-external-audit-v3` precedent. Moving it therefore means creating a *new* tag object and force-updating the ref — the original tagger, date and message are replaced, not preserved. Worth knowing if anything downstream reads the tag message or date, and a small point in favour of the `-v2` fallback, which keeps both objects intact.

Nothing blocking. With the tense fixed and `#331` in the range, this is done.

PR #332, opened 2026-08-19
docs(runbook): concrete Dispenser/staking deploy scripts + explicit ETH-L1-only

kupermind:
Makes the Dispenser + staking-stack deploy phases in the public migration runbook **concrete and executable**, and makes the ETH-mainnet-L1-only case explicit at the command level.

## What changed (docs-only, §6 Phase 2/3/4)
The runbook described the deploy phases correctly but generically ("deploy the L1 processors", "deploy the L2 dispensers"). Now each step names the exact script:
- **Step 10 (L1 deposit processors)** — the bridge-paired scripts per chain (`staking/deploy_02…07`, `deploy_11`), **plus an explicit ETH-L1-only callout**: `EthereumDepositProcessor` is a *different* contract + script (`deploy_08_eth_deposit_processor.sh`, ctor `(olas, dispenserProxy, stakingFactory, timelock)`), with **no** bridge relayer / `l2TargetDispenser` / L2 counterpart — skipped in steps 11, 15 and Phase 3.
- **Step 11 (L2 target dispensers)** — the per-chain subfolder scripts; **no ETH entry**.
- **Step 15 (L1↔L2 link)** — `setL2TargetDispenser` via `script_02_set_target_dispenser_l2_all.sh` (or `script_01_…` per chain; `deploy_09_*.js` equivalent).
- **Step 16 (VoteWeighting wiring)** — names `script_dispenser_change_managers.sh` (`changeManagers(0, voteWeighting)`), and notes it's also how a *future* standalone VoteWeighting redeploy is repointed onto the existing Dispenser.
- **Step 19 (`setDepositProcessorChainIds`)** — `deploy_10_set_deposit_processors.js`, including the ETH processor under the mainnet chainId.

All 23 referenced script paths verified to exist.

## Note (answers the "is there a VoteWeighting-update script?" question)
Yes — **`scripts/deployment/script_dispenser_change_managers.sh`** updates VoteWeighting on the live DispenserProxy via `changeManagers(address,address)` with `treasury = 0` (no-op) + the new `voteWeighting`. Now referenced in step 16.

---
### Added: ordered deploy-command reference
Includes the two front steps that were missing — `deploy_07a_dispenser.sh` (impl) + `deploy_07b_dispenser_proxy.sh` (proxy) on mainnet — then `script_dispenser_change_managers.sh` (VoteWeighting), `deploy_08_eth_deposit_processor.sh` for eth (L1-only, one contract, run directly) plus the per-chain `multi_deploy_01` x7 for the L2 chains (arbitrum, gnosis, optimism, celo, polygon, base, mode), and `deploy_10` (register processors). DAO/migration steps stay in the prose phases.

---
### Added: forge whitelist script `staking/deploy_10_set_deposit_processors.sh`
Forge/cast equivalent of the hardhat `deploy_10_set_deposit_processors.js`, calling `setDepositProcessorChainIds(address[],uint256[])` on the DispenserProxy. Two material differences from the hardhat script:
- **It registers Mode.** The hardhat script has seven rows — arbitrum, base, celo, ethereum, gnosis, optimism, polygon — and **no Mode row**, so pre-existing tooling would leave Mode unregistered on a fresh Dispenser. This script adds it (7 L2 chains + the L1-only Ethereum processor).
- **It hard-fails on any missing/zero processor or chainId before sending**, then re-reads `mapChainIdDepositProcessors` afterwards to confirm each entry took (this also covers `cast send`'s unchecked exit status — a failed send surfaces as a mismatch). The zero-address hard-fail is a deliberate tightening: the contract *permits* a zero processor as a per-chain disable path, which this script does not expose.

Note the failure mode of forgetting a chain: the Dispenser has **no zero-processor guard**, so an unregistered chain resolves to `address(0)` and **reverts the claim** — and in the batch path the zero-address call reverts the whole batch, taking the other chains' claims with it (not a silent skip). Script header + runbook step 19 corrected accordingly. Reads `dispenserProxyAddress` (the proxy, not the implementation) — the same impl-vs-proxy distinction being documented from the governance side in `autonolas-governance#218`.


DavidMinarsch (commented):
Checked every script path and read the wrapper and the all-chains linker rather than reading the diff alone. The direction is right — naming the exact scripts is a real improvement over "deploy the L1 processors" — and the ETH-L1-only distinction is worth making explicit. **Two things in the new command block are wrong, and both would bite an operator following it literally.**

## Verified

- **All referenced script paths exist** on the branch — 23 of them by my count (the body says 22), including `deploy_11_mode_deposit_processor.sh`, `mode/deploy_11_mode_target_dispenser.sh`, both `script_01`/`script_02` linkers and `deploy_09`/`deploy_10`.
- **`script_02_set_target_dispenser_l2_all.sh` really does cover all seven L2s** — it has ARBITRUM / BASE / CELO / GNOSIS / MODE / OPTIMISM / POLYGON sections and **no ETH section**, which matches step 15's "(ETH is skipped — no L2 side.)" exactly.
- The arg convention is right: `deploy_08_eth_deposit_processor.sh mainnet` matches what the wrapper passes (`${1#*_}`).

## Must fix 1 — "do NOT use `multi_deploy_01`" for ETH is backwards

The block says:

> `# 3a. … do NOT use multi_deploy_01, which is the processor+dispenser+link wrapper for the L2 chains`

The wrapper is generic and **already handles ETH explicitly**:

```bash
network=${1%_*}
./scripts/deployment/staking/deploy_*_${network}_deposit_processor.sh ${1#*_}

# No further L2 deployment is needed for ethereum
if [ "$network" == "eth" ]; then
  exit 0
fi
```

So `multi_deploy_01_processor_dispenser_link.sh eth_mainnet` resolves the glob to `deploy_08_eth_deposit_processor.sh mainnet` and exits before any L2 step — **precisely the command §3a tells you to run manually instead.** The note warns operators off the one path whose author coded for this exact case, and the `exit 0` guard is the wrapper's own statement of the ETH-L1-only fact this PR is trying to document.

Either use the wrapper uniformly for all seven and let it branch, or keep the direct call and justify it as a preference for explicitness — but not as "the wrapper is for L2 chains", which the source contradicts.

## Must fix 2 — Mode is missing from the command block

Steps 10 and 11 both list Mode (`deploy_11`, `mode/deploy_11`); Phase 3 line 105 lists Mode among *"every chain with an L2 dispenser"*; `script_02_…_all.sh` has a MODE section. But §3b has **six** `multi_deploy_01` lines — base, optimism, polygon, arbitrum, celo, gnosis — and the PR body says "the per-chain `multi_deploy_01` **x7**".

An operator following the block top-to-bottom deploys six of seven chains, with **no error** — Mode is simply never deployed, and step 19's `deploy_10_set_deposit_processors.js` then registers a processor set missing Mode. Nothing prevents it either: the wrapper is glob-driven, and both Mode scripts exist, so `multi_deploy_01_processor_dispenser_link.sh mode_mainnet` works today.

## Minor

- **Polygon has a hidden extra step, and it argues for the wrapper.** `multi_deploy_01` runs `script_03_set_deposit_processor_l1_polygon.sh` after the link for Polygon only — the `fxRootTunnel` binding. The prose phases never mention it, so anyone deploying Polygon by the individual scripts in steps 10/11/15 misses it silently. Worth a note in step 15, and a further reason to prefer the wrapper over hand-run scripts.
- **Ordering.** §3b runs base → optimism → polygon → arbitrum → celo → gnosis, while steps 10/11 list arbitrum → gnosis → optimism → celo → polygon → base → mode. Harmless, but for a runbook meant to be followed top-to-bottom, matching the order removes a reconciliation step for the reader.
- Body says "All 22 referenced script paths verified"; I count 23 distinct paths.

Nothing structural — with Mode added and the ETH note corrected (or dropped in favour of the wrapper), this reads well.

DavidMinarsch (commented):
Re-reviewed `0a6f5d23` + `6dce1c0e`. Both must-fixes are addressed correctly, and the PR has grown a **new 152-line deploy script**, so I reviewed that against the contract rather than as docs.

## The two must-fixes

- **Mode added**, and the block is now seven chains in the same order as steps 10/11. 
- **The ETH note is right now.** It no longer says "do NOT use `multi_deploy_01`"; it states that the wrapper *would* work, globbing to this exact `deploy_08 … mainnet` call and hitting its own `exit 0` ETH guard, and frames the direct call as a preference for explicitness. That matches the wrapper's source.
- Polygon's `fxRootTunnel` step is now called out in **both** step 15 and the command block, including the "easy to miss if run by hand" framing. Good.

## The new script — verified against `Dispenser.sol`

- **ABI is exact**: `setDepositProcessorChainIds(address[] memory, uint256[] memory)` at `:808`, and `mapChainIdDepositProcessors` is a `public` mapping at `:327`, so the verification loop's `mapChainIdDepositProcessors(uint256)(address)` is the right getter.
- **The ownership note is load-bearing and correct** — the setter is `msg.sender != owner → OwnerOnly`, so "do not run this directly against a DAO-owned proxy" is exactly right.
- **The zero-chainId comment matches the contract**: `chainIds[i] == 0 → revert ZeroValue()`. Note the contract explicitly *permits* a zero **processor** ("might be zero if there is a need to stop processing a specific L2 chain Id") — so the script's hard-fail on a zero address is a deliberate tightening beyond the contract, which is the right call for this use but worth knowing if anyone ever needs the disable path.
- **"additionally registers the Mode processor" is true and material.** The hardhat `deploy_10_set_deposit_processors.js` registers seven entries — arbitrum, base, celo, ethereum, gnosis, optimism, polygon — and has **no Mode row at all**. So the pre-existing tooling would leave Mode unregistered on a fresh Dispenser. That is a real gap this script closes, and it is worth saying so in the PR body.
- It reads **`dispenserProxyAddress`**, which is the correct key — and is precisely the implementation-vs-proxy distinction that `autonolas-governance#218` is currently trying to document from the other side. Worth cross-linking the two.
- Design is sound: hard-fail on any missing/zero entry *before* sending, then re-read the mapping afterwards. The post-call verification also covers the un-checked exit status of `cast send` — a failed send surfaces as a mismatch rather than passing silently.

## One error — "silently skipped" is not what happens

Both the script header and runbook step 19 say *"a chain left unregistered … is silently skipped at distribution."* There is **no zero-processor guard anywhere in `Dispenser.sol`** — `_distributeStakingIncentives` does:

```solidity
address depositProcessor = mapChainIdDepositProcessors[chainId];
if (transferAmount > 0) { IToken(olas).transfer(depositProcessor, transferAmount); }
IDepositProcessor(depositProcessor).sendMessage{value: msg.value}(...);
```

With an unregistered chain that resolves to `address(0)`, and `sendMessage` is a void high-level call — so Solidity's `extcodesize` check fires and the call **reverts**. The distribution does not skip the chain; it fails. And in `_distributeStakingIncentivesBatch` the same read happens inside the chain loop, so **one unregistered chain reverts the entire batch**, taking the other chains' claims down with it.

That is a louder and more disruptive failure than "silently skipped", and it strengthens rather than weakens the case for this script — forgetting Mode does not quietly under-distribute, it blocks claims. Suggest: *"a chain left unregistered here resolves to a zero processor, which reverts the claim — and in the batch path reverts the whole batch."*

(Reachability depends on there being live nominees on the unregistered chain; on Mode specifically several nominees have been removed, so it may not bite immediately.)

## Minor

Exit-code inconsistency inside the one file: missing globals file and missing Alchemy key `exit 0`, while missing/zero addresses `exit 1`. The `exit 0` matches the convention in `script_02_set_target_dispenser_l2_all.sh`, but mixing both in a script that is meant to be chained means a caller using `&&` treats "globals not found" as success.

DavidMinarsch (commented):
Re-reviewed `877dc241`. Both items fixed, and the failure-mode correction is **more precise than what I reported** — worth recording why.

I attributed the revert to `IDepositProcessor(depositProcessor).sendMessage(...)` in `_distributeStakingIncentives`. The commit attributes it to `getBridgingDecimals()`, and that is the better call: it is invoked at `:687` in the batch path and `:1154` in the single-claim path, both **before** distribution is reached, and unlike the void `sendMessage` it is `external pure returns (uint256)` — so the zero-address call fails on decoding empty returndata rather than only on the `extcodesize` check. The revert therefore happens earlier in the flow than I said, and the batch consequence (one unregistered chain taking down the whole batch) holds for the same reason. Both the script header and runbook step 19 now describe this accurately.

Exit codes are consistent at `1` throughout, with a comment explaining the deliberate deviation from the `exit 0` used in the per-network scripts — correct for a terminal step that gets `&&`-chained.

Nothing further from me on this one.

PR #333, opened 2026-08-20
fix: align collected V3 fees with the caller-provided token ordering

kupermind:
## The problem

`LiquidityManagerCore.collectFees()` is permissionless, and Uniswap resolves the same pool whichever way round the pair is passed. But the two halves of the function disagree about what the ordering means:

- `_collectFees(positionId)` returns `(amount0, amount1)` against **the position's** `token0` / `token1`;
- `_manageAmounts(tokens, amounts, true)` assigns the OLAS and secondary-token roles from **the caller's** `tokens` array as supplied.

So a call whose `tokens` array is ordered differently from the position burns the *secondary-token* fee amount as OLAS and transfers the *OLAS* fee amount as the secondary token. The two quantities are simply transposed, and the difference is drawn from whatever is staged on the contract.

## The fix

Read the position's `token0` and swap the collected amounts when the caller's first token is not it:

```solidity
(,, address positionToken0,,,,,,,,,) = IPositionManagerV3(positionManagerV3).positions(positionId);
if (tokens[0] != positionToken0) {
    (amounts[0], amounts[1]) = (amounts[1], amounts[0]);
}
```

Alignment is by **token identity**, not by address ordering. An earlier draft compared `tokens[0] > tokens[1]` on the assumption that a position is always canonically sorted — that assumption held on the fork but broke `LiquidityManagerCoreCollectDecrease`, whose mock position manager defines its own `token0`/`token1`. Identity matching is correct in both worlds and makes no assumption about how the position was constructed.

`_manageUtilityAmounts()` is unaffected: it builds `amounts[i]` from `balanceOf(tokens[i])`, so its amounts and roles are already keyed to the same array. `collectFees` was the only path mixing the two conventions.

## Tests

New `test/LiquidityManagerCollectFeesTokenOrderForkETH.t.sol`, on the real OLAS/WETH V3 position:

- **positive** — `test_collectFeesCanonicalOrderRoutesFees`: the matching ordering burns the OLAS fee, forwards the secondary fee to the treasury, and leaves staged balances untouched.
- **negative** — `test_collectFeesReversedOrderRoutesIdentically`: takes a state snapshot, collects canonically, reverts, re-stages, collects reversed, and asserts the two produce identical routing. It also pins the staged balances at zero delta, which is the invariant that actually distinguishes the two behaviours, and explicitly asserts the transposition is absent.

Verified the negative test fails without the contract change:

```
[FAIL: reversed order must burn the same OLAS fee: 29901070206668 != 3198978924562019405]
```

— i.e. it burnt the secondary-token fee amount rather than the OLAS one. With the change, both pass.

Run with:

```bash
forge test --mc LiquidityManagerCollectFeesTokenOrderForkETH --fork-url <eth-rpc> -vv
```

## Checks run locally

| Check | Result |
|---|---|
| `LiquidityManagerCoreCollectDecrease` | 2 passed |
| `LiquidityManagerCorePriceGuard` | 11 passed |
| `LiquidityManagerPriceGuardRegression` | 5 passed |
| `LiquidityManagerRatioCheckUnit` | 6 passed |
| New fork suite | 2 passed |
| `solhint` (CI scope) | no new findings |
| `eslint` | 0 errors |
| `npx hardhat compile --force` | 156 files compiled |

## Deployment note

**`LiquidityManagerCore` is live.** Merging this makes the source diverge from the deployed bytecode until the contract is redeployed, so it should be scheduled with any other pending change to the same contract rather than shipped alone.

Until a redeploy, the existing operational guidance still bounds the issue: fee routing can only misdirect value that is **staged on the contract between transactions**, so not leaving staged balances there reduces the exposure to zero.

### DIFF
--- a/contracts/pol/LiquidityManagerCore.sol
+++ b/contracts/pol/LiquidityManagerCore.sol
@@ -940,6 +940,17 @@ abstract contract LiquidityManagerCore is ERC721TokenReceiver {
         // Collect fees
         amounts = _collectFees(positionId);
 
+        // Align collected amounts with the caller-provided token ordering. The position manager returns
+        // (amount0, amount1) against the position's own token0 / token1, whereas _manageAmounts() assigns
+        // the OLAS and secondary-token roles from the tokens array exactly as supplied. The pool resolves
+        // for either ordering, so without this alignment a call whose tokens array is ordered differently
+        // from the position would burn the secondary-token fee amount as OLAS and transfer the OLAS fee
+        // amount as the secondary token, drawing the difference from balances staged on the contract.
+        (,, address positionToken0,,,,,,,,,) = IPositionManagerV3(positionManagerV3).positions(positionId);
+        if (tokens[0] != positionToken0) {
+            (amounts[0], amounts[1]) = (amounts[1], amounts[0]);
+        }
+
         // Check for zero values
         if (amounts[0] == 0 && amounts[1] == 0) {
             revert ZeroValue();
--- a/docs/Vulnerabilities_list_tokenomics.md
+++ b/docs/Vulnerabilities_list_tokenomics.md
@@ -35,6 +35,7 @@
   - [29. numNewOwners is attributable via permissionless unit creation, contributing to IDF](#29-numnewowners-is-attributable-via-permissionless-unit-creation-contributing-to-idf)
   - [30. claimStakingIncentives skips a stakingFraction == 0 epoch carrying a positive staking incentive](#30-claimstakingincentives-skips-a-stakingfraction--0-epoch-carrying-a-positive-staking-incentive)
   - [31. Dispenser retains caller-supplied bridge value for zero-output staking claims](#31-dispenser-retains-caller-supplied-bridge-value-for-zero-output-staking-claims)
+  - [32. Depository accepts a zero-payout bond and strands the collateral](#32-depository-accepts-a-zero-payout-bond-and-strands-the-collateral)
   - [27. BuyBackBurner buyBack unused in default operation — swap paths retained as compatibility surface](#27-buybackburner-buyback-unused-in-default-operation--swap-paths-retained-as-compatibility-surface)
 ## Involved contracts and level of the bugs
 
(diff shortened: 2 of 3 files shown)


DavidMinarsch (commented):
Reviewed the contract change against the surrounding code and the test against what it would actually catch.

## The fix is correct — and, importantly, complete

The alignment itself is right, and the reason it is *sufficient* is worth recording because it is not obvious from the diff:

- `_getV3Pool` resolves through `factory.getPool(tokens[0], tokens[1], …)` in **both** leaf implementations (`LiquidityManagerTargetUniV3`, `LiquidityManagerTargetSlipstream`), and that call is order-insensitive. A wrong pair resolves to a different pool or zero, and `mapPoolAddressPositionIds[pool] == 0` then reverts. So the caller-controlled input space is **exactly the two orderings of the real pair** — which makes `tokens[0] != positionToken0` ⟺ `tokens[0] == positionToken1`, and the unconditional swap correct with no third case to handle.
- I checked the other `_manageAmounts` call site for the same defect. `_manageUtilityAmounts` builds `amounts[i] = IToken(tokens[i]).balanceOf(...)` — derived *from* `tokens`, so self-consistent by construction and not exposed to this. **`collectFees` was the only site where position-derived `(amount0, amount1)` met a caller-supplied `tokens` array**, so this is the whole of the bug, not one instance of a pattern.

The swap sits before the zero-check, which is order-independent (`both == 0`), so no interaction there.

## The test is strong

Two details raise it above a smoke test: it snapshots and replays so both orderings run against **identical** state, and it asserts the two fee amounts are **unequal** before comparing (`:82`, `:117`). Without that precondition the transposition is invisible and the whole test passes vacuously. The staged-balance assertions also pin the funding-source half of the impact, not just the routing half, and `:118` explicitly pins the pre-fix transposition as absent.

## Deployment status checks out

Entry #28's "fix is in source only — no implementation carrying it has been deployed" is accurate. The live implementations behind the three V3-enabled `LiquidityManagerProxy` instances — Ethereum `0x0171D7178fbf3B48E75675DEE2EA3EfecEf775C1`, Base `0x1D64516E7654F23dB7cE2ddaD00FFb3E765cf5ce`, Optimism `0xCf9B4710fe450Dca52374f28e4917FDCd44F9487` (read from the `keccak256("PROXY_LIQUIDITY_MANAGER")` slot on each proxy) — do not carry it.

Worth confirming before the redeploy is scheduled that any other pending source-level changes to this contract go out in the same implementation, rather than shipping one and leaving the rest queued.

## Minor

- **Return-value semantics changed.** `collectFees` now returns `amounts` in *caller* order rather than position order, and `PositionFeesCollected` emits `tokens, amounts` accordingly. That is the right alignment, but it is an observable interface change for anything parsing that event — worth a line in the entry.
- `positions(positionId)` decodes twelve fields to use one. `IUniswapV3Pool(pool).token0()` gives the same answer more cheaply, and `pool` is already in scope. Cosmetic.

PR #338, opened 2026-08-22
fix: bridge the Celo LP migration inputs to LPSwapCelo, not the bridge mediator

DavidMinarsch:
Steps 2b and 4b bridge the LP token and the OLAS pre-funding to `bridgeMediatorAddress`. They need to go to `LPSwapCelo`, because that contract reads both from its **own** balance:

```solidity
uint256 liquidity = IToken(LP_TOKEN).balanceOf(address(this));
...
if (IToken(OLAS).balanceOf(address(this)) == 0) {
```

With the assets sitting at the mediator, `swapLiquidity()` has nothing to work with and the migration can't start. Recovering would take a second governance proposal just to forward both assets to the swap contract.

The mediator **is** the right destination for the swap's *outputs* — which `LPSwapCelo` already handles itself: `addLiquidity()` mints to `BRIDGE_MEDIATOR`, and leftover WCELO is transferred there. It just isn't the destination for the inputs.

**Fix:** point both bridge legs at `lpSwapCeloAddress`, read from the same Celo globals file that already supplies `bridgeMediatorAddress` and written there by `scripts/deployment/utils/deploy_05_lp_swap_celo.sh`. Added a guard so the script fails loudly if LPSwapCelo hasn't been deployed yet, rather than quietly encoding a `null` recipient into governance calldata.

**Unchanged, and correct as-is:** the `CrossDomainMessenger` target, and `migrate()` on the old `CeloTargetDispenserL2` — that OLAS is dispenser funds, not swap input, so the mediator is the right recipient there.

Caught before the proposal was executed, so nothing needs recovering.

### DIFF
--- a/scripts/proposals/proposal_23_transfer_lp_token_celo.sh
+++ b/scripts/proposals/proposal_23_transfer_lp_token_celo.sh
@@ -37,8 +37,19 @@ celoL1StandardBridgeProxyAddress=$(jq -r '.celoL1StandardBridgeProxyAddress' $gl
 # Celo OLAS address (L2 token)
 celoOLASAddress=$(jq -r '.celoOLASAddress' $globalsStaking)
 
-# Bridge mediator on Celo (recipient of LP tokens)
+# Bridge mediator on Celo (recipient of the LPSwapCelo *outputs*, and migrate() target below)
 bridgeMediatorAddress=$(jq -r '.bridgeMediatorAddress' $globalsCelo)
+
+# LPSwapCelo on Celo — recipient of the bridged LP and the OLAS pre-funding.
+# swapLiquidity() reads both from its OWN balance (IToken(...).balanceOf(address(this))), so the
+# inputs must land on this contract. It then sends the re-added liquidity and any dust on to
+# BRIDGE_MEDIATOR itself. Bridging the inputs to the mediator instead would strand the migration
+# until governance forwarded them here in a second proposal.
+lpSwapCeloAddress=$(jq -r '.lpSwapCeloAddress' $globalsCelo)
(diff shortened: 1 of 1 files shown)


DavidMinarsch:
Two follow-ups after checking this more carefully — one is a bug in my own change.

**Fixed a wrong globals path.** The first version read `lpSwapCeloAddress` from the *staking* globals, but `deploy_05_lp_swap_celo.sh` writes it to `scripts/deployment/utils/globals_celo_mainnet.json` (`"$(dirname "$0")/globals_<network>.json"`) — the same file that already holds the `lpSwapMaxSlippage` constructor argument. The key would never have appeared in the file being read, so the new guard would have tripped on every run even after deployment. Now reads from the right file.

**Confirmed this proposal has not executed**, from both ends:

- L1: the Treasury still holds **210,254.80** of the LP token, so Step 1 (`Treasury.withdraw`) has not run.
- Celo: the old dispenser's `owner()` is still the bridge mediator rather than `0x0`, so Step 5's `migrate()` has not run either; and there is no LP at the mediator.

**And the intent is not ambiguous.** `LPSwapCelo.sol` and this proposal script were added in the *same* commit — `7856262d`, "feat and test: LPSwapCelo contract to migrate whOLAS-CELO LP to OLAS". The proposal exists to supply the swap contract, so bridging its inputs somewhere the swap cannot read them is an oversight rather than deliberate sequencing.

Note `lpSwapCeloAddress` is still absent from the utils globals, i.e. LPSwapCelo does not appear to be deployed yet. That is what the guard is for, and it means this change has not been exercised end-to-end — it should be run once after deployment to confirm the generated calldata carries the swap contract as recipient.

PR #352, opened 2026-08-31
docs: repair the vulnerabilities-list links, README paths and licence metadata

DavidMinarsch:
Documentation and repo hygiene. No contract changes.

## Every "Source code" link in the public vulnerabilities list was dead

**26 of its 27 relative links**, all the same mistake: written repo-root-relative (`contracts/Dispenser.sol`) from a file that lives in `docs/`, so each needed `../`. All 26 now resolve.

This is the document external researchers are pointed at, and the whole of it linked nowhere.

## Also

- **README paths:** `TokenomicsProxy` is at `contracts/proxies/`, and `EthereumDepositProcessor` carried a duplicated `.sol.sol` extension.
- **Heading levels:** eight mid-document sections used `#` (H1) — the oracle, LPSwapCelo and six fork-test blocks — so the README rendered as **nine sibling documents** with the testing instructions detached from `## Development`. Now `###`, matching the neighbouring `### Audits` / `### Linters` / `### Github workflows` which were already correct.
- **Licence metadata:** `package.json` said `"license": "ISC"` while the repo ships an **MIT** `LICENSE` and MIT SPDX headers. ISC appears nowhere else here — an untouched `npm init` default.
- **`SECURITY.md`** added; the repo lacked one despite being in scope for the bounty programme.
- **Bounty link** added; it was reachable from the governance README only.

Verified: 65 relative links across README and `docs/`, **0 broken**.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,6 +1,14 @@
 # Security Policy
 This document outlines security procedures and general policies for the `autonolas-tokenomics` project.
 
+## Bug Bounty
+Vulnerabilities in the deployed contracts are covered by the Autonolas bug bounty programme on Immunefi:
+https://immunefi.com/bounty/autonolas/
+
+Submitting through the programme is the preferred route for anything in its scope — it gives the report a
+tracked triage path and makes it eligible for a reward. Use the email below for anything outside that
+scope, or if you are unsure whether a finding qualifies.
+
 ## Reporting a Vulnerability
 The `autonolas-tokenomics` team and community take all security bugs in `autonolas-tokenomics` very seriously.
 Thank you for improving the security of `autonolas-tokenomics`. We appreciate your efforts and responsible disclosure and will make
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,27 @@
+# Security Policy
+This document outlines security procedures and general policies for the `autonolas-tokenomics` project.
+
+## Reporting a Vulnerability
+The `autonolas-tokenomics` team and community take all security bugs in `autonolas-tokenomics` very seriously.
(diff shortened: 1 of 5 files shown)


SECURITY.md
```diff
@@ -0,0 +1,27 @@
+# Security Policy
```
DavidMinarsch, line 1: Does this match the other repos? We should mention the bug bounty

DavidMinarsch: Both parts checked — and the answer to the second is broader than the file you commented on.

## Does it match the other repos? Yes, now exactly

The three pre-existing `SECURITY.md` files (governance, registries, marketplace) were **byte-identical to each other** apart from the repo name, and the one added here matched them. The only divergence was that the three originals were **missing a trailing newline** and mine had one. I added the newline to the three rather than removing it from the two new ones, since that is the correct form. All five now hash identically modulo the repo name.

## The bounty was missing from all five, not just this one

You are right that it should be mentioned — and **none** of the existing three mentioned it either. So the disclosure route every one of these files pointed a researcher at was the email, with no reference to the programme that actually covers these contracts.

All five now carry:

```
## Bug Bounty
Vulnerabilities in the deployed contracts are covered by the Autonolas bug bounty
programme on Immunefi: https://immunefi.com/bounty/autonolas/

Submitting through the programme is the preferred route for anything in its scope — it
gives the report a tracked triage path and makes it eligible for a reward. Use the email
below for anything outside that scope, or if you are unsure whether a finding qualifies.
```

Placed **before** `## Reporting a Vulnerability`, so the programme is what a reader meets first and the email reads as the fallback rather than the default. Pushed to all five branches.

PR #357, opened 2026-09-06
feat(staking): Robinhood Chain (4663) globals, and restore the missing dispenserProxyAddress

DavidMinarsch:
Config only, two parts.

## 1. Robinhood Chain (4663) globals

Robinhood Chain is an Arbitrum Orbit rollup settling to Ethereum with ETH gas, so it reuses `ArbitrumDepositProcessorL1` and `ArbitrumTargetDispenserL2` **unchanged**. All values verified against the live chains on 2026-09-06.

**L2** — `staking/arbitrum/globals_robinhood_mainnet.json`. `deploy_02_arbitrum_target_dispenser.sh` takes the globals suffix as `$1`, so `./deploy_02_arbitrum_target_dispenser.sh robinhood_mainnet` works with **no script change**. The `arbitrum*` key names hold Robinhood values — accurate for the bridge family, but worth knowing when reading the file.

**L1** — a `robinhood*` block in `staking/globals_mainnet.json`, following the per-chain prefixed convention already used for gnosis/optimism/celo/polygon/base/mode.

⚠️ **Six of the eight `ArbitrumDepositProcessorL1` constructor arguments are Robinhood-specific.** Reusing an Arbitrum One value would route funds into Arbitrum's escrow under a 4663 chain tag — a silent, unrecoverable mis-send. Diff the constructor args against the deployed Arbitrum processor before broadcasting; six of eight must differ.

Note `bridgeMediatorAddress` on the L2 file is byte-identical to Arbitrum's. That is correct — same L1 Timelock, and address aliasing is chain-independent.

## 2. `dispenserProxyAddress` was missing entirely — this affects every chain

`deploy_02_arbitrum_deposit_processor` (both `.js` and `.sh`) and `deploy_10_set_deposit_processors` all read **`dispenserProxyAddress`**, but **no globals file in the tree defines it**. `globals_mainnet.json` carries the value under `dispenserAddress`.

So `jq -r '.dispenserProxyAddress'` returns `null`, the processor is deployed with a zero `_l1Dispenser`, and `DefaultDepositProcessorL1`'s constructor reverts:

```solidity
if (_l1Dispenser == address(0) || _l1TokenRelayer == address(0) || _l1MessageRelayer == address(0)) {
    revert ZeroAddress();
}
```

This is not Robinhood-specific — the L1 deposit-processor path is broken for any chain today. Added alongside the existing `dispenserAddress` rather than renaming, so nothing that reads the old key breaks.

## Still outstanding — script changes, not config, so not in this PR

1. **`deploy_02_arbitrum_deposit_processor.sh` cannot deploy the Robinhood processor.** It reads `arbitrum*`-prefixed keys and writes back `arbitrumDepositProcessorL1Address`, so running it for 4663 would clobber Arbitrum One's recorded address. A `robinhood` variant of the script is needed.
2. **`deploy_10_set_deposit_processors.js` registers seven chains and omits both mode and robinhood.** The Mode gap is already tracked as `OLAS1-145` — the live mapping was completed by hand. A fresh Dispenser deployment would reproduce neither route.

Happy to follow up with both if you'd like them in the same cycle.

### DIFF
--- a/scripts/deployment/staking/_preflight_dispenser.sh
+++ b/scripts/deployment/staking/_preflight_dispenser.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+# Shared preflight for every script that binds or registers against the L1 Dispenser.
+#
+# WHY THIS EXISTS. DefaultDepositProcessorL1.l1Dispenser is `immutable`, so a processor deployed against the
+# wrong Dispenser can never be repaired — claims from the live one revert ManagerOnly. The trap is that
+# deploy_07b_dispenser_proxy.sh writes `dispenserProxyAddress` into scripts/deployment/globals_<network>.json,
+# a DIFFERENT file from the staking globals these scripts read, so a Dispenser proxy migration silently leaves
+# the value here pointing at the retired contract.
+#
+# Expects in scope: $globals, $dispenserProxyAddress, $1 (network suffix), and the red/green/reset vars.
+# Source it immediately after reading dispenserProxyAddress and before building any constructor args.
+
+zeroAddress="0x0000000000000000000000000000000000000000"
+if [ -z "$dispenserProxyAddress" ] || [ "$dispenserProxyAddress" == "null" ] \
+   || [ "$dispenserProxyAddress" == "$zeroAddress" ]; then
+  echo "${red}!!! dispenserProxyAddress is not set (or zero) in $globals${reset}"
+  exit 1
+fi
+
+deploymentGlobals="$(dirname "$0")/../globals_${1}.json"
+if [ ! -f "$deploymentGlobals" ]; then
+  # Warn rather than fail: not every network has a deployment-side globals file, and failing here would
+  # regress runs that work today. Silence would be the wrong shape for a guard against an unrecoverable
+  # binding, so it is loud.
+  echo "${red}!!! WARNING: cannot cross-check the Dispenser binding — $deploymentGlobals not found.${reset}"
+  echo "${red}!!!          l1Dispenser is immutable. Confirm $dispenserProxyAddress is the live Dispenser${reset}"
+  echo "${red}!!!          before continuing.${reset}"
+else
+  liveDispenser=$(jq -r '.dispenserProxyAddress // .dispenserAddress // empty' "$deploymentGlobals")
+  if [ -z "$liveDispenser" ]; then
+    echo "${red}!!! $deploymentGlobals has neither dispenserProxyAddress nor dispenserAddress${reset}"
+    exit 1
+  fi
+  if [ "$(echo "$liveDispenser" | tr '[:upper:]' '[:lower:]')" \
+     != "$(echo "$dispenserProxyAddress" | tr '[:upper:]' '[:lower:]')" ]; then
+    echo "${red}!!! Dispenser binding mismatch — refusing to bind to a stale address${reset}"
+    echo "${red}    $globals            dispenserProxyAddress = $dispenserProxyAddress${reset}"
+    echo "${red}    $deploymentGlobals  says the live Dispenser is $liveDispenser${reset}"
+    echo "${red}    Reconcile the two before deploying; l1Dispenser cannot be changed afterwards.${reset}"
+    exit 1
+  fi
+  echo "${green}Dispenser binding cross-checked against $deploymentGlobals${reset}"
+fi
--- a/scripts/deployment/staking/arbitrum/globals_robinhood_mainnet.json
+++ b/scripts/deployment/staking/arbitrum/globals_robinhood_mainnet.json
@@ -1,0 +1,17 @@
+{
+  "contractVerification": true,
+  "useLedger": true,
+  "derivationPath": "m/44'/60'/2'/0/0",
+  "providerName": "robinhood",
+  "chainId": "4663",
+  "networkURL": "https://rpc.mainnet.chain.robinhood.com",
+  "blockscoutURL": "https://robinhoodchain.blockscout.com",
+  "gasPriceInGwei": "1",
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
+  "serviceStakingFactoryAddress": "",
+  "bridgeMediatorAddress": "0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F",
+  "arbitrumArbSysAddress": "0x0000000000000000000000000000000000000064",
+  "l1ChainId": "1",
+  "arbitrumDepositProcessorL1Address": "",
+  "arbitrumTargetDispenserL2Address": ""
+}
--- a/scripts/deployment/staking/deploy_10_set_deposit_processors.js
+++ b/scripts/deployment/staking/deploy_10_set_deposit_processors.js
@@ -9,6 +9,23 @@
  * the claim, and in the batch path it takes the other chains' claims down with it.
  *
  * `deploy_10_set_deposit_processors.sh` is the equivalent shell route and is the preferred one.
+ *
+ * TWO CONSEQUENCES OF THE HARD FAIL, both deliberate:
+ *
+ *   1. This script is inoperative on `main` until wave 1 deploys. `robinhoodDepositProcessorL1Address`
+ *      ships empty, so until Robinhood Chain is live the script throws for EVERY chain, not just
+ *      Robinhood. That is the ordering rule above doing its job, but it means someone reaching for
+ *      this during an unrelated incident hits a wall that has nothing to do with their chain.
+ *
+ *   2. It cannot express "disable a route". `Dispenser.setDepositProcessorChainIds` treats a zero
+ *      processor as meaningful — "might be zero if there is a need to stop processing a specific L2
+ *      chain Id" — and `requireAddress` rejects zero. Disabling a chain is a targeted call with a
+ *      one-element array, not this bulk-register tool.
+ *
+ * NOTE ON PERMISSIONS: `Dispenser.owner()` is the Timelock, so on mainnet this call cannot be sent
+ * from the deploying EOA at all — it reverts OwnerOnly and belongs in a governance proposal. Both
+ * routes are therefore fresh-deployment tooling and a calldata reference, not an operational path
+ * against the live Dispenser.
  */
 
 const { ethers } = require("hardhat");
--- a/scripts/deployment/staking/deploy_10_set_deposit_processors.js
+++ b/scripts/deployment/staking/deploy_10_set_deposit_processors.js
@@ -107,6 +107,29 @@ async function main() {
     console.log("You are signing the following transaction: Dispenser.connect(EOA).setDepositProcessorChainIds()");
     const result = await dispenser.connect(EOA).setDepositProcessorChainIds(depositProcessors, chainIds);
     console.log("Transaction:", result.hash);
+
+    // Await the receipt. Submitting is not succeeding: with the Dispenser Timelock-owned, OwnerOnly is the
+    // likeliest failure here, and it reverts AFTER submission - so without this the script would print a
+    // hash and exit 0 on a transaction that did nothing.
+    const receipt = await result.wait();
+    if (receipt.status !== 1) {
+        throw new Error(`setDepositProcessorChainIds reverted in ${result.hash}`);
+    }
+
+    // Read back every route, mirroring deploy_10_set_deposit_processors.sh. A partially registered Dispenser
+    // is the exact state the hard-fail guards above exist to prevent, so it is verified rather than assumed.
+    let mismatch = false;
+    for (let i = 0; i < depositProcessors.length; i++) {
+        const onchain = await dispenser.mapChainIdDepositProcessors(chainIds[i]);
+        if (onchain.toLowerCase() !== depositProcessors[i].toLowerCase()) {
+            console.error(`  chainId ${chainIds[i]}: on-chain ${onchain} != ${depositProcessors[i]}`);
+            mismatch = true;
+        }
+    }
+    if (mismatch) {
+        throw new Error("One or more deposit processors did not take effect");
+    }
+    console.log("All deposit processors whitelisted");
 }
 
 main()
--- a/scripts/deployment/staking/deploy_10_set_deposit_processors.js
+++ b/scripts/deployment/staking/deploy_10_set_deposit_processors.js
@@ -1,34 +1,61 @@
 /*global process*/
 /*
- * NOTE: this script does NOT register the Mode processor.
+ * Registers every L1 deposit processor in the Dispenser.
  *
- * Mode's L1 processor is deployed at step 11, after this step runs, so its address does not exist
- * yet here. `deploy_10_set_deposit_processors.sh` is the current equivalent and registers all eight
- * routes including Mode; prefer it. A chain left unregistered resolves to a zero processor and
- * reverts the claim - in the batch path it takes the other chains' claims down with it - so a
- * fresh deployment that uses this script must register Mode separately afterwards.
+ * ORDERING: despite the file number, this is a terminal step rather than step 10 of a linear run.
+ * Mode's L1 processor is deployed at step 11 and Robinhood's at step 13, so this must run only once
+ * every per-chain processor deploy has completed. It hard-fails on a missing or zero entry rather
+ * than registering a partial set: a chain left unregistered resolves to a zero processor and reverts
+ * the claim, and in the batch path it takes the other chains' claims down with it.
+ *
+ * `deploy_10_set_deposit_processors.sh` is the equivalent shell route and is the preferred one.
  */
-
 
 const { ethers } = require("hardhat");
 const { LedgerSigner } = require("@anders-t/ethers-ledger");
+
+// Processor rows: label, globals key holding the address, globals key holding the L2 target chain Id.
+// The Ethereum row is L1-only, so its chain Id is the network's own rather than a globals key.
+const PROCESSOR_ROWS = [
+    ["Arbitrum", "arbitrumDepositProcessorL1Address", "arbitrumL2TargetChainId"],
+    ["Base", "baseDepositProcessorL1Address", "baseL2TargetChainId"],
+    ["Celo", "celoDepositProcessorL1Address", "celoL2TargetChainId"],
+    ["Gnosis", "gnosisDepositProcessorL1Address", "gnosisL2TargetChainId"],
+    ["Mode", "modeDepositProcessorL1Address", "modeL2TargetChainId"],
+    ["Optimism", "optimismDepositProcessorL1Address", "optimismL2TargetChainId"],
+    ["Polygon", "polygonDepositProcessorL1Address", "polygonL2TargetChainId"],
+    ["Robinhood", "robinhoodDepositProcessorL1Address", "robinhoodL2TargetChainId"],
+    ["Ethereum", "ethereumDepositProcessorAddress", null]
+];
+
+const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000";
+
+function requireAddress(parsedData, label, key) {
+    const value = parsedData[key];
+    if (!value || value === ZERO_ADDRESS) {
+        throw new Error(`${label}: ${key} is not set (or zero) in globals.json. Every processor must be ` +
+            "deployed before this step runs - see the ordering note at the top of this file.");
+    }
+    return value;
+}
+
+function requireChainId(parsedData, label, key) {
+    const value = parsedData[key];
+    if (!value || Number(value) === 0) {
+        throw new Error(`${label}: ${key} is not set (or zero) in globals.json.`);
+    }
+    return value;
+}
 
 async function main() {
     const fs = require("fs");
     const globalsFile = "globals.json";
     const dataFromJSON = fs.readFileSync(globalsFile, "utf8");
-    let parsedData = JSON.parse(dataFromJSON);
+    const parsedData = JSON.parse(dataFromJSON);
     const useLedger = parsedData.useLedger;
     const derivationPath = parsedData.derivationPath;
     const providerName = parsedData.providerName;
-    const arbitrumDepositProcessorL1Address = parsedData.arbitrumDepositProcessorL1Address;
-    const baseDepositProcessorL1Address = parsedData.baseDepositProcessorL1Address;
-    const celoDepositProcessorL1Address = parsedData.celoDepositProcessorL1Address;
-    const gnosisDepositProcessorL1Address = parsedData.gnosisDepositProcessorL1Address;
-    const optimismDepositProcessorL1Address = parsedData.optimismDepositProcessorL1Address;
-    const polygonDepositProcessorL1Address = parsedData.polygonDepositProcessorL1Address;
-    const ethereumDepositProcessorAddress = parsedData.ethereumDepositProcessorAddress;
-    const dispenserProxyAddress = parsedData.dispenserProxyAddress;
+    const dispenserProxyAddress = requireAddress(parsedData, "Dispenser", "dispenserProxyAddress");
     let EOA;
 
     const provider = await ethers.providers.getDefaultProvider(providerName);
@@ -43,24 +70,25 @@ async function main() {
     const deployer = await EOA.getAddress();
     console.log("EOA is:", deployer);
 
-    // Get all the contracts
-    const arbitrumDepositProcessorL1 = await ethers.getContractAt("ArbitrumDepositProcessorL1", arbitrumDepositProcessorL1Address);
-    const baseDepositProcessorL1 = await ethers.getContractAt("OptimismDepositProcessorL1", baseDepositProcessorL1Address);
-    const celoDepositProcessorL1 = await ethers.getContractAt("OptimismDepositProcessorL1", celoDepositProcessorL1Address);
-    const gnosisDepositProcessorL1 = await ethers.getContractAt("GnosisDepositProcessorL1", gnosisDepositProcessorL1Address);
-    const optimismDepositProcessorL1 = await ethers.getContractAt("OptimismDepositProcessorL1", optimismDepositProcessorL1Address);
-    const polygonDepositProcessorL1 = await ethers.getContractAt("PolygonDepositProcessorL1", polygonDepositProcessorL1Address);
+    const ethereumChainId = (await provider.getNetwork()).chainId;
+
+    // Build the two aligned arrays, hard-failing on any missing or zero entry
+    const depositProcessors = [];
+    const chainIds = [];
+    for (const [label, addressKey, chainIdKey] of PROCESSOR_ROWS) {
+        const depositProcessor = requireAddress(parsedData, label, addressKey);
+        const chainId = chainIdKey === null ? ethereumChainId : requireChainId(parsedData, label, chainIdKey);
+        depositProcessors.push(depositProcessor);
+        chainIds.push(chainId);
+        console.log(`  ${label}: ${depositProcessor} -> chainId ${chainId}`);
+    }
+
     const dispenser = await ethers.getContractAt("Dispenser", dispenserProxyAddress);
 
     // Transaction signing and execution
     console.log("10. EOA to set deposit processors in Dispenser");
     console.log("You are signing the following transaction: Dispenser.connect(EOA).setDepositProcessorChainIds()");
-    const ethereumChainId = (await provider.getNetwork()).chainId;
-    const result = await dispenser.connect(EOA).setDepositProcessorChainIds([arbitrumDepositProcessorL1Address,
-        baseDepositProcessorL1Address, celoDepositProcessorL1Address, ethereumDepositProcessorAddress,
-        gnosisDepositProcessorL1Address, optimismDepositProcessorL1Address, polygonDepositProcessorL1Address],
-    [parsedData.arbitrumL2TargetChainId, parsedData.baseL2TargetChainId, parsedData.celoL2TargetChainId, ethereumChainId,
-        parsedData.gnosisL2TargetChainId, parsedData.optimismL2TargetChainId, parsedData.polygonL2TargetChainId]);
+    const result = await dispenser.connect(EOA).setDepositProcessorChainIds(depositProcessors, chainIds);
     console.log("Transaction:", result.hash);
 }
 
--- a/scripts/deployment/staking/deploy_10_set_deposit_processors.sh
+++ b/scripts/deployment/staking/deploy_10_set_deposit_processors.sh
@@ -68,6 +68,7 @@ rows=(
   "Optimism optimismDepositProcessorL1Address optimismL2TargetChainId"
   "Polygon polygonDepositProcessorL1Address polygonL2TargetChainId"
   "Mode modeDepositProcessorL1Address modeL2TargetChainId"
+  "Robinhood robinhoodDepositProcessorL1Address robinhoodL2TargetChainId"
   "Ethereum ethereumDepositProcessorAddress __L1__"
 )
 
--- a/scripts/deployment/staking/deploy_10_set_deposit_processors.sh
+++ b/scripts/deployment/staking/deploy_10_set_deposit_processors.sh
@@ -50,6 +50,9 @@ chainId=$(jq -r '.chainId' $globals)
 networkURL=$(jq -r '.networkURL' $globals)
 dispenserProxyAddress=$(jq -r '.dispenserProxyAddress' $globals)
 
+# Preflight: cross-check the Dispenser binding (see _preflight_dispenser.sh).
+. "$(dirname "$0")/_preflight_dispenser.sh"
+
 zeroAddress="0x0000000000000000000000000000000000000000"
 
 if [ -z "$dispenserProxyAddress" ] || [ "$dispenserProxyAddress" == "null" ] \
--- a/scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
+++ b/scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
@@ -8,14 +8,14 @@ reset=$(tput sgr0)
 globals="$(dirname "$0")/globals_$1.json"
 if [ ! -f $globals ]; then
   echo "${red}!!! $globals is not found${reset}"
-  exit 0
+  exit 1
 fi
 
 # Get globals file for L2
 globalsL2="$(dirname "$0")/robinhood/globals_robinhood_$1.json"
 if [ ! -f $globalsL2 ]; then
   echo "${red}!!! $globalsL2 is not found${reset}"
-  exit 0
+  exit 1
 fi
 
 # Read variables using jq
@@ -33,6 +33,29 @@ robinhoodL2TargetChainId=$(jq -r '.robinhoodL2TargetChainId' $globals)
 robinhoodL1ERC20GatewayAddress=$(jq -r '.robinhoodL1ERC20GatewayAddress' $globals)
 robinhoodOutboxAddress=$(jq -r '.robinhoodOutboxAddress' $globals)
 robinhoodBridgeAddress=$(jq -r '.robinhoodBridgeAddress' $globals)
+
+# Preflight on the Dispenser binding. l1Dispenser is `immutable` in DefaultDepositProcessorL1, so a processor
+# deployed against the wrong Dispenser cannot be repaired — claims from the live one revert ManagerOnly.
+# deploy_07b_dispenser_proxy.sh writes dispenserProxyAddress into scripts/deployment/globals_<network>.json,
+# a DIFFERENT file from this one, so a proxy migration silently leaves the value here stale.
+zeroAddress="0x0000000000000000000000000000000000000000"
+if [ -z "$dispenserProxyAddress" ] || [ "$dispenserProxyAddress" == "null" ] \
+   || [ "$dispenserProxyAddress" == "$zeroAddress" ]; then
+  echo "${red}!!! dispenserProxyAddress is not set (or zero) in $globals${reset}"
+  exit 1
+fi
+deploymentGlobals="$(dirname "$0")/../globals_${1}.json"
+if [ -f "$deploymentGlobals" ]; then
+  liveDispenser=$(jq -r '.dispenserProxyAddress // .dispenserAddress // empty' "$deploymentGlobals")
+  if [ -n "$liveDispenser" ] \
+     && [ "$(echo "$liveDispenser" | tr '[:upper:]' '[:lower:]')" != "$(echo "$dispenserProxyAddress" | tr '[:upper:]' '[:lower:]')" ]; then
+    echo "${red}!!! Dispenser binding mismatch — refusing to deploy an immutable binding to a stale address${reset}"
+    echo "${red}    $globals            dispenserProxyAddress = $dispenserProxyAddress${reset}"
+    echo "${red}    $deploymentGlobals  says the live Dispenser is $liveDispenser${reset}"
+    echo "${red}    Reconcile the two before deploying; l1Dispenser cannot be changed afterwards.${reset}"
+    exit 1
+  fi
+fi
 
 
 # Check for Alchemy keys
@@ -79,7 +102,7 @@ outputLength=${#robinhoodDepositProcessorL1Address}
 # Check for the deployed address
 if [ $outputLength != 42 ]; then
   echo "${red}!!! The contract was not deployed...${reset}"
-  exit 0
+  exit 1
 fi
 
 # Write new deployed contract back into JSON
--- a/scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
+++ b/scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+reset=$(tput sgr0)
+
+# Get globals file
+globals="$(dirname "$0")/globals_$1.json"
+if [ ! -f $globals ]; then
+  echo "${red}!!! $globals is not found${reset}"
+  exit 0
+fi
+
+# Get globals file for L2
+globalsL2="$(dirname "$0")/robinhood/globals_robinhood_$1.json"
+if [ ! -f $globalsL2 ]; then
+  echo "${red}!!! $globalsL2 is not found${reset}"
+  exit 0
+fi
+
+# Read variables using jq
+contractVerification=$(jq -r '.contractVerification' $globals)
+useLedger=$(jq -r '.useLedger' $globals)
+derivationPath=$(jq -r '.derivationPath' $globals)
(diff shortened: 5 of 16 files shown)


scripts/deployment/staking/robinhood/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,17 @@
...
+  "useLedger": true,
+  "derivationPath": "m/44'/60'/2'/0/0",
+  "providerName": "robinhood",
+  "chainId": "4663",
+  "networkURL": "https://rpc.mainnet.chain.robinhood.com",
+  "blockscoutURL": "https://robinhoodchain.blockscout.com",
+  "gasPriceInGwei": "1",
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
+  "serviceStakingFactoryAddress": "",
+  "bridgeMediatorAddress": "0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F",
```
DavidMinarsch, line 12: L2 alias of L1 timelock

scripts/deployment/staking/arbitrum/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,17 @@
...
+  "derivationPath": "m/44'/60'/2'/0/0",
+  "providerName": "robinhood",
+  "chainId": "4663",
+  "networkURL": "https://rpc.mainnet.chain.robinhood.com",
+  "blockscoutURL": "https://robinhoodchain.blockscout.com",
+  "gasPriceInGwei": "1",
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
+  "serviceStakingFactoryAddress": "",
+  "bridgeMediatorAddress": "0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F",
+  "arbitrumArbSysAddress": "0x0000000000000000000000000000000000000064",
```
DavidMinarsch: Validated against arbitrum

DavidMinarsch:
Both validations confirmed, and the PR has grown to close **D2** as well (`142030b`).

## A correction to my own first attempt

I'd put the L2 globals in `staking/arbitrum/`, reasoning that `deploy_02_arbitrum_target_dispenser.sh` could be reused with a `robinhood` suffix. It can — but **`script_01_set_target_dispenser_l2.sh` can't**. That script is fully generic: it derives the network from `$1` and reads `${network}DepositProcessorL1Address` from the L1 globals and `${network}TargetDispenserL2Address` from `${network}/globals_$1.json`. Reusing `arbitrum/` would have left the L1↔L2 wiring step with no file to read.

So the globals moved to **`staking/robinhood/`** with `robinhood`-prefixed keys, which is what the generic script already expects.

## D2 — now closed here

- **`deploy_13_robinhood_deposit_processor.sh`** and **`robinhood/deploy_02_robinhood_target_dispenser.sh`**, both generated from the Arbitrum originals by key renaming only, so they stay in lockstep. Contract paths are unchanged — Robinhood Chain is an Orbit rollup and uses `ArbitrumDepositProcessorL1` / `ArbitrumTargetDispenserL2` verbatim.
- **A Robinhood row in `deploy_10_set_deposit_processors.sh`.**

Worth flagging what I found while doing it: **the shell route is in better shape than the `.js` twin.** `deploy_10_set_deposit_processors.sh` already carries Mode and already hard-fails on a missing or zero `dispenserProxyAddress`. It's `deploy_10_set_deposit_processors.js` that's stale — it omits both Mode and Robinhood. I've left it alone, since `OLAS1-6` already resolved to retiring the JS path rather than maintaining two; say the word if you'd rather it were fixed in this cycle.

Every key the three scripts read is present in the globals, checked mechanically.

## On the two you validated

`bridgeMediatorAddress` is the L2 alias of the L1 Timelock, re-derived from scratch and matching, and identical to Arbitrum's for the aliasing reason. The `arbitrum`-validated values check out too — worth noting six of the eight `ArbitrumDepositProcessorL1` constructor args must *differ* from Arbitrum One's, which is the check I'd run hardest before broadcasting.

DavidMinarsch:
JS route fixed too (`7407629`). CI green — build, gitleaks and the gate all pass.

Rather than bolt two rows onto a list that already disagreed with the shell version, I restructured it to the **same table-driven shape as `deploy_10_set_deposit_processors.sh`**: one row per chain, both arrays built from it, and a hard failure on any missing or zero address or chain Id. All nine routes are now registered — the seven it had, plus Mode and Robinhood.

**The note at the top needed correcting, not just extending.** It said the script omits Mode because Mode's processor is deployed at step 11, after step 10 runs. That's the symptom rather than the constraint: the real point is that this is a **terminal step, not step 10 of a linear run**, and must follow every per-chain processor deploy. The shell route already says exactly that. Both now agree.

Hard-failing rather than filtering is deliberate, and matches the shell script: registering a partial set is the outcome worth preventing, since an unregistered chain resolves to a zero processor and reverts the claim — and in the batch path takes other chains' claims down with it.

**One pre-existing thing I noticed but did not change:** `globals_sepolia.json` has neither `ethereumDepositProcessorAddress` nor the robinhood keys, so *both* routes hard-fail there. That's the shell script's existing behaviour rather than anything introduced here — worth a separate look if the sepolia path is still exercised.

scripts/deployment/staking/globals_mainnet.json
```diff
@@ -58,5 +58,15 @@
...
   "modeDepositProcessorL1Address": "0x6A5F38aAdCc1c049E39D80147b64a9Ff947d208B",
-  "modeTargetDispenserL2Address": "0xEB5638eefE289691EcE01943f768EDBF96258a80"
+  "modeTargetDispenserL2Address": "0xEB5638eefE289691EcE01943f768EDBF96258a80",
+  "dispenserProxyAddress": "0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0",
+  "robinhoodL1ERC20GatewayRouterAddress": "0x6a2E3a1e16FC29f27Ce61429746D558d656975bB",
+  "robinhoodL1ERC20GatewayAddress": "0x85001CC4867C5e1C22dA4B79BB8852B9e2a06da0",
+  "robinhoodInboxAddress": "0x1A07cc4BD17E0118BdB54D70990D2158AbAD7a2D",
+  "robinhoodOutboxAddress": "0xf0ce991ea4A0d2400A4AB49b20ae333f6Dce3DE9",
+  "robinhoodBridgeAddress": "0xDf8755334ce7A73cCF6b581C02eA649AE3E864b3",
+  "robinhoodOLASAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
```
OjusWiZard: **Fails EIP-55 checksum.**

```
here       0x092963938DEBd8013a2E545B3549f8a5EC0d2286
correct    0x092963938deBD8013a2e545b3549f8A5ec0D2286
```

The address is right case-insensitively — it matches what both the L1 and L2 gateway routers return from `calculateL2TokenAddress` — but the casing is invalid. Same string appears in `autonolas-registries#324` and in the plan doc, so it is one transcription artifact copied three ways; worth fixing in one pass.

Tolerated by `jq`/`cast`, rejected by ethers' `getAddress()`, which throws on a **mixed-case** address failing EIP-55 (all-lowercase would pass) — so the repo's `.js` deploy and verify scripts are the ones that would break.

---

**Separately, this key looks like dead config.** For the OP-stack chains, `<chain>OLASAddress` is a live 6th constructor arg — `deploy_04`/`05`/`07`/`11` all read `optimismOLASAddress`, `celoOLASAddress`, `baseOLASAddress`, `modeOLASAddress`. `ArbitrumDepositProcessorL1` takes no L2-OLAS argument, which is why `arbitrumOLASAddress` does not exist, and `deploy_13` correctly does not read this key.

So this line puts the **L2** token into the L1 globals, next to `olasAddress` holding the **L1** one, in a file where every sibling key of that shape is wired into a constructor. Given this PR's own mis-send warning, either drop it or add a note saying it is reference-only.

scripts/deployment/staking/robinhood/globals_robinhood_mainnet.json
```diff
@@ -0,0 +1,17 @@
+{
+  "contractVerification": true,
+  "useLedger": true,
+  "derivationPath": "m/44'/60'/2'/0/0",
+  "providerName": "robinhood",
+  "chainId": "4663",
+  "networkURL": "https://rpc.mainnet.chain.robinhood.com",
+  "blockscoutURL": "https://robinhoodchain.blockscout.com",
+  "gasPriceInGwei": "1",
+  "olasAddress": "0x092963938DEBd8013a2E545B3549f8a5EC0d2286",
```
OjusWiZard: **Fails EIP-55 checksum** — same string as `robinhoodOLASAddress` in `globals_mainnet.json`:

```
here       0x092963938DEBd8013a2E545B3549f8a5EC0d2286
correct    0x092963938deBD8013a2e545b3549f8A5ec0D2286
```

This one *is* consumed — `deploy_02_robinhood_target_dispenser.sh` passes it as `_olas` to `ArbitrumTargetDispenserL2`. The shell route tolerates it via `jq`/`cast`; the `.js` path would not.

scripts/deployment/staking/deploy_10_set_deposit_processors.js
```diff
@@ -1,34 +1,61 @@
...
+    ["Polygon", "polygonDepositProcessorL1Address", "polygonL2TargetChainId"],
+    ["Robinhood", "robinhoodDepositProcessorL1Address", "robinhoodL2TargetChainId"],
+    ["Ethereum", "ethereumDepositProcessorAddress", null]
+];
+
+const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000";
+
+function requireAddress(parsedData, label, key) {
+    const value = parsedData[key];
+    if (!value || value === ZERO_ADDRESS) {
```
OjusWiZard, line 52: Two notes on the hard-fail, both minor — the change itself is a clear improvement over silent partial registration.

**1. This makes the script inoperative on `main` until wave 1 deploys.** `robinhoodDepositProcessorL1Address` ships as `""`, so from merge until 4663 is live, `deploy_10` throws for every chain, not just Robinhood. That is intentional per the ordering note at the top, but worth stating explicitly somewhere a reader will hit it — someone reaching for this during an incident runs into a wall that has nothing to do with their chain.

**2. It forbids a zero address that `Dispenser` treats as meaningful.** From `setDepositProcessorChainIds`:

> `// Note: depositProcessors[i] might be zero if there is a need to stop processing a specific L2 chain Id`

`requireAddress` rejects `ZERO_ADDRESS`, so this script can no longer express "disable a route". Probably right for a bulk-register tool — but it is now the only script that does this call, so that capability has no home. Worth a line in the header note saying disabling is a targeted call, not this.

OjusWiZard (dismissed):
Approving. I verified the constructor wiring against the contract sources rather than the description, since a wrong argument here is exactly the "silent, unrecoverable mis-send" this PR warns about. The wiring is right.

**Both constructor arg lists match the contracts exactly:**

| | script order | contract |
|---|---|---|
| `ArbitrumDepositProcessorL1` | `olas, dispenserProxy, gatewayRouter, inbox, chainId, erc20Gateway, outbox, bridge` | `_olas, _l1Dispenser, _l1TokenRelayer, _l1MessageRelayer, _l2TargetChainId, _l1ERC20Gateway, _outbox, _bridge` |
| `ArbitrumTargetDispenserL2` | `olas, stakingFactory, arbSys, l1DepositProcessor, l1ChainId` | `_olas, _proxyFactory, _l2MessageRelayer, _l1DepositProcessor, _l1SourceChainId` |

The `cast abi-encode` signatures used for verification match both, `uint256` in position 5 included.

`deploy_13` reads `olasAddress` for `_olas` — the **L1** token `0x0001A500…`, not the bridged L2 one. That is the most dangerous confusion available in this change and it is correct. The **six of eight** claim is also exactly right: `_olas` and `_l1Dispenser` come from shared keys, the other six from `robinhood*` keys, and I checked that all six differ from Arbitrum's values.

D2 is genuinely fixed — `deploy_13` writes `robinhoodDepositProcessorL1Address` into both globals and never touches an `arbitrum*` key. D1 is fixed correctly — the value matches the existing `dispenserAddress` and is added alongside rather than renaming.

`deploy_10.js` is better than advertised: it closes the Mode omission (the `OLAS1-145` script gap) as well as adding Robinhood, and turns silent partial registration into a hard fail. The reordering is safe — `setDepositProcessorChainIds` is a keyed loop (`mapChainIdDepositProcessors[chainIds[i]] = depositProcessors[i]`), so array order carries no meaning.

---

### One thing that is not inline-able: the PR description is stale and understates the change

Three contradictions with the actual diff:

- **"Config only, two parts"** — this PR adds two new shell scripts (105 + 101 lines) and rewrites `deploy_10.js` (+57 −29).
- **"Still outstanding — script changes, not config, so not in this PR"**, listing the `deploy_02` clobbering problem and the `deploy_10` omissions — **both are in this PR**, as `deploy_13_robinhood_deposit_processor.sh` and the `deploy_10` rewrite.
- **"L2 — `staking/arbitrum/globals_robinhood_mainnet.json` … works with no script change"** — it is at `staking/robinhood/`, with a purpose-built `deploy_02_robinhood_target_dispenser.sh`.

The plan doc's §9 records this correction; the body never caught up. Worth fixing before merge, because a reviewer who trusts "config only, and those two items aren't here" will skim past the riskiest 200 lines in the change.

Inline comments below: one blocking-ish nit repeated twice (EIP-55) and two advisories.

scripts/deployment/staking/globals_mainnet.json
```diff
@@ -58,5 +58,15 @@
   "modeOLASAddress": "0xcfD1D50ce23C46D3Cf6407487B2F8934e96DC8f9",
   "modeL2TargetChainId": "34443",
   "modeDepositProcessorL1Address": "0x6A5F38aAdCc1c049E39D80147b64a9Ff947d208B",
-  "modeTargetDispenserL2Address": "0xEB5638eefE289691EcE01943f768EDBF96258a80"
+  "modeTargetDispenserL2Address": "0xEB5638eefE289691EcE01943f768EDBF96258a80",
+  "dispenserProxyAddress": "0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0",
```
dagacha, line 62: **[P1] Use the deployed migration proxy address instead of aliasing the legacy Dispenser**

The verified contract at `0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0` is the existing non-proxied Dispenser, not the new `DispenserProxy`. This matters for the migration documented in `docs/dispenser_migration_runbook_public.md`: `deploy_07b_dispenser_proxy.sh` writes the new proxy into `scripts/deployment/globals_mainnet.json`, whereas the processor deployment scripts read this separate `scripts/deployment/staking/globals_mainnet.json`. Unless the operator manually replaces this newly populated value, deploying the proxy and then the processors succeeds with the processors still bound to the old Dispenser. Their `l1Dispenser` is immutable, so claims sent by the new proxy revert `ManagerOnly`, and correcting the JSON afterwards cannot repair the deployed contracts. Please read/synchronize the actual proxy deployment output before constructing processors, or keep an explicitly separate legacy-stack configuration with a migration preflight that rejects this stale binding.

Evidence: [verified legacy Dispenser](https://eth.blockscout.com/address/0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0?tab=contract).

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
```diff
@@ -0,0 +1,105 @@
...
+deploymentOutput=$($execCmd)
+robinhoodDepositProcessorL1Address=$(echo "$deploymentOutput" | grep 'Deployed to:' | awk '{print $3}')
+
+# Get output length
+outputLength=${#robinhoodDepositProcessorL1Address}
+
+# Check for the deployed address
+if [ $outputLength != 42 ]; then
+  echo "${red}!!! The contract was not deployed...${reset}"
+  exit 0
```
dagacha: **[P2] Return failure when Forge does not deploy the contract**

If `forge create` fails (for example, a rejected Ledger signature, RPC error, or invalid constructor arguments), no deployment address is captured and this branch returns `0`. I reproduced this with a mocked Forge command that exits `1`: both this script and `robinhood/deploy_02_robinhood_target_dispenser.sh` print that the contract was not deployed but exit successfully. An `&&`-chained deployment therefore continues with missing or stale addresses instead of stopping. Please propagate the Forge command's failure and return a nonzero status when the address is missing; apply the same correction to the L2 script's corresponding branch.

dagacha (commented):
Review performed with Codex using GPT-6 Astra (`gpt-6-astra`), with **high** reasoning effort, as recorded in the reviewing session's metadata.

Reviewed commit `740762954c28e9a793ba7ad262f6e38d24481be2`. Two findings are attached inline: the legacy Dispenser address being supplied as the migration proxy, and successful exit codes on failed deployments in both new shell scripts.

Validation: checked both constructor argument lists against the Solidity contracts, passed shell and JavaScript syntax checks, inspected the public explorer's verified Dispenser ABI, and reproduced the deployment-failure behavior in isolated script copies with mocked Forge/Cast commands. No transactions were broadcast and no repository files were changed.

scripts/deployment/staking/globals_mainnet.json
```diff
@@ -58,5 +58,15 @@
...
+  "robinhoodL1ERC20GatewayRouterAddress": "0x6a2E3a1e16FC29f27Ce61429746D558d656975bB",
```
bennyjo, line 63: On every other chain the marketplace balance trackers drain to this repo's BuyBackBurnerProxy, and valory-xyz/autonolas-marketplace#195 leaves `drainerAddress` empty, but there are no `oracles/` or `utils/` Robinhood globals here; is the 4663 buyback stack a follow-up PR?

bennyjo (commented):
Not approving yet. Independently verified all 20 bridge-topology checks and the six-of-eight constructor diff against the live Arbitrum processor; both new scripts drift zero from their Arbitrum siblings. Two things hold this back: the `dispenserProxyAddress` binding question already raised inline needs an explicit decision in the PR, and the 4663 buyback stack has no globals here (inline below). Note that Robinhood's Blockscout API returns a Cloudflare 403 to non-browser clients, so the Blockscout verify leg fails as well as the Etherscan one.

DavidMinarsch:
All five items addressed in `e054217`. CI green.

**@dagacha P1 — Dispenser binding preflight.** This was the most valuable finding in the round, and it was partly created by my own D1 fix: adding `dispenserProxyAddress` to the *staking* globals duplicated a value that `deploy_07b_dispenser_proxy.sh` writes to `scripts/deployment/globals_<network>.json` — a different file. `l1Dispenser` is immutable, so a proxy migration would have bound a new processor to the retired Dispenser with no way to repair it.

`deploy_13` now refuses to deploy unless the two files agree, and refuses a zero or unset value outright. Tested in a sandbox rather than asserted:

```
A: stale binding   -> "Dispenser binding mismatch — refusing to deploy an immutable binding
                       to a stale address", exit 1
B: matching pair   -> preflight passes, proceeds to forge
C: zero/unset      -> "dispenserProxyAddress is not set (or zero)", exit 1
```

The comparison is case-insensitive so checksum casing can't trip it, and the guard prints both paths and both values so the operator sees exactly what disagrees.

**@dagacha P2 — exit codes.** Both new scripts now `exit 1` when `forge create` produces no address; verified the same way (`exit code: 1`, was `0`). I extended it to the globals-not-found branches too, same class of silent success. This is a deliberate divergence from the Arbitrum originals they were generated from — those share the bug and are worth the same fix, but that's a broader change than this PR.

**@OjusWiZard — EIP-55.** Fixed here and in the three other artifacts in one pass, as you suggested.

**@OjusWiZard — `robinhoodOLASAddress` is dead config.** Dropped entirely rather than annotated. You're right that `ArbitrumDepositProcessorL1` takes no L2-OLAS argument — which is exactly why `arbitrumOLASAddress` doesn't exist — and `deploy_13` never read it. Keeping an L2 token address beside the L1 `olasAddress`, in a file where every sibling key of that shape feeds a constructor, was the confusion this PR's own mis-send warning is about.

**@OjusWiZard — `deploy_10.js` header.** Both points recorded, plus a third I owe to @bennyjo's review of the plan: `Dispenser.owner()` is the Timelock, so **neither route can run against the live Dispenser from the EOA** — it reverts `OwnerOnly`. Registering 4663 is a governance entry; the script change is hygiene so a *fresh* Dispenser deployment reproduces all nine routes. Your "disable a route" point is in there verbatim as the reason that capability now lives in a targeted one-element call.

**@bennyjo — the 4663 buyback stack is a follow-up PR**, confirmed. Worth flagging the consequence you'll have spotted: `BalanceTrackerBase.drainer` is immutable and reverts on zero, and the fleet's only exception is the L1 OLAS tracker draining straight to the `Burner` — L1-only. So the marketplace balance trackers defer with the buyback stack, and `drainerAddress` staying empty in marketplace#195 is correct rather than an omission. Wave 1 ships Karma, MechMarketplace and the two factories; payments wait for the follow-up. That's now written into the rollout plan's §3.5/§3.6.

scripts/deployment/staking/deploy_10_set_deposit_processors.js
```diff
@@ -43,24 +87,25 @@ async function main() {
...
     // Transaction signing and execution
     console.log("10. EOA to set deposit processors in Dispenser");
     console.log("You are signing the following transaction: Dispenser.connect(EOA).setDepositProcessorChainIds()");
-    const ethereumChainId = (await provider.getNetwork()).chainId;
-    const result = await dispenser.connect(EOA).setDepositProcessorChainIds([arbitrumDepositProcessorL1Address,
-        baseDepositProcessorL1Address, celoDepositProcessorL1Address, ethereumDepositProcessorAddress,
-        gnosisDepositProcessorL1Address, optimismDepositProcessorL1Address, polygonDepositProcessorL1Address],
-    [parsedData.arbitrumL2TargetChainId, parsedData.baseL2TargetChainId, parsedData.celoL2TargetChainId, ethereumChainId,
-        parsedData.gnosisL2TargetChainId, parsedData.optimismL2TargetChainId, parsedData.polygonL2TargetChainId]);
+    const result = await dispenser.connect(EOA).setDepositProcessorChainIds(depositProcessors, chainIds);
```
OjusWiZard, line 108: **This route submits and exits without awaiting the receipt; the `.sh` route verifies its effect on-chain. Both were touched in this PR.**

This line submits; the next logs `result.hash`, and the promise chain then calls `process.exit(0)`. There is no `.wait()`, so a transaction that reverts after submission is reported as success.

The shell counterpart added a post-send check in this same PR (`deploy_10_set_deposit_processors.sh:143-156`):

```bash
for i in "${!chainIds[@]}"; do
  onchain=$(cast call ... "mapChainIdDepositProcessors(uint256)(address)" "${chainIds[$i]}")
  ... mismatch=1
done
[ "$mismatch" != "0" ] && exit 1
```

So the two routes register the same nine chains but do **not** have equivalent failure detection.

What makes this worth closing rather than noting: this file's own header warns the call is owner-gated once the Dispenser is Timelock-owned. `OwnerOnly` is therefore the single likeliest failure mode, and it is precisely the one that would print a transaction hash and exit 0. A partially-registered Dispenser is the exact state the hard-fail guards above were added to prevent.

`await result.wait()` and a non-zero exit on a reverted receipt would close it. Mirroring the `.sh` loop's `mapChainIdDepositProcessors` read-back would close it properly. (The same gap exists in the untouched `deploy_09_set_targer_dispensers.js`, so it is the pre-existing convention — but the `.sh` sibling has now moved past it.)

scripts/deployment/staking/robinhood/deploy_02_robinhood_target_dispenser.sh
```diff
@@ -0,0 +1,101 @@
...
+    80002)    API_KEY=$ALCHEMY_API_KEY_AMOY;    keyName="ALCHEMY_API_KEY_AMOY" ;;
+  esac
+  if [ -n "$keyName" ] && [ "$API_KEY" == "" ]; then
+    echo "set $keyName env variable"
+    exit 0
+  fi
+fi
+
+contractPath="contracts/staking/ArbitrumTargetDispenserL2.sol:ArbitrumTargetDispenserL2"
+constructorArgs="$olasAddress $serviceStakingFactoryAddress $robinhoodArbSysAddress $robinhoodDepositProcessorL1Address $l1ChainId"
```
OjusWiZard, line 67: **No validation on any of these five inputs, where the sibling script in this same PR gained a preflight.**

`deploy_13` now guards `dispenserProxyAddress` before building its constructor args. This script guards nothing, and `serviceStakingFactoryAddress` (line 29) currently ships as `""` in `robinhood/globals_robinhood_mainnet.json`.

Because `--constructor-args $constructorArgs` is unquoted, an empty value does not produce an empty positional token — word-splitting collapses it, so a five-parameter constructor receives four arguments with `l1ChainId` dropped. Today that fails loudly at forge's arity check, and the `outputLength != 42` guard catches the rest. The concern is the next case: a stale-but-well-formed address yields five syntactically valid arguments and deploys successfully against a wrong **immutable** binding, with nothing here to notice.

`robinhoodDepositProcessorL1Address` is the field most exposed to that, since it is populated by `deploy_13`'s write-back rather than by hand.

Suggest the same empty/`"null"`/zero-address rejection `deploy_13` and `deploy_10.sh` already use, applied to the four non-constant fields before line 49.

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
```diff
@@ -0,0 +1,128 @@
...
+# deploy_07b_dispenser_proxy.sh writes dispenserProxyAddress into scripts/deployment/globals_<network>.json,
+# a DIFFERENT file from this one, so a proxy migration silently leaves the value here stale.
+zeroAddress="0x0000000000000000000000000000000000000000"
+if [ -z "$dispenserProxyAddress" ] || [ "$dispenserProxyAddress" == "null" ] \
+   || [ "$dispenserProxyAddress" == "$zeroAddress" ]; then
+  echo "${red}!!! dispenserProxyAddress is not set (or zero) in $globals${reset}"
+  exit 1
+fi
+deploymentGlobals="$(dirname "$0")/../globals_${1}.json"
+if [ -f "$deploymentGlobals" ]; then
```
OjusWiZard, line 58: **The guard silently no-ops when this file is absent — including no warning.**

`scripts/deployment/` currently holds only `globals_mainnet.json` and `globals_legacy_goerli.json`; there is no `globals_sepolia.json`. So for any `$1` other than `mainnet`, the `-f` test is false and the whole cross-check is skipped with no output at all.

Not reachable today — a sepolia run already exits at the `globalsL2` check on line 18 — so this is a note rather than a defect. But this block exists specifically to stop an unrecoverable `l1Dispenser` binding, and a guard whose failure mode is "quietly does nothing" is the wrong shape for that job. An `else` branch that warns loudly, or a hard failure when the file is missing, would keep it honest as the script picks up more networks.

Minor, and inherited rather than introduced: the `.dispenserProxyAddress // .dispenserAddress` fallback on line 49 resolves via the second key today, since `scripts/deployment/globals_mainnet.json` has no `dispenserProxyAddress`. It happens to be correct because both hold the same value.

OjusWiZard (dismissed):
Approving. Second pass over this PR, this time with the review agents I should have run the first time — the earlier round was a manual read and it missed things.

**Re-verified as fixed:** `robinhoodOLASAddress` removed entirely (closes both the casing and the dead-key point), the L2 `olasAddress` checksum, and the `deploy_10.js` hard-fail consequences now documented in the header. Constructor argument order still matches both contracts exactly — `ArbitrumDepositProcessorL1(_olas, _l1Dispenser, _l1TokenRelayer, _l1MessageRelayer, _l2TargetChainId, _l1ERC20Gateway, _outbox, _bridge)` and `ArbitrumTargetDispenserL2(_olas, _proxyFactory, _l2MessageRelayer, _l1DepositProcessor, _l1SourceChainId)` — with `_olas` correctly L1 on the processor and L2-bridged on the dispenser.

**The `exit 0` -> `exit 1` change is the most valuable thing here** and I want to flag that I missed it first time round. The originals reported success on "globals not found" and "contract not deployed", so a chained run never stopped on a real failure. Your commit message calls the divergence deliberate and promises a backport to the five siblings — worth actually scheduling, because the tree now carries two conventions for the same failure mode and the next chain script gets templated off whichever sibling someone opens first.

The new Dispenser-binding preflight checks out: paths resolve correctly for `$1=mainnet`, both globals hold `0x5650300f...` so there is no false positive, and `deploy_07b_dispenser_proxy.sh` does write `dispenserProxyAddress` into the file the comment names.

Three inline comments below. None of them block merge — the first is the one I would most want addressed.

scripts/deployment/staking/globals_mainnet.json
dagacha: Follow-up review at `e054217`: **partially addressed; P1 remains for the shared configuration.** The new preflight correctly blocks Robinhood when the deployment globals contain a new proxy and the staking globals still contain the legacy Dispenser, including a case-insensitive comparison. However, this `dispenserProxyAddress` also feeds the other processor deployment scripts, which have no equivalent check. I reproduced the migration scenario using isolated copies and mocked Forge/Cast: `deploy_13_robinhood_deposit_processor.sh mainnet` exits 1 before Forge, while `deploy_02_arbitrum_deposit_processor.sh mainnet` exits 0 and calls Forge with the old `0x5650…93f0` as its immutable Dispenser binding. Please synchronize/read the authoritative proxy address for all consumers, or apply the mismatch preflight consistently to the processor deployment routes. Protecting only Robinhood leaves the original cross-chain migration failure possible.

Reviewed with Codex, GPT-6 Astra (`gpt-6-astra`), medium reasoning effort.

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
dagacha: Follow-up review at `e054217`: **addressed for the reported failure cases.** I reran both scripts against a mocked `forge create` that exits 1 without a deployment address; both now exit 1 instead of 0. Missing-globals branches also return 1. Shell syntax checks pass. No transactions were broadcast.

Reviewed with Codex, GPT-6 Astra (`gpt-6-astra`), medium reasoning effort.

dagacha (commented):
Follow-up review performed with **Codex**, using **GPT-6 Astra (`gpt-6-astra`) with medium reasoning effort**, as recorded in this session's metadata.

Reviewed changes from `7407629` to `e054217`. Posted the detailed results as replies in the existing inline threads:

- **P1 partially addressed:** the new Dispenser mismatch preflight protects Robinhood, but the shared legacy `dispenserProxyAddress` remains usable by the other processor deployment scripts. An isolated Arbitrum reproduction still reaches Forge with the stale immutable binding after a proxy deployment.
- **P2 addressed:** both new scripts return exit 1 on a simulated failed Forge deployment without an address.

Also verified the Robinhood preflight rejects mismatched/zero bindings, accepts matching bindings regardless of checksum casing, and the updated scripts pass shell/JavaScript syntax checks. The checksum correction and removal of the unused L1-side Robinhood token key are consistent with the constructor wiring.

No additional defects identified in this follow-up diff. Tests used isolated files and mocked Forge/Cast; no transactions were broadcast or repository files changed.

dagacha (dismissed):

DavidMinarsch:
Correction to my previous answer, @bennyjo: **the 4663 buyback stack is not a follow-up PR — it stays in wave 1 and the globals will be added in a further revision of this PR.**

I'd said "follow-up" and then traced the consequence: `BalanceTrackerBase.drainer` is immutable and reverts on zero, so deferring the buyback stack also defers the two marketplace balance trackers, and the marketplace can't take payments until it lands. That's a bigger cost than the gap it was avoiding, so the decision is reversed.

So this PR still has an **open task**: no `oracles/` or `utils/` Robinhood globals, which is exactly what your inline comment identified. To close it, a further revision here needs:

- `Bridge2BurnerArbitrum` inputs — the L2 OLAS, the L2 Gateway Router and the L1 OLAS.
- `BuyBackBurnerUniswap` + the **current** `BuyBackBurnerProxy` (not `V1`, not either legacy revision — see the fleet drift in the plan's §1b). Constructor is `(_liquidityManager, _bridge2Burner, _treasury, _swapRouter)`; `_treasury` is the bridge mediator, and `_liquidityManager`/`_swapRouter` are zero on purpose to disable the V3 path, since PoL is out of scope for 4663.
- `UniswapPriceOracle` — and its parameters are still undecided: TWAP window, update interval, staleness threshold. Celo runs 900 / 900 / 86400.

Two ordering constraints the globals need to respect, both verified against the contracts:

1. **`Bridge2BurnerArbitrum` before `BuyBackBurner`** — the latter reverts on a zero `_bridge2Burner` or `_treasury`.
2. **The whole of §3.6 before the marketplace trackers** — `drainerAddress` in marketplace#195 is the 4663 `BuyBackBurnerProxy`, and it's immutable once set.

The rollout plan is updated to match (valory-xyz/autonolas-analytics#254 rev 14): §3.6 runs before §3.5, and this gap is open question 1 with a "not ready" row in the readiness table. Your question is what surfaced the ordering problem, so thanks — it would have been found at deploy time otherwise, after the trackers had already been built.

DavidMinarsch:
All four addressed in `1dfd265`. Thanks both — and @dagacha, you were right that the first fix was the wrong shape.

**P1 reopened and now closed properly.** Protecting only Robinhood left the hazard intact for the eight other scripts reading the same key, and your reproduction proved it. The guard is now `_preflight_dispenser.sh`, sourced by **all ten** scripts that touch `dispenserProxyAddress`. Reproduced your scenario against the Arbitrum script:

```
proxy migrated, staking globals stale, deploy_02_arbitrum_deposit_processor.sh mainnet
  -> "Dispenser binding mismatch — refusing to bind to a stale address"
     exit 1        (was: exit 0, and Forge called with 0x5650…93f0)
synchronised pair -> "Dispenser binding cross-checked", proceeds
```

**@OjusWiZard, the silent no-op.** Now loud when the deployment-side globals are missing. I made it warn rather than abort: not every network has that file, and hard-failing would regress runs that work today. But your framing was the right one — a guard against an unrecoverable binding should never be quiet about not running.

**Unvalidated L2 constructor args.** `deploy_02_robinhood_target_dispenser.sh` now rejects empty, `"null"` and zero for all five inputs before building the args. Your word-splitting analysis was exactly right, and I verified it against the shipped globals where `serviceStakingFactoryAddress` is still empty — exits 1 naming the key. The point that matters is the one you identified: arity catches an *empty* value, but a stale-but-well-formed address would deploy cleanly against a wrong immutable binding.

**Submit-without-await.** `deploy_10.js` now awaits the receipt, throws on a reverted status, and reads back every route from `mapChainIdDepositProcessors`, mirroring the shell route. Your reasoning for why it's worth closing rather than noting is the whole argument: the file's own header says the call is owner-gated, so `OwnerOnly` is the likeliest failure, it reverts *after* submission, and that is precisely the case that printed a hash and exited 0.

**On the backport you flagged** — agreed the tree now carries two conventions for the same failure mode, and templating off the wrong sibling is a real risk. The `exit 0` → `exit 1` change is still only on the two new scripts. I'd rather do that as its own PR than widen this one further; it touches every chain's deploy path and deserves to be reviewed on its own terms. Happy to raise it now if you'd prefer it not to wait.

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
```diff
@@ -0,0 +1,138 @@
...
+robinhoodL2TargetChainId=$(jq -r '.robinhoodL2TargetChainId' $globals)
+robinhoodL1ERC20GatewayAddress=$(jq -r '.robinhoodL1ERC20GatewayAddress' $globals)
+robinhoodOutboxAddress=$(jq -r '.robinhoodOutboxAddress' $globals)
+robinhoodBridgeAddress=$(jq -r '.robinhoodBridgeAddress' $globals)
+
+# Preflight: cross-check the Dispenser binding (see _preflight_dispenser.sh).
+. "$(dirname "$0")/_preflight_dispenser.sh"
+
+robinhoodL1ERC20GatewayRouterAddress=$(jq -r '.robinhoodL1ERC20GatewayRouterAddress' $globals)
+robinhoodInboxAddress=$(jq -r '.robinhoodInboxAddress' $globals)
```
OjusWiZard, line 41: **Leftover from the extraction — lines 40-68 should go.**

The shared helper is sourced at line 38, but the code it replaced is still here:

- **41-45** re-read the same six `robinhood*` values already read at 30-35 (`robinhoodInboxAddress` at 31 and again at 41, `robinhoodBridgeAddress` at 35 and again at 45).
- **51-68** are a second, hand-written copy of the zero-address check and the `deploymentGlobals` cross-check the helper just ran.

**This is inert today, not dangerous.** The sourced helper runs first and `exit 1`s on every condition the two share, so the inline block is only reachable when the helper has already passed with the same inputs — it cannot produce a different verdict.

The reason to remove it anyway is the point of the refactor. `_preflight_dispenser.sh` was extracted to become the single source of truth after review found the Robinhood-only guard wasn't fleet-wide. Leaving the Robinhood-only copy in place means a future tightening of the helper — extra checks, changing the missing-file behaviour, adding format validation — silently doesn't reach whatever this block still does on its own. That is the same "one script forgotten" shape this PR was written to close for the other eight.

scripts/deployment/staking/deploy_10_set_deposit_processors.sh
```diff
@@ -50,6 +50,9 @@ chainId=$(jq -r '.chainId' $globals)
 networkURL=$(jq -r '.networkURL' $globals)
 dispenserProxyAddress=$(jq -r '.dispenserProxyAddress' $globals)
 
+# Preflight: cross-check the Dispenser binding (see _preflight_dispenser.sh).
+. "$(dirname "$0")/_preflight_dispenser.sh"
+
 zeroAddress="0x0000000000000000000000000000000000000000"
```
OjusWiZard, line 56: **Minor: this check is now subsumed by the helper sourced at line 53.**

Lines 56-62 test `dispenserProxyAddress` for empty / `"null"` / zero — exactly what `_preflight_dispenser.sh` lines 14-18 do, and it runs first. Pre-existing rather than introduced here, and only five lines, so much smaller than the `deploy_13` case.

Same drift argument at lower stakes: two copies of one rule means changing the helper leaves this one behind. Worth deleting while the refactor is fresh, or leaving deliberately — either is fine, but it's cheaper to decide now than to discover the divergence later.

OjusWiZard (approved):
Re-reviewed at `1dfd265`. All three of my earlier comments are properly closed, and I checked the refactor rather than the commit message.

**The highest-risk thing about this change is fine.** Sourcing a helper that calls `exit 1` into nine previously-working scripts could have aborted runs that succeed today, so I re-derived the ordering at every call site instead of sampling:

| script | `globals` | `dispenserProxyAddress` | source line |
|---|---|---|---|
| `deploy_02_arbitrum` … `deploy_07_base`, `deploy_11_mode` | 8 | 29 | 31 |
| `deploy_08_eth` | 8 | 22 | 24 |
| `deploy_10_set_deposit_processors.sh` | 38 | 51 | 53 |
| `deploy_13_robinhood` | 8 | 29 | 37 |

Both variables are assigned before the source in all ten. No regression.

Also verified: `deploy_10.js` now does `await result.wait()`, throws on `receipt.status !== 1`, and the read-back loop covers every registered row — with `main().catch()` exiting 1, so a revert can no longer report success. And `robinhood/deploy_02` correctly does **not** source the helper (it never reads `dispenserProxyAddress`); its own five-argument check is the right substitute and runs before `constructorArgs` is built.

**One thing I want to retract before it wastes your time.** I nearly raised that the cross-check silently degrades to a warning on sepolia, since `scripts/deployment/globals_sepolia.json` doesn't exist. It doesn't degrade: `staking/globals_sepolia.json` has no `dispenserProxyAddress` (only `dispenserAddress`), so `jq` yields `"null"` and the helper's *first* check exits 1 before the missing-file branch is reached. That branch is unreachable in every current configuration — it only becomes live for a future network that has the staking key but no deployment-side file. Worth knowing so nobody "fixes" a path that never runs.

Two inline comments, both maintainability rather than correctness — neither blocks merge.

Two smaller notes that aren't this PR's job:

- `"$(dirname "$0")/../globals_${1}.json"` assumes every caller sits directly in `staking/`. True for all ten today, but there are already nine sibling subdirectories (`arbitrum, base, celo, gnosis, mode, optimism, polygon, robinhood, wormhole`) where a future L1-binding script copying this pattern would resolve one level too shallow and cross-check the wrong file silently. Worth one line in the helper's header, which doesn't currently state the assumption.
- The `exit 0` → `exit 1` backport promised in the earlier commit hasn't landed — `deploy_02_arbitrum_deposit_processor.sh` still exits 0 on both missing-globals paths, so the two conventions are still live across the tree.

scripts/deployment/staking/_preflight_dispenser.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+# Shared preflight for every script that binds or registers against the L1 Dispenser.
```
bennyjo, line 2: The Hardhat routes `deploy_02..08` and `deploy_11` `.js` still pass `parsedData.dispenserProxyAddress` into the immutable `_l1Dispenser` with no check, so the stale-binding hazard stays open through `npx hardhat run`; either add the same check there or state that the `.sh` route is the only supported one.

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
```diff
@@ -0,0 +1,138 @@
...
+if [[ "$networkURL" == *"alchemy.com"* ]]; then
+  case $chainId in
+    1)        API_KEY=$ALCHEMY_API_KEY_MAINNET; keyName="ALCHEMY_API_KEY_MAINNET" ;;
+    11155111) API_KEY=$ALCHEMY_API_KEY_SEPOLIA; keyName="ALCHEMY_API_KEY_SEPOLIA" ;;
+    137)      API_KEY=$ALCHEMY_API_KEY_MATIC;   keyName="ALCHEMY_API_KEY_MATIC" ;;
+    80002)    API_KEY=$ALCHEMY_API_KEY_AMOY;    keyName="ALCHEMY_API_KEY_AMOY" ;;
+  esac
+  if [ -n "$keyName" ] && [ "$API_KEY" == "" ]; then
+    echo "set $keyName env variable"
+    exit 0
```
bennyjo, line 81: This missing-Alchemy-key branch still exits 0 with nothing deployed, and the mainnet `networkURL` is an Alchemy URL, so an unset key reads as success in a chained run; `deploy_10.sh:118` uses `exit 1` here, and `robinhood/deploy_02` line 62 has the same branch.

bennyjo (commented):
Follow-up verified: reproduced the preflight in every scenario, including a stale binding on the Arbitrum script, and each guard behaves as described. Two one-line leftovers below; both can ride the buyback-globals revision, which is what I am waiting for before approving.

PR #358, opened 2026-09-08
fix(staking): one Dispenser key, and revert the preflight over-reach from #357

DavidMinarsch:
Fix-forward on #357, which was merged before this came to light. **Net −143/+45.** Two problems, and I introduced the second one.

## What is actually deployed — verified on-chain, not assumed

| Check | Result |
|---|---|
| `0x5650300f…93f0` EIP-1967 impl slot | **zero** — it is a plain Dispenser, **not a proxy** |
| Its code size | 18,684 bytes |
| `l1Dispenser()` on all seven L1 processors | all return `0x5650300f…93f0` |
| `DispenserProxy` on-chain / in `configuration.json` | **none** |

## The root cause, which predates #357

Commit `9bfd58f` renamed `dispenserAddress` → `dispenserProxyAddress` across the script **readers**, in anticipation of the Dispenser proxy migration — but never renamed the **definitions**. Result: `dispenserProxyAddress` was defined in **zero** globals files while `dispenserAddress` was defined in **five**.

That incomplete rename is what #357 recorded as D1. It was never a missing key; it was a key renamed on one side only.

## Where #357 went wrong

It fixed D1 by **adding** `dispenserProxyAddress` alongside `dispenserAddress`, both holding the same value. That left two keys for one value, under a name describing a contract that does not exist — and the duplication then had to be defended by a cross-file staleness preflight sourced into **ten scripts**, nine of which have nothing to do with Robinhood.

This PR reverts to **one key: `dispenserAddress`** — matching deployed reality, already defined in both `globals_mainnet.json` and `globals_sepolia.json`. Key coverage re-checked mechanically: **zero missing keys** across every staking deploy script on mainnet.

## The preflight is removed

Including its ten call sites and a leftover inline copy in `deploy_13` that the rename had quietly degraded into a no-op (`.dispenserAddress // .dispenserAddress`).

Its purpose was to catch a stale binding during a proxy migration. But per `docs/dispenser_migration_runbook_public.md`, that migration is **a full cascade by design**: redeploying the Dispenser forces redeploying every L1 processor and every L2 target dispenser, because those bindings are `immutable`. A cross-file staleness check inside a procedure that replaces all of them belongs in the migration PR, if anywhere.

@dagacha — this supersedes the P1 thread rather than closing it as you'd framed it. You were right that protecting only Robinhood was the wrong shape; the conclusion I've reached is that the shared-key duplication I added was itself the defect, so removing it removes the hazard rather than guarding it.

## Kept from #357

Robinhood globals and scripts; the EIP-55 checksum; `exit 0` → `exit 1` on the two new scripts; the L2 constructor-argument validation; and `deploy_10`'s receipt-await and read-back. All either Robinhood-scoped or independently correct.

## One question for the migration owner — deliberately not changed here

Five files **outside** `scripts/deployment/staking/` still read `dispenserProxyAddress`, and `scripts/deployment/globals_mainnet.json` does not define it:

| file | assessment |
|---|---|
| `deploy_07b_dispenser_proxy.sh` | **legitimate** — it *writes* the key |
| `script_dispenser_change_implementation.sh` | **legitimate** — genuinely means the proxy |
| `deploy_08_09_change_managers.js` | ambiguous — targets the proxy post-migration, the plain Dispenser today |
| `deploy_10_14_change_ownerships.js` | same |
| `script_dispenser_change_managers.sh` | same |

Within a migration run these are self-consistent, since `07b` creates the key before the later scripts read it. Run today, the last three would read `null`. That is the migration's call, not this PR's — flagging rather than widening scope again, which is the mistake that produced #357's diff.


scripts/deployment/staking/globals_mainnet.json
```diff
@@ -59,7 +59,6 @@
   "modeL2TargetChainId": "34443",
   "modeDepositProcessorL1Address": "0x6A5F38aAdCc1c049E39D80147b64a9Ff947d208B",
   "modeTargetDispenserL2Address": "0xEB5638eefE289691EcE01943f768EDBF96258a80",
-  "dispenserProxyAddress": "0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0",
```
OjusWiZard, line 62: **Blocking.** Two files outside `scripts/deployment/staking/` read `dispenserProxyAddress` **from this file**, and neither appears in the PR description's table:

- `scripts/proposals/proposal_10_dispenser_set_deposit_processor_chain_ids.js:7,15,25,29` — reads this exact path, then `getContractAt("Dispenser", undefined)` and `targets = [undefined]`. It is the governance twin of `deploy_10_set_deposit_processors`, which this PR did rename in both `.js` and `.sh`.
- `scripts/audit_chains/audit_contracts_setup.js:457` (and `:515`) — `checkDepositProcessorL1` asserts `l1Dispenser() === globalsInstance["dispenserProxyAddress"]` and is called with this file for all eight L1 processors (lines 1289–1310). `customExpect` catches the mismatch and only `console.log`s it — no throw, no non-zero exit. I ran its exact body against the post-PR globals: it prints `AssertionError: expected 0x5650300f… to equal undefined` and the process exits 0. The audit that exists to catch a wrong Dispenser binding silently stops checking it.

History of this key in this file: absent at `9ca6c342` (pre-#357) → present on `main` → absent here. So this is a regression against the base branch rather than a return to the status quo.

Fix is one line at each site, in the same direction as the rest of the PR: `dispenserProxyAddress` → `dispenserAddress`.

OjusWiZard (commented):
One blocking finding, inline on `scripts/deployment/staking/globals_mainnet.json`.

The diagnosis and its evidence hold up — I re-verified independently rather than taking the description's word: EIP-1967 impl slot at `0x5650300f…93f0` is zero, code is 18,684 bytes, and `l1Dispenser()` returns that address on all seven `DefaultDepositProcessorL1` processors (the eighth, `EthereumDepositProcessor`, reverts on that getter and returns the same address from `dispenser()` — so "all seven" is precisely right). The zero-vs-five globals census is exact at `9ca6c342`: 43 files, 5 defining `dispenserAddress`, 0 defining `dispenserProxyAddress`. Collapsing to one key is the right call, and the rename inside `scripts/deployment/staking/` is complete.

The blocker is scope, not direction: the "five files outside `scripts/deployment/staking/`" table misses four readers (`audit_chains/audit_contracts_setup.js`, `proposals/proposal_10_…js`, `proposals/proposal_09_…js`, `fork/tokenomics_retain_calculation.js`), and two of those read the staking globals file this PR edits.

Non-blocking observations I have left off this review: the null/zero guard lost in nine scripts and its interaction with the pre-existing `exit 0` on the "contract was not deployed" branch, and the surviving `DispenserProxy` wording in `deploy_10_set_deposit_processors.sh`. Happy to post those separately.

DavidMinarsch:
Blocking finding fixed in `45338d0`, plus your two non-blocking ones in `646051c`.

**You were right, and it was worse than an incomplete census — I introduced a regression that silences a guard.**

`audit_contracts_setup.js` asserts `l1Dispenser() === globalsInstance["dispenserProxyAddress"]` for all eight L1 processors, passing `globalsStaking`. I removed that key from that file. I confirmed the effect by reading the key out of both branches rather than trusting the description:

```
main (today)  globalsStaking['dispenserProxyAddress'] = '0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0'  -> PASSES
my PR #358    globalsStaking['dispenserProxyAddress'] = None                                          -> MISMATCH
```

and `customExpect` catches, `console.log`s, and does not throw — so the process exits 0 either way. **The audit that exists to catch a wrong Dispenser binding would have stopped checking it, silently.** That is precisely the failure mode this whole thread has been about, and I built it while removing a guard against the same thing. Your framing — regression against base, not a return to the status quo — is the part I had wrong.

`proposal_10_…js` likewise: `getContractAt("Dispenser", undefined)` and `targets = [undefined]`.

Both now read `dispenserAddress`; the audit assertion resolves and passes.

**I also renamed the other two you listed** — `proposal_09_change_dispenser_and_disable_previous_one.js` and `fork/tokenomics_retain_calculation.js`. They read a copied `globals.json`, and since every globals file defines `dispenserAddress` and none defines the proxy key, they mean the live Dispenser too.

That leaves a rule the tree can be checked against, which is what was missing all along:

> **`dispenserProxyAddress` exists only as `deploy_07b_dispenser_proxy.sh`'s output, and means the proxy specifically.**

The five remaining references are that script plus four that operate on a proxy once one exists — so they stay, and they are already correct if the migration lands.

**Your two non-blocking points, both taken:**

- The `deploy_10` header called the target "the live DispenserProxy" five times and told the operator to run it "after the DispenserProxy is live". Corrected — same anticipatory-proxy confusion that produced the split key.
- While there I found its **ownership note was backwards for the live chain**: it said the owner is the deploying EOA and only becomes the Timelock later. True of a fresh deployment; on mainnet `Dispenser.owner()` is already the Timelock, so the call reverts `OwnerOnly` and belongs in a proposal. Now says which situation is which.

On the null/zero guard lost in nine scripts — please do post it. My reasoning was that the guard existed to defend a duplication that no longer exists, but if it has standalone value against an unset key I would rather hear the argument than assume.

scripts/audit_chains/audit_contracts_setup.js
```diff
@@ -294,7 +294,7 @@ async function checkTokenomicsProxy(chainId, provider, globalsInstance, configCo
 
     // Check dispenser
     const dispenser = await tokenomics.dispenser();
-    customExpect(dispenser, globalsInstance["dispenserProxyAddress"], log + ", function: dispenser()");
+    customExpect(dispenser, globalsInstance["dispenserAddress"], log + ", function: dispenser()");
```
dagacha: **[P2] Resolve the live Dispenser separately from its implementation in the parent globals**

The earlier review correctly identified that the processor checks receive `globalsStaking`, whose removed proxy key needed replacing. This Tokenomics check (and the Treasury check at line 338) instead receives `globals[0]`, loaded from `scripts/deployment/globals_mainnet.json`. Those files have different semantics once the existing deployment scripts run: `deploy_07a_dispenser.sh` writes the standalone implementation to `dispenserAddress`, and `deploy_07b_dispenser_proxy.sh` writes the live proxy to `dispenserProxyAddress`. The unchanged manager-update scripts then wire Tokenomics/Treasury to that proxy.

Consequently, after that deployment sequence these changed assertions flag correct proxy wiring as a mismatch and accept incorrect wiring to the implementation. I reproduced the comparison with distinct implementation and proxy values: the base accepts the proxy; this revision accepts the implementation. This is separate from the already-fixed missing key in the staking globals and does not require retaining duplicate keys for the currently deployed plain Dispenser.

Please resolve the live address from a configured, nonzero `dispenserProxyAddress`, falling back to `dispenserAddress` for the current non-proxied deployment, in the Tokenomics and Treasury checks. That preserves both current-chain auditing and the existing proxy deployment workflow.

dagacha (commented):
Review performed with Codex, using GPT-6 Astra (`gpt-6-astra`), reasoning effort: Medium.

Reviewed `646051c` and the existing reviews/discussion on #358 and its predecessor #357. One additional finding is attached inline. The previously raised stale-binding/preflight concern and null/zero validation concern are not repeated here. The missing staking-globals readers reported in the earlier #358 review have been updated in this revision.

Validation: syntax checks passed for all 22 changed JavaScript files and 10 changed shell scripts; checked the audit expectations against the existing implementation/proxy deployment writers and reproduced the comparison behavior with distinct implementation/proxy addresses. Live RPC verification using the configured endpoint failed, so this review does not independently confirm the on-chain claims. No transactions were broadcast.

OjusWiZard (dismissed):
Approved.

DavidMinarsch:
Fixed in `d7ea3f9`. @dagacha — you're right, and this one was created by my own rename rather than inherited.

The blanket `sed` treated all eight dispenser assertions alike, but they don't read the same globals:

| Lines | Check | Receives | Correct key |
|---|---|---|---|
| L1 deposit processors | `l1Dispenser()` / `dispenser()` | `globalsStaking` | `dispenserAddress` — the rename was right |
| Tokenomics, Treasury | `dispenser()` | `globals[0]` = `scripts/deployment/globals_<network>.json` | **the proxy, once one exists** |

In that parent file the two keys genuinely mean different contracts — `deploy_07a` writes the implementation to `dispenserAddress`, `deploy_07b` writes the live proxy to `dispenserProxyAddress`, and the manager scripts wire Tokenomics/Treasury to the proxy. So my version would have **flagged correct proxy wiring as a mismatch and accepted wrong wiring to the implementation** — the precise inversion of the check's purpose.

Those two now resolve through a `liveDispenser()` helper: prefer a configured, non-zero `dispenserProxyAddress`, fall back to `dispenserAddress`. Verified across the four cases rather than the one:

```
no proxy key (today)                -> implementation
both keys, distinct (post-migration) -> proxy
proxy key present but zero           -> implementation
both keys equal (mainnet today)      -> that address
```

The helper carries an explicit note **not** to use it for the deposit-processor checks: those read the staking globals, which has only the one key, and `l1Dispenser` is an immutable binding to whichever Dispenser was live at deploy time. That is the distinction the original blanket rename lost, so it seemed worth writing down rather than leaving to the next reader.

Also worth recording, since it is the through-line of this whole PR: `dispenserAddress` and `dispenserProxyAddress` mean the same thing in the *staking* globals (one Dispenser, no proxy) and different things in the *deployment* globals (implementation vs proxy). Collapsing them everywhere was too blunt; the rule now is per-file, and the helper encodes it.

@OjusWiZard — your null/zero guard argument would still be welcome, but I won't hold the PR for it.

scripts/audit_chains/audit_contracts_setup.js
dagacha: Follow-up review at `d7ea3f98817e60f25d000eeaee8324faa6979bb3`: **addressed**. I executed the new `liveDispenser()` helper from this commit against missing, null, empty, zero, distinct-proxy, and equal-address configurations; all six select the expected address. Confirmed that both Tokenomics and Treasury checks call the helper, while the bridge and Ethereum processor checks continue to compare against the staking globals' `dispenserAddress`. JavaScript syntax and diff whitespace checks pass.

No additional findings in the `646051c..d7ea3f9` follow-up diff. This closes this audit finding; it does not reclassify the separate preflight concern already discussed elsewhere. Validation was local; no transactions were broadcast.

Reviewed with Codex, GPT-6 Astra (`gpt-6-astra`), reasoning effort: Medium.

dagacha (approved):

PR #359, opened 2026-09-09
fix(staking): propagate deploy failures instead of reporting success

DavidMinarsch:
Backport of the exit-code fix from #357/#358, which applied it to the two new Robinhood scripts only — leaving the tree with two conventions for the same failure mode, and the next chain script templated off whichever sibling someone opens first. Raised separately because it touches every chain's deploy path. Requested by @OjusWiZard in the #357 review.

## The exit codes — 21 scripts

Every `exit 0` that follows an error message becomes `exit 1`:

| branch | fires when |
|---|---|
| `$globals` / `$globalsL1` / `$globalsL2` not found | wrong network argument |
| `set $keyName env variable` | missing Alchemy key |
| `The contract was not deployed...` | **`forge create` returned no address** — rejected Ledger signature, RPC error, bad constructor args |

**One `exit 0` is deliberately kept:** `multi_deploy`'s early return for ethereum, which is a legitimate success — no L2 deployment is needed there.

## The exit codes alone would not have been enough

`multi_deploy_01_processor_dispenser_link.sh` chains its three steps as bare commands — **no `set -e`, no `&&`** — so a non-zero return was discarded and the chain continued anyway.

It now sets `set -e`, and that is load-bearing rather than hygiene. Each step binds the next **by address**, and two of those bindings are `immutable`: `DefaultDepositProcessorL1.l1Dispenser`, and the aliased `l1DepositProcessor` that `DefaultTargetDispenserL2` computes in its constructor. A dispenser deployed against a stale processor address **cannot be corrected afterwards** — it needs both contracts redeployed.

Demonstrated with a stubbed chain where step 1 fails:

```
before   [step 1] simulating forge failure
         !!! The contract was not deployed...
         [step 2] L2 dispenser RAN — would bind to a stale l1DepositProcessor (immutable!)
         [step 3] link RAN
         chain exit=0

after    [step 1] simulating forge failure
         !!! The contract was not deployed...
         chain exit=1
```

No behavioural change on a successful run.

## Why now

This is on the critical path for the Robinhood Chain rollout: §3.4 of the plan (valory-xyz/autonolas-analytics#254) runs exactly this chain, and it is the step where a silent continuation would produce an unrecoverable binding.

### DIFF
--- a/scripts/deployment/staking/arbitrum/deploy_02_arbitrum_target_dispenser.sh
+++ b/scripts/deployment/staking/arbitrum/deploy_02_arbitrum_target_dispenser.sh
@@ -8,14 +8,14 @@ reset=$(tput sgr0)
 globals="$(dirname "$0")/globals_$1.json"
 if [ ! -f $globals ]; then
   echo "${red}!!! $globals is not found${reset}"
-  exit 0
+  exit 1
 fi
 
 # Get globals file for L1: globals_mainnet or globals_sepolia
 globalsL1="$(dirname "$0")/../globals_${1#*_}.json"
 if [ ! -f $globalsL1 ]; then
   echo "${red}!!! $globalsL1 is not found${reset}"
-  exit 0
+  exit 1
 fi
 
 # Read variables using jq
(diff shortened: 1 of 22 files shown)


OjusWiZard (approved):
Approved.

PR #360, opened 2026-09-09
deploy(robinhood): cross-chain incentive route on chain 4663

DavidMinarsch:
§3.4 of the Robinhood rollout: the OLAS staking-incentive route between Ethereum and chain 4663. Two contracts deployed and linked, plus hardening of the deployment script that was already on `main`.

| contract | chain | address |
|---|---|---|
| `ArbitrumDepositProcessorL1` | Ethereum | [`0xb9DfcC6155Ba4F211DCf8e6eCc9976Be11bB7a77`](https://etherscan.io/address/0xb9DfcC6155Ba4F211DCf8e6eCc9976Be11bB7a77) |
| `ArbitrumTargetDispenserL2` | 4663 | `0xc40C79C275F3fA1F3f4c723755C81ED2D53A8D81` |

Linked with `setL2TargetDispenser`, which is one-shot and self-revoking, so the L1 processor is **now ownerless** — `owner() == 0`, matching Arbitrum One's.

## Script hardening

`deploy_13_robinhood_deposit_processor.sh` shipped with the six Robinhood `jq` reads **duplicated verbatim** (lines 30–35 repeated at 37–42) and **no input validation at all**, while its L2 counterpart already validates. `--constructor-args $constructorArgs` is unquoted, so an empty value collapses under word-splitting instead of being passed as an empty token, and the constructor would silently receive seven arguments instead of eight.

Added the **distinctness guard** the rollout plan calls for. Robinhood and Arbitrum One are both Orbit chains sharing `ArbitrumDepositProcessorL1`, and six of eight constructor arguments are chain-specific — carrying an Arbitrum One value over would deploy cleanly and then route OLAS into Arbitrum's escrow tagged 4663. Only `_olas` and `_l1Dispenser` are legitimately shared. Exercised against six fixtures before use:

| fixture | result |
|---|---|
| real globals | pass |
| blanked / zeroed value | exit 1 |
| `arbitrumInboxAddress` carried over | exit 1 |
| `arbitrumL2TargetChainId` carried over | exit 1 |
| `arbitrumBridgeAddress` carried over | exit 1 |

`serviceStakingFactoryAddress` was empty in the L2 globals; set to the wave-1 `StakingFactory`, whose `verifier()` reads back as the wave-1 `StakingVerifier`.

## Cross-checked against every live chain before signing

**Shared arguments** — `olas` `0x0001A500…` and `l1Dispenser` `0x5650300f…` are identical across all seven live L1 deposit processors, and each carries its own `l2TargetChainId`. Robinhood matches that pattern.

**Chain-specific arguments** — all six differ from Arbitrum One's live immutables:

| arg | Arbitrum One | Robinhood |
|---|---|---|
| `l1TokenRelayer` | `0x72Ce9c84…31ef` | `0x6a2E3a1e…75bB` |
| `l1MessageRelayer` | `0x4Dbd4fc5…AB3f` | `0x1A07cc4B…7a2D` |
| `l1ERC20Gateway` | `0xa3A7B6F8…0EeC` | `0x85001CC4…6da0` |
| `outbox` | `0x0B9857ae…4840` | `0xf0ce991e…3DE9` |
| `bridge` | `0x8315177a…ed3a` | `0xDf875533…64b3` |
| `l2TargetChainId` | 42161 | 4663 |

**L1 bridge topology**, read on-chain: `Inbox.bridge() == Bridge`, `Bridge.rollup() == Outbox.rollup() == 0x23A19d23…`, `Bridge.allowedDelayedInboxes(Inbox) == true`, `Router.getGateway(OLAS)` → the standard gateway, and `calculateL2TokenAddress(OLAS)` → `0x0929…2286`, the live L2 OLAS.

**Aliasing** — the `+0x1111…1111 mod 2^160` formula was validated against Arbitrum One's live pair *before* being relied on, and the deployed `l1AliasedDepositProcessor` came out at the predicted `0xcAf0CC6155Ba4F211dcf8E6ECC9976bE11bB8b88`.

**All immutables read back** on both contracts after deployment: 8/8 and 5/5 plus the alias.

## Build provenance

Built from `main` @ `af0eb6b` — solc 0.8.30, viaIR, optimizer 200, prague; hardhat and foundry configs agree.

Each deployed runtime was checked against the artifact by substituting the expected immutables in and requiring an exact match: **zero residual bytes on both**.

> **Correction — thanks @OjusWiZard.** That comparison was on bytecode with the CBOR metadata trailer *stripped*, and the original wording did not say so. The full runtimes do differ from `abis/0.8.30/`, by the 32-byte metadata hash at `len-43` (solc version bytes identical). The logic claim stands; the wording overreached.
>
> This is also the fix for the artifact question: both entries now point at `abis/deployed/RobinhoodDepositProcessorL1.json` and `RobinhoodTargetDispenserL2.json`, whose metadata trailer matches the deployed code **exactly** — so the two Tier-2 drift warnings are gone (4663-section warnings 1 → 0) and the regeneration hazard with them.

⚠️ **This is the first deployment of the 0.8.30 build of this pair** — the fleet runs the older `abis/deployed/` build (6,713 / 8,435 B vs 5,088 / 7,327 B). Only three commits touched the five-file import closure since the 2025-07-14 fleet deployment:

| commit | change |
|---|---|
| `396e0a5` | vuln-list #10 — `Migrated` event gains `withheldAmount` |
| `326379d` | NatSpec only |
| `2b822c4` | pragma `^0.8.30` + ABI refresh, no logic |

So the **logic is unchanged**; the size delta is viaIR. **Function selectors are identical** — 23 and 27, none added or removed — so the binding to the deployed `Dispenser` is unaffected, and the only ABI change is that event. Consequence worth recording: with no byte-identical twin on another chain, Blockscout's cross-chain bytecode database will **not** auto-verify the L2 side the way it did for wave 1. The L1 side verified normally on Etherscan and eth.blockscout.

## The static audit was broken, and now covers 4663

**It never ran.** On `main` it crashed partway through chain 1 — `invalid address or ENS name (value=undefined)` — and reached no L2 chain at all, so every chain has been unaudited. Confirmed against a pristine `origin/main` worktree: identical crash, exit 1. Not introduced by this branch.

`checkLiquidityManagerProxy` looks up `LiquidityManagerProxy` in `configuration.json`, finds nothing, and passes `undefined` as an address. **The proxies are deployed**, so the fix is the missing entries, not disabling the check:

| chain | address | code | `owner()` | slot → impl |
|---|---|---|---|---|
| mainnet | `0x83Adc54B…a215` | 202 B | deployer | matches config |
| optimism | `0x1f84F8F7…3B3A` | 202 B | deployer | matches config |
| base | `0x2ACD313B…fC2a` | 202 B | deployer | matches config |

`abis/0.8.30/LiquidityManagerProxy.json` is a 202-byte length match on all three. Disabling the check would have permanently hidden three live, deployer-owned contracts — a check that silently does not run is worse than one that fails.

**With that repaired the audit completes all chains and finds real drift**: `LiquidityManagerProxy.maxSlippage()` reads **1000** on mainnet, optimism and base while `scripts/deployment/pol/globals_*.json` record **500**. Not touched here — whether it was deployed at 1000 or changed later is a separate question. Note the process still **exits 0** despite this, since `customExpect` logs without throwing and only bytecode mismatches set the blocking flag.

**4663 registered.** The L2 dispenser entry, the Robinhood `ArbitrumDepositProcessorL1` as the *second* entry of that name under `mainnet`, and the three hardcoded maps (`globalNames`, `providerLinks`, `chainSlug`). Robinhood is appended last because the L2 dispatch walks `configs` in order via `chainNumber++`.

Robinhood reuses the Arbitrum pair with its own bridge addresses — the same shape as Base/Celo/Mode reusing the Optimism pair — so it gets the same per-chain wrapper idiom. `checkRobinhoodDepositProcessorL1` passes **`idx 1`** to `findContractInstance` and `checkBytecode`; without it the check would silently re-verify Arbitrum One's address and a divergence on 4663 would be invisible — precisely the defect the `idx` parameter was added for.

Result on 4663: the L1 processor **passes every check**, including all six Robinhood-specific bridge addresses and `l2TargetDispenser`. The one failure is the L2 dispenser's `owner()` still being the deployer, which is item 1 below and is recorded in `ownable_owners.csv` as `ownershipChangeRequired: yes`.

## Outstanding — and the order matters

⚠️ **Item 1 must land before item 2.** While `owner()` is the deployer EOA, that key can `pause()` then `migrate()` the dispenser's entire OLAS balance to any contract. Exposure is zero *today* only because the dispenser holds 0 OLAS and `mapChainIdDepositProcessors(4663)` is still zero — that safety is incidental to item 2 not having happened, not a property of the design. Registering the processor with governance before ownership moves would open a window in which an EOA can drain incentives. Flagged by @OjusWiZard; the plan's §3.7 is being updated to say so.

1. **L2 dispenser ownership** — `owner()` is still the deployer and must move to the bridge mediator `0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F`, as on Arbitrum One (`script_04_target_dispenser_l2_change_owner.sh`).
2. **Dispenser registration is governance** — `Dispenser.owner()` is the Timelock, so `mapChainIdDepositProcessors(4663)` is still zero and `deploy_10_set_deposit_processors` belongs to wave 2. For comparison, 42161 reads `0xFceFB015…2e67`.
3. Manual verification of the L2 dispenser on Robinhood's Blockscout.

Wave 1: valory-xyz/autonolas-registries#325.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
```diff
@@ -34,12 +34,38 @@ robinhoodL1ERC20GatewayAddress=$(jq -r '.robinhoodL1ERC20GatewayAddress' $global
...
+done
+
+# Robinhood and Arbitrum One are both Arbitrum Orbit chains sharing ArbitrumDepositProcessorL1, and six of
+# the eight constructor arguments are chain-specific. Carrying an Arbitrum One value over would deploy
+# successfully and then route OLAS into Arbitrum's escrow under a 4663 chain tag - a silent, unrecoverable
+# mis-send. Only _olas and _l1Dispenser are legitimately shared.
+for key in L1ERC20GatewayRouterAddress InboxAddress L2TargetChainId L1ERC20GatewayAddress OutboxAddress BridgeAddress; do
+  rhVal=$(jq -r ".robinhood${key}" $globals)
+  arbVal=$(jq -r ".arbitrum${key}" $globals)
+  if [ "$arbVal" != "null" ] && [ "$rhVal" == "$arbVal" ]; then
```
dagacha: [P2] Normalize addresses before the distinctness comparison

This compares case-sensitive strings, so an Arbitrum address copied in lowercase bypasses the guard despite identifying the same contract. I reproduced this for all five bridge-address fields: exact copies fail, while lowercase copies pass validation. For example, setting robinhoodInboxAddress to 0x4dbd4fc535ac27206064b68ffcf827b0a60bab3f passes even though it is the configured Arbitrum One inbox. This allows the wrong-chain configuration the guard is intended to prevent. Normalize both address values before comparing and add a lowercase-copy fixture; keep the chain-ID comparison separate.

dagacha (commented):
Review performed with Codex using GPT-6, reasoning effort: Medium.

Found one issue, documented inline. Shell and JavaScript syntax checks passed, all configured artifacts resolve, and the Robinhood addresses agree across configuration files. The deployment guard was exercised with exact-copy and lowercase-copy fixtures. The full live-chain audit was not run.

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
OjusWiZard: This comparison is exact-string, so the guard is case-sensitive. I defeated it by running this loop verbatim against a copy of `globals_mainnet.json` with Arbitrum One's router pasted into the robinhood slot in lowercase:

```
lowercase 0x72ce9c846789fdb6fc1f34ac4ad25dd9ef7031ef  -> guard fired? 0
checksummed 0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef -> GUARD FIRED
```

Same address, opposite outcome. Etherscan URLs and most CLI output are lowercase, `forge` accepts either, and the constructor would bind the wrong immutable - which is exactly the mis-send this block exists to prevent, arriving through the most likely paste path.

```sh
rhVal=$(jq -r ".robinhood${key}" $globals | tr 'A-F' 'a-f')
arbVal=$(jq -r ".arbitrum${key}" $globals | tr 'A-F' 'a-f')
```

While here, two smaller things in this block:

- The `[ "$arbVal" != "null" ]` escape means a missing `arbitrum*` key produces no output and the deploy proceeds. All six exist in `globals_mainnet.json` today, so this is latent, but a guard that cannot report its own inapplicability is worth one `echo`.
- The comment at :37-39 says an empty value means the constructor "would **silently** receive seven arguments instead of eight." It would not - `forge create` rejects the arity. Your own sibling gets this right at `robinhood/deploy_02_robinhood_target_dispenser.sh:36`: "That fails at forge's arity check today, but a stale-but-well-formed address would deploy successfully." That sibling names the hazard this guard actually addresses; worth aligning the wording.

docs/configuration.json
```diff
@@ -58,6 +58,11 @@
         "artifact": "abis/deployed/ArbitrumDepositProcessorL1.json",
         "address": "0xFceFB015372e84FC465Cf2778889ee231a6E2e67"
       },
+      {
+        "name": "ArbitrumDepositProcessorL1",
+        "artifact": "abis/0.8.30/ArbitrumDepositProcessorL1.json",
```
OjusWiZard: This entry, and the L2 one at :355, point at `abis/0.8.30/` - which `abis/README.md` reserves for code that is not deployed yet:

> `docs/configuration.json` describes **what is deployed** [...] So an artifact under a name that several deployments share cannot be regenerated in place for whichever one was built most recently - doing that silently breaks the comparison for all the others.

As of this PR, `ArbitrumDepositProcessorL1` and `ArbitrumTargetDispenserL2` are both shared names with two deployments and two distinct builds:

| name | deployment | artifact |
|---|---|---|
| ArbitrumDepositProcessorL1 | Arbitrum One | `abis/deployed/` |
| ArbitrumDepositProcessorL1 | Robinhood | `abis/0.8.30/` |
| ArbitrumTargetDispenserL2 | Arbitrum One | `abis/deployed/` |
| ArbitrumTargetDispenserL2 | Robinhood | `abis/0.8.30/` |

That is precisely the case `abis/deployed/` was introduced for, and the standing-red note at `scripts/audit_chains/audit_contracts_setup.js:185-215` records 22 Tier-1 failures caused by exactly this - "every `chore: updating ABIs` regenerated that file for the newest deployment and broke the comparison for the rest."

Failure scenario: the next ABI regeneration rewrites `abis/0.8.30/ArbitrumDepositProcessorL1.json` at whatever settings the repo then uses. If the length moves, `checkBytecode` sets `bytecodeMismatchFound` and the audit exits 1 - the regression this directory exists to prevent, reintroduced by the entry that repairs the audit.

The metadata-trailer delta in the review body is the same point from the other direction: these artifacts already are not the deployed builds. Copying both deployed runtimes into `abis/deployed/` under per-deployment names and repointing these two entries fixes the drift warnings and the regeneration hazard together. The README paragraph enumerating which names are shared needs the two Arbitrum names added as well.

OjusWiZard (dismissed):
Approving. I verified the deployment on-chain rather than the checklist: 8/8 immutables on the L1 processor, 6/6 on the L2 dispenser, `owner() == 0` on L1, and `l2TargetDispenser` linked.

One check worth adding to your evidence, because it is the one that actually binds this route to Robinhood rather than merely to "not Arbitrum One":

```
Bridge.rollup() == Outbox.rollup() == 0x23A19d23...
rollup.chainId()                    -> 4663
Gateway.calculateL2TokenAddress(OLAS) -> 0x092963938deBD8013a2e545b3549f8A5ec0D2286
  and that token's l1Address()        -> 0x0001A500... (L1 OLAS)
```

The six-address distinctness table proves the values differ from Arbitrum One's; `rollup.chainId()` proves they belong to 4663. Also confirmed independently: `stakingFactory` and its `verifier()` match registries' wave-1 globals exactly, and the three `LiquidityManagerProxy` repairs are live with implementation slots matching the entries you added. The `maxSlippage` drift is real - 1000 on-chain against `liquidityManagerMaxSlippage: "500"` on mainnet, optimism and base.

Two inline comments. Neither blocks; both are worth landing before the next Orbit chain reuses this script.

---

**A correction to one claim, and an ordering point.**

"Zero residual bytes on both" is not quite right. Each deployed runtime differs from its `abis/0.8.30/` artifact by a 32-byte run at exactly `len-43` - the CBOR metadata hash:

```
L1 onchain  44b4ec51...c0c0 64736f6c634300081e0033
L1 artifact 6bda4534...0139 64736f6c634300081e0033
L2 onchain  e4fa8831...facf 64736f6c634300081e0033
L2 artifact 5fc0e062...5e84 64736f6c634300081e0033
```

solc version bytes identical, every other differing byte is an immutable - so the logic claim holds. But the artifacts are not the deployed builds, which means two standing Tier-2 `metadata-trailer drift` WARN lines on every audit run from now on. That is also the evidence for the second inline comment.

Separately, the two Outstanding items are order-dependent and the list does not say so. While `owner()` is the deployer EOA, that key can `pause()` then `migrate()` the entire OLAS balance to any contract. Exposure is zero today - the dispenser holds 0 OLAS and `mapChainIdDepositProcessors(4663)` is zero - but that safety is incidental to item 2 not having happened yet. Item 1 (ownership to the bridge mediator) should land before item 2 (governance registration), or there is a window where an EOA can drain incentives.

scripts/deployment/staking/deploy_13_robinhood_deposit_processor.sh
DavidMinarsch: Fixed in `4fb4315`, and you were right that it fails on the most likely path.

I reproduced it before changing anything — `robinhoodInboxAddress` set to Arbitrum One's inbox in lowercase passed the guard, the same address checksummed was blocked. Explorers and CLI output render addresses unchecksummed, so the guard was weakest exactly where a real mistake would arrive.

Both sides are now lowercased before comparison. Fixtures, all verified:

```
A) real globals                     exit=0
B) arbitrum inbox, exact copy       exit=1
C) arbitrum inbox, LOWERCASE copy   exit=1   <- was exit=0
D) arbitrum bridge, LOWERCASE copy  exit=1
E) arbitrum router, LOWERCASE copy  exit=1
F) arbitrum key missing             exit=1
```

Also took the other two points. A missing `arbitrum*` key is now **fatal** rather than a silent skip — a guard that cannot run should say so rather than wave the deployment through. And the `:37-39` comment is corrected: `forge` does reject the arity, so the hazard is a stale-but-well-formed address deploying against a wrong immutable, which is what the sibling `robinhood/deploy_02` already said. Aligned the wording to it.

DavidMinarsch: Same fix, `4fb4315` — see the reply on @dagacha's thread for the fixture table. Your lowercase-router case is one of them and now exits 1.

Both smaller points taken as well: the missing-key escape is now fatal rather than a silent skip, and the `:37-39` comment is corrected to match what `robinhood/deploy_02` already said.

docs/configuration.json
DavidMinarsch: Fixed in `4fb4315`. You were right on both the convention and the underlying fact, and the second one I could confirm directly:

```
ArbitrumDepositProcessorL1 metadata trailer
  onchain      44b4ec51...c0c0 64736f6c634300081e0033
  forge build  44b4ec51...c0c0 64736f6c634300081e0033   match
  abis/0.8.30  6bda4534...0139 64736f6c634300081e0033   NO match
```

Same source, same solc, same settings — different metadata hash. So `abis/0.8.30/` genuinely was not the deployed build, and pointing `configuration.json` at it was both a permanent Tier-2 warning and the regeneration hazard you describe.

The forge artifacts *do* match the deployed code's trailer exactly, so those are now recorded as `abis/deployed/RobinhoodDepositProcessorL1.json` and `RobinhoodTargetDispenserL2.json` — following the `CeloDepositProcessorL1.json` naming already in use — and both entries repointed.

Verified after the change: length **and** metadata trailer both match, and the audit's 4663-section warnings go **1 → 0** with the Robinhood L1 processor warning gone. Fleet-wide warnings are unchanged at 41, since only these two artifacts were touched.

`abis/README.md` now names the two Arbitrum Orbit entries among the shared names, and records why identical logic was not sufficient here — that was the part I had wrong, and it is worth having in the file rather than only in a PR thread.

Your review-body points are handled too: the "zero residual bytes" claim is corrected in the description (it was on *stripped* bytecode and did not say so), and the Outstanding list now states that ownership must land **before** governance registration, with the `pause()` + `migrate()` reasoning. That ordering is going into the plan's §3.7 as well, since it belongs there rather than in a PR body.

scripts/audit_chains/ownable_owners.csv
```diff
@@ -20,4 +20,5 @@ chainId,contractName,contractAddress,ownerAddress,ownerCategory,expectedDaoExecu
 8453,BuyBackBurnerProxy,0x3FD8C757dE190bcc82cF69Df3Cd9Ab15bCec1426,0xE49CB081e8d96920C38aA7AB90cb0294ab4Bc8EA,dao_executor,0xE49CB081e8d96920C38aA7AB90cb0294ab4Bc8EA,no
 42220,OptimismTargetDispenserL2,0x4891f5894634DcD6d11644fe8E56756EF2681582,0xC14E191A64a7FB0e5790a8a0B9a58683dFFce04d,dao_executor,0xC14E191A64a7FB0e5790a8a0B9a58683dFFce04d,no
 42220,BuyBackBurnerProxy,0x847bBE8b474e0820215f818858e23F5f5591855A,0xC14E191A64a7FB0e5790a8a0B9a58683dFFce04d,dao_executor,0xC14E191A64a7FB0e5790a8a0B9a58683dFFce04d,no
-34443,OptimismTargetDispenserL2,0xEB5638eefE289691EcE01943f768EDBF96258a80,0x9338b5153AE39BB89f50468E608eD9d764B755fD,dao_executor,0x9338b5153AE39BB89f50468E608eD9d764B755fD,no
\ No newline at end of file
+34443,OptimismTargetDispenserL2,0xEB5638eefE289691EcE01943f768EDBF96258a80,0x9338b5153AE39BB89f50468E608eD9d764B755fD,dao_executor,0x9338b5153AE39BB89f50468E608eD9d764B755fD,no
+4663,ArbitrumTargetDispenserL2,0xc40C79C275F3fA1F3f4c723755C81ED2D53A8D81,0xEB2A22b27C7Ad5eeE424Fd90b376c745E60f914E,autonolas_deployer,0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F,yes
```
dagacha, line 24: Non-blocking rollout prerequisite, already documented in the PR: live RPC checks confirm this dispenser is still owned by the deployer, while mapChainIdDepositProcessors(4663) remains zero. Before governance registers the processor, transfer dispenser ownership to bridge mediator 0x4d30F68F5AA342d296d4deE4bB1Cacca912dA70F and verify owner() on-chain. The current owner can pause and migrate the dispenser balance, so this ordering remains necessary before incentives are enabled. No new code defect is being raised here.

Reviewed with Codex — GPT-6, effort Medium.

dagacha (commented):
Review performed with Codex — LLM model: GPT-6; reasoning effort: Medium.

Reviewed commit 4fb4315b9ee11dda2e10c5e56a77783037d8ca1b. No new actionable bugs found in the changes. One non-blocking inline note records the already documented ownership prerequisite.

Validation completed:
- Bash and JavaScript syntax checks; all configured artifact paths resolve and parse.
- 39 deployment-input fixtures passed, covering valid configuration, empty/null/zero inputs, and copied Arbitrum values including lowercase addresses.
- Both Robinhood artifacts have equivalent ABIs to the existing 0.8.30 artifacts (ignoring entry order), and matching bytecode outside the metadata trailer.
- Read-only RPC checks confirmed both deployed runtimes match the committed deployment artifacts including metadata after substituting the verified immutable values. All immutable getters and embedded values match the expected configuration, including the L1 address alias.
- Confirmed the L1→L2 dispenser link, Inbox/Bridge/Outbox topology, allowed inbox, OLAS gateway and L2 token mapping.
- Confirmed L1 processor owner is zero, L2 dispenser owner remains the deployer, and Dispenser registration for chain 4663 remains zero.

Scope limitation: this was a diff review with focused validation and live deployment checks; I did not run the full multi-chain audit or execute an end-to-end incentive transfer.

dagacha (approved):

No blocking issues. Independent on-chain read-back, not the description.

**L1** [`0xb9DfcC6155Ba4F211DCf8e6eCc9976Be11bB7a77`](https://etherscan.io/address/0xb9DfcC6155Ba4F211DCf8e6eCc9976Be11bB7a77) — codesize 5087, `owner()==0`, 8/8 immutables match `globals_mainnet.json`, metadata trailer matches `abis/deployed/RobinhoodDepositProcessorL1.json`.

**L2** `0xc40C79C275F3fA1F3f4c723755C81ED2D53A8D81` — codesize 7380, trailer matches `RobinhoodTargetDispenserL2.json`. `l1DepositProcessor()` is the unaliased L1 processor; `l1AliasedDepositProcessor()` is `0xb9Df… + 0x1111…1111` (`0xcAf0CC61…8b88`). OLAS balance 0, `paused==1`.

**Topology binds to 4663**, not merely “not Arbitrum One”: `Inbox.bridge() == Bridge`, `Bridge.rollup() == Outbox.rollup() == 0x23A19d23…`, `rollup.chainId() == 4663`, `Router.getGateway(OLAS)` is the configured gateway, `calculateL2TokenAddress(OLAS)` is live L2 OLAS whose `l1Address()` is L1 OLAS.

Factory `0x1BD1505B…` / `verifier() == 0x75D529FA…` match registries #325. `Dispenser.mapChainIdDepositProcessors(4663)` is still zero.

`0x4d30F68F…A70F` is the aliased L1 Timelock (`0x3C1fF68f… + 0x1111…1111`). Codesize 0 on 4663 and on Arbitrum One is expected; that is the correct `changeOwner` target.

Distinctness guard: live globals pass; exact-copy and lowercase Arbitrum copies, and a missing `arbitrum*` key, all exit 1.

Ops already in the body, confirmed: **item 1 before item 2**. Owner is still the deployer. Registering 4663 on Dispenser first would open `pause()` + `migrate()`.

OjusWiZard (approved):
Approving at 4fb4315b. Both review points are fixed and verified.

The distinctness guard now normalises case and fails on a missing key. I ran the loop from this commit against fixtures: lowercase router copy, uppercase-hex outbox copy, copied L2TargetChainId and a missing arbitrum key all exit 1; the real globals exit 0.

Both Robinhood entries now point at `abis/deployed/` artifacts in the same shape as their siblings (`deployedBytecode.object`, which `checkBytecode` reads), and each matches its deployed runtime on length and metadata trailer, with differences only at immutable positions.

dagacha (approved):

PR #361, opened 2026-09-10
deploy(robinhood): buyback stack on chain 4663

DavidMinarsch:
§3.6 of the Robinhood rollout: the buyback stack on chain 4663. **Stacked on #360** — branched from `deploy/robinhood-crosschain`, so review that first; this PR's diff against it is the buyback work alone.

This unblocks §3.5: `BuyBackBurnerProxy` is the address the marketplace binds as its immutable `drainer`.

| contract | address |
|---|---|
| OLAS/WETH UniswapV2 pair | `0xc2eA98b5A75Fd85f7ce57Af856baaFfECD445659` |
| `UniswapPriceOracle` | `0x98677Fd8Ac2BD2AF4BD2BFd4eEB651394362445d` |
| `Bridge2BurnerArbitrum` | `0xe282cDe2718DbE656394AeE36C104776C30a5CB4` |
| `BuyBackBurnerUniswap` (impl) | `0x1fe74A08ac89300B102AdCd474C721AE8764E850` |
| **`BuyBackBurnerProxy`** | **`0x209694516495253E366E6FB0dD0eb1F275354947`** |

## The pair was created and seeded atomically, on purpose

`router.addLiquidityETH` rather than `createPair` followed by a separate `addLiquidity`. A separate `createPair` leaves the pair briefly empty, and **whoever adds liquidity first sets the initial price permanently** — the router creates it inside the same call as the mint, so there is no window.

`amountTokenMin` and `amountETHMin` were set **equal** to the desired amounts, so the call could only succeed at our ratio: with reserves at zero the router consumes exactly the desired amounts, and against any pre-existing reserves it would have reverted rather than accept someone else's price.

Seeded **22,918.5 OLAS / 0.25 WETH** at the live mainnet ratio of 91,674 OLAS/WETH, minting `75.694286442240803044` LP. LP `totalSupply` exceeds the deployer's balance by exactly **1000** — the `MINIMUM_LIQUIDITY` burn of a genuine first mint. The address matches the CREATE2 prediction, which was itself validated by reproducing an existing 4663 pair from its tokens and init-code hash.

A bot bought 91 OLAS for 0.001 WETH **fifteen blocks later** — the clearest available evidence the pool is indexed on 4663.

## Versions: in step with the fleet, not ahead of it

Each build was proven before signing by substituting the reference chain's own immutables into the artifact and requiring an exact match:

| contract | reference | residual |
|---|---|---|
| `UniswapPriceOracle` | mainnet's live oracle | **0 bytes** |
[...]

PR #362, opened 2026-09-10
audit: equal length is not proof the artifact is the deployed build

DavidMinarsch:
`checkBytecode`'s Tier 1 compared **code lengths only**, and two revisions of the same contract can compile to the same length. Four L2 target dispensers do exactly that.

## What the length check was missing

`gnosis`, `optimism`, `base` and `mode` each match their artifact's length **exactly** while running a build a year newer than it. Diffing gnosis's dispatch table against its artifact's ABI:

```
in ARTIFACT but not deployed:  processDataMaintenance(bytes)
in DEPLOYED but not artifact:  processDataMaintenance(bytes,bool)     <- extra arg
                               updateWithheldAmountMaintenance(uint256)
```

`abis/0.8.25/OptimismTargetDispenserL2.json` was last regenerated **2024-07-09** (`c39050e`); those two functions arrive in the **2025-06** commits (`8056c87`, `1555832`). The artifact is a year stale, Tier 1 passed on all four, and the audit instantiated those contracts with an ABI that does not describe them. It happens not to break today only because the getters the checks call exist in both revisions.

## Tier 1b

Strip the CBOR trailer, then require every differing run in the executable bytecode to be **immutable-shaped**: an EVM immutable is at most one 32-byte word, so a longer contiguous run cannot be one.

The separation is wide, which is what makes this a safe check rather than a heuristic:

| | longest differing run |
|---|---|
| legitimate immutable deltas, all 9 chains | **≤ 20 B** (an inlined address) |
| the four wrong artifacts | **282 – 695 B** |

Run against the fleet it flags **exactly those four**, with **no false positives** among the 37 other contracts that carry metadata-trailer warnings.

## Why the four are allowlisted rather than left red

`KNOWN_STALE_ARTIFACTS` keeps the check **blocking for everything else** instead of weakening it to a warning. The four are reported loudly on every run, and deleting an entry is the point — it goes when the real artifact is recovered.

I tried to recover them and could not. From `513a2fc` (the commit that has both new functions), twelve combinations of solc 0.8.25, optimizer runs 100–200, viaIR on/off and four EVM versions produce 7672 / 7769 / 8436 / 8453 / 8487 / 8489 / 8660 stripped bytes — never the deployed 8463, and the lengths aren't even monotonic in `runs`. That is the same dead end this file already records for base's `BuyBackBurnerProxy`, which is left red on purpose.

So this PR does not add the missing artifacts. It makes their absence explicit, bounded and blocking for anything new, which is the part I can do correctly.

## One detail worth reviewing

The allowlist is keyed on **address AND contract name**. Addresses are not unique across chains in `configuration.json` — `0x4891f589…` is optimism's `BuyBackBurnerProxy` *and* celo's target dispenser. An address-only key would silently exempt an unrelated contract if one ever collided. I made precisely that mistake while investigating this and produced a false finding, which is why it is guarded here.

Also corrected the final summary line, which said "length mismatch (Tier 1)" and now names either cause.

## Scope

Independent of the Robinhood rollout — these are pre-existing deployments on four other chains. Split out of #361 review discussion at @DavidMinarsch's request.

### DIFF
--- a/scripts/audit_chains/audit_contracts_setup.js
+++ b/scripts/audit_chains/audit_contracts_setup.js
@@ -152,6 +152,59 @@ async function checkOwner(chainId, contract, globalsInstance, log) {
     };
 }
 
+// Four L2 target dispensers whose committed artifact is a year older than the deployment. Their
+// abis/0.8.25/ artifacts were last regenerated 2024-07-09 (c39050e), while the deployed code carries
+// processDataMaintenance(bytes,bool) and updateWithheldAmountMaintenance(uint256), added in the
+// 2025-06 commits (8056c87, 1555832) and shipped by 513a2fc. Each artifact happens to match its
+// deployment's length exactly, which is why the Tier-1 length check never caught this.
+//
+// The correct artifacts are not in this repo and could not be rebuilt: from 513a2fc, twelve
+// combinations of solc 0.8.25, optimizer runs 100-200, viaIR on/off and four EVM versions produce
+// 7672 / 7769 / 8436 / 8453 / 8487 / 8489 / 8660 stripped bytes and never the deployed 8463 (gnosis
+// 8492). The same dead end as the base BuyBackBurnerProxy noted above.
+//
+// Listed here so Tier 1b stays blocking for everything else rather than being weakened to a warning.
+// These are reported loudly on every run. Deleting an entry is the point: it goes when the real
+// artifact is recovered and committed under abis/deployed/.
+// Keyed by address AND the configuration entry's contract name. Addresses are not unique across
+// chains in this file - 0x4891f589... is optimism's BuyBackBurnerProxy and celo's target dispenser -
+// so an address-only key would silently exempt an unrelated contract if one ever collided.
+const KNOWN_STALE_ARTIFACTS = {
+    "0x5b6c538c7b2e0b44fa8a3b7a0532ef797b07d0e9": {name: "GnosisTargetDispenserL2", label: "gnosis"},
+    "0xaea9ef993d8a1a164397642648df43f053d43d85": {name: "OptimismTargetDispenserL2", label: "optimism"},
+    "0x9ec97be9ff55ff11606ce7c589956f7bf3d0b241": {name: "OptimismTargetDispenserL2", label: "base"},
+    "0xeb5638eefe289691ece01943f768edbf96258a80": {name: "OptimismTargetDispenserL2", label: "mode"},
+};
+
+// An EVM immutable occupies at most one 32-byte word, so a contiguous run of differing bytes longer
+// than this cannot be an immutable - it means the artifact is a different build.
+const MAX_IMMUTABLE_RUN_BYTES = 32;
+
+// Drop the CBOR metadata trailer, whose last two bytes give its own length. Returns a lowercase hex
+// string without the 0x prefix, or the input unchanged when the trailer is not well formed.
+function stripMetadata(hex) {
+    const h = hex.startsWith("0x") ? hex.slice(2) : hex;
+    if (h.length < 4) return h.toLowerCase();
+    const cborLen = parseInt(h.slice(-4), 16);
+    const cut = (cborLen + 2) * 2;
+    if (!Number.isFinite(cborLen) || cut <= 0 || cut > h.length) return h.toLowerCase();
+    return h.slice(0, h.length - cut).toLowerCase();
+}
+
+// Group differing byte positions of two equal-length hex strings into contiguous runs.
+function diffByteRuns(aHex, bHex) {
+    const runs = [];
+    let start = -1;
+    const n = Math.min(aHex.length, bHex.length) / 2;
+    for (let i = 0; i < n; i++) {
+        const differs = aHex.slice(i * 2, i * 2 + 2) !== bHex.slice(i * 2, i * 2 + 2);
+        if (differs && start < 0) start = i;
+        if (!differs && start >= 0) { runs.push({start, length: i - start}); start = -1; }
+    }
+    if (start >= 0) runs.push({start, length: n - start});
+    return runs;
+}
+
 // Check the bytecode
 // idx selects which entry to check when configuration.json holds more than one contract under the
 // same name — mirroring findContractInstance(). Without it every same-named entry resolved to the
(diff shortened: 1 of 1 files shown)


scripts/audit_chains/audit_contracts_setup.js
```diff
@@ -220,6 +273,43 @@ async function checkBytecode(provider, configContracts, contractName, log, idx =
...
+            // difference to be immutable-shaped. An EVM immutable is at most one 32-byte word, so a
+            // contiguous differing run longer than that cannot be one: it is a different build. Inlined
+            // addresses (20 B) and small ints sit far below the threshold - across the fleet the largest
+            // legitimate run is 20 B, while a wrong artifact shows runs of 282-695 B.
+            const execArtifact = stripMetadata(bytecode);
+            const execOnChain = stripMetadata(onChainCode);
+            if (execArtifact.length === execOnChain.length) {
+                const diffRuns = diffByteRuns(execArtifact, execOnChain);
+                const longest = diffRuns.reduce((m, r) => Math.max(m, r.length), 0);
+                if (longest > MAX_IMMUTABLE_RUN_BYTES) {
```
OjusWiZard, line 293: This establishes that a **FAIL** is real, but a **pass** doesn't establish that the artifact is the deployed build. Only the longest single run is tested.

**What passes today.** A same-length build whose differences are all ≤ 32 B clears Tier 1b. For example:
- a changed hard-coded `PUSH20` address;
- a flipped `LT`/`GT`;
- a changed `PUSH4` selector.

It then reaches Tier 2 and prints the same one-line `metadata-trailer drift` WARN as the ~37 benign ones, and exits 0. That is the "ABI doesn't describe the contract" case this PR is closing, just with short diffs. So "a safe check rather than a heuristic" doesn't quite hold: the 282-695 B separation is a threshold that happens to hold on today's fleet.

**An exact check costs about the same.** Allow differences only inside the artifact's **unfilled immutable slots**: `PUSH32` operands that are all zero in the artifact's code. Everything else must match byte for byte. It needs no `immutableReferences`, so it works on the hardhat `abis/<solc>/` artifacts too.

**Fleet run.** I ran it over every `configuration.json` entry on all nine chains, against live code, with the CBOR trailer stripped as in `stripMetadata`:

```
genuine contracts            59 / 59 pass, 0 false positives
the four stale dispensers    fail - 7500 / 7496 / 7494 / 7496 differing bytes outside immutable slots
zero-PUSH32 scan vs forge immutableReferences   identical on all 15 forge artifacts
```

A linear opcode walk is enough - no fleet contract misaligned on embedded data.

```js
// Byte ranges of PUSH32 operands that are all-zero in the artifact: unfilled immutable slots.
function immutableSlots(hex) {
    const b = Buffer.from(hex, "hex");
    const slots = [];
    for (let i = 0; i < b.length; ) {
        const op = b[i];
        if (op >= 0x60 && op <= 0x7f) {
            if (op === 0x7f && b.subarray(i + 1, i + 33).every((x) => x === 0)) slots.push([i + 1, i + 33]);
            i += 1 + (op - 0x5f);
        } else {
            i += 1;
        }
    }
    return slots;
}
```

Then count differing bytes between `execArtifact` and `execOnChain` that fall outside those ranges, and FAIL on any. `KNOWN_STALE_ARTIFACTS` keeps working unchanged on top of that.

OjusWiZard (approved):
Approving - strictly better than `main`.

I replicated Tier 1b over every configured contract on all nine chains. It flags exactly the four allowlisted dispensers, with no false positives, and no contract hits the stripped-length skip today. It also merges cleanly with #361.

One inline comment, on what a pass does and doesn't establish, with an exact alternative that validates on the same fleet.

scripts/audit_chains/audit_contracts_setup.js
```diff
@@ -220,6 +273,43 @@ async function checkBytecode(provider, configContracts, contractName, log, idx =
...
+                    const total = diffRuns.reduce((t, r) => t + r.length, 0);
+                    const entry = KNOWN_STALE_ARTIFACTS[configContracts[i]["address"].toLowerCase()];
+                    const known = (entry && entry.name === configContracts[i]["name"])
```
dagacha, line 296: [P1] Validate immutable offsets instead of accepting every short difference

A differing run of at most 32 bytes is not sufficient to establish that it is an immutable: changed opcodes, function selectors, and constants also fit below this threshold. With the committed DonatorBlacklist artifact, replacing its first opcode (0x60) with STOP (0x00), while keeping all other bytes and metadata unchanged, leaves bytecodeMismatchFound false and emits no warning. This allows materially different executable behavior to pass the new build check. Please compare executable bytes after masking only compiler-provided immutable reference ranges (and explicitly supported link references), rejecting differences outside those ranges. Add a regression case for a one-byte opcode change alongside a legitimate immutable substitution.

scripts/audit_chains/audit_contracts_setup.js
dagacha, line 293: [P2] Reject unequal executable lengths instead of skipping Tier 1b

Equal total bytecode lengths do not imply equal executable lengths when the metadata trailers have different lengths. This conditional skips all executable verification in that case and falls through to a warning. In an isolated reproduction, changing 64 consecutive executable bytes correctly sets bytecodeMismatchFound; additionally changing the final trailer-length field from 0x0033 to 0x0032 makes the same mismatch warning-only. stripMetadata currently trusts that field without validating the CBOR, so malformed trailers can trigger this too. Please validate the metadata trailer and make unequal stripped executable lengths blocking, with regression coverage for both unequal metadata lengths and malformed trailers.

dagacha (commented):
Review performed with Codex (LLM: GPT-6; reasoning effort: Medium).

Two actionable findings are attached inline. Both were reproduced with isolated calls to checkBytecode from this exact PR revision, using the committed DonatorBlacklist artifact and a mocked provider. JavaScript syntax validation passed; the full live-chain audit was not run.

scripts/audit_chains/audit_contracts_setup.js
```diff
@@ -220,6 +273,43 @@ async function checkBytecode(provider, configContracts, contractName, log, idx =
...
+            // target dispensers each matched their artifact's length exactly while running a build a year
+            // newer than it, with processDataMaintenance(bytes) replaced by processDataMaintenance(bytes,
+            // bool) and updateWithheldAmountMaintenance(uint256) added. The length check passed and the
+            // audit instantiated those contracts with an ABI that did not describe them.
+            //
+            // So compare the executable bytecode, with the CBOR trailer removed, and require every
+            // difference to be immutable-shaped. An EVM immutable is at most one 32-byte word, so a
+            // contiguous differing run longer than that cannot be one: it is a different build. Inlined
+            // addresses (20 B) and small ints sit far below the threshold - across the fleet the largest
+            // legitimate run is 20 B, while a wrong artifact shows runs of 282-695 B.
```
dagacha, line 287: Actionable: this sentence is false, and it is the justification people will use to retune the constant.

Excluding the four stale dispensers, the fleet longest-run histogram is 0 (17), 16, 19, **20 (35)**, and **32 (1)** — mainnet `Dispenser` `0x5650300f…`, 34 runs / 324 B total. That 32 B run is a full-word immutable, not a different revision.

`longest > 32` therefore passes it. `MAX_IMMUTABLE_RUN_BYTES = 32` is the right value (one EVM word). If anyone tightens it to 20 to match this comment, Dispenser fails Tier 1b.

Replace “largest legitimate run is 20 B” with “largest observed is 32 B (Dispenser); addresses are 20 B”.

dagacha (approved):

Independent fleet run: all 60 `configuration.json` entries, same strip-and-run logic as this patch.

Tier 1b flags **exactly the four allowlisted dispensers** (longest run 282 / 695 / 299 / 695 B). Full lengths 8545/8545 gnosis, 8516/8516 optimism+base+mode; stripped 8492 and 8463. No other contract has a run > 32 B. Same-total-length pairs always strip to the same exec length (CBOR 51 everywhere), so the equal-exec-length gate does not skip anyone.

Selectors on those four: artifact has `processDataMaintenance(bytes)`; on-chain has `processDataMaintenance(bytes,bool)` and `updateWithheldAmountMaintenance(uint256)`. Polygon / Arb / Celo / Robinhood (`abis/deployed/`) match on both sides.

Allowlist addresses and names match config. `0x4891f589…` is optimism `BuyBackBurnerProxy` (266 B) and celo `OptimismTargetDispenserL2` (7520 B) — address-only would be wrong.

One leftover, inlined. Not a merge blocker — the constant is already right.

PR #363, opened 2026-09-10
audit: check immutable slots exactly, instead of a run-length threshold

DavidMinarsch:
The Tier 1b threshold shipped in #362 was not sound, and its justifying comment was false. Both caught by @dagacha and @OjusWiZard; this replaces it with an exact check.

## What was wrong

**The threshold didn't establish what it claimed.** A differing run of ≤32 B does not mean "immutable" — a changed `PUSH20` address, a flipped `LT`/`GT`, a changed `PUSH4` selector all fit under it. @dagacha demonstrated it directly: replacing `DonatorBlacklist`'s first opcode with `STOP` passes.

**The comment justifying the constant was factually wrong.** I wrote that the fleet's longest legitimate run is 20 B. It is **32 B** — mainnet's `Dispenser` `0x5650300f…` carries a full-word immutable. So `longest > 32` sat exactly on a real value, one byte from rejecting a genuine contract, and that sentence is what someone would have used to retune it.

**Unequal executable lengths skipped the check entirely.** Equal *total* length doesn't imply equal executable length — the two CBOR trailers can themselves differ in length. The old code skipped the whole comparison and fell through to a warning.

## The exact check

@OjusWiZard's approach: a difference is allowed **only** where the artifact has an unfilled immutable slot — a `PUSH32` whose operand is all zero. Everything else must match byte for byte. **No constant left to tune.**

It needs no `immutableReferences`, so it works on the hardhat `abis/<solc>/` artifacts as well as forge ones.

## I verified the premise rather than trusting it

Scanning for zero-`PUSH32` operands agrees **exactly** with forge's own `immutableReferences` on all **13/13** forge artifacts `configuration.json` references.

It can over-approximate in principle — a contract legitimately pushing an all-zero 32-byte constant would have that treated as a slot. One such artifact exists in `out/` (`ZuniswapV2Library`, 4 scanned vs 1 declared) and is **not** an audited contract. That direction is permissive, never a false FAIL, and the caveat is recorded next to the function rather than left for someone to rediscover.

Unequal executable length is now a mismatch rather than a skip.

## Result

Fleet run unchanged in outcome: **no hard FAILs, the same four known-stale dispensers, exit 0.** `diffByteRuns` removed as dead.

`KNOWN_STALE_ARTIFACTS` works unchanged on top, still keyed on address **and** contract name.

### DIFF
--- a/scripts/audit_chains/audit_contracts_setup.js
+++ b/scripts/audit_chains/audit_contracts_setup.js
@@ -176,9 +176,47 @@ const KNOWN_STALE_ARTIFACTS = {
     "0xeb5638eefe289691ece01943f768edbf96258a80": {name: "OptimismTargetDispenserL2", label: "mode"},
 };
 
-// An EVM immutable occupies at most one 32-byte word, so a contiguous run of differing bytes longer
-// than this cannot be an immutable - it means the artifact is a different build.
-const MAX_IMMUTABLE_RUN_BYTES = 32;
+// Byte ranges of PUSH32 operands that are all zero in the artifact's code: solc reserves one 32-byte
+// slot per immutable and leaves it zeroed in the compiled output, filling it at construction. Those
+// ranges are the only places a deployed runtime may legitimately differ from its artifact.
+//
+// A linear opcode walk is enough - skipping each PUSH's operand keeps the walk aligned, and no
+// contract in the fleet misaligns on embedded data. This needs no immutableReferences, so it works on
+// the hardhat abis/<solc>/ artifacts as well as the forge ones.
+//
+// Verified against forge's own immutableReferences: the scan agrees exactly on all 13 forge artifacts
+// configuration.json references. It can over-approximate in principle - a contract that legitimately
+// pushes an all-zero 32-byte constant would have that treated as a slot, permitting a difference there.
+// One such case exists in out/ (ZuniswapV2Library, 4 scanned vs 1 declared) and it is not an audited
+// contract. The failure direction is permissive, never a false FAIL.
+function immutableSlots(hex) {
+    const b = Buffer.from(hex, "hex");
+    const slots = [];
+    for (let i = 0; i < b.length; ) {
+        const op = b[i];
+        if (op >= 0x60 && op <= 0x7f) {
+            if (op === 0x7f && b.subarray(i + 1, i + 33).every((x) => x === 0)) slots.push([i + 1, i + 33]);
+            i += 1 + (op - 0x5f);
+        } else {
+            i += 1;
+        }
+    }
+    return slots;
+}
+
+// Differing bytes between two equal-length hex strings that fall outside the given byte ranges.
+function diffsOutsideSlots(aHex, bHex, slots) {
+    const inSlot = new Uint8Array(aHex.length / 2);
+    for (const [from, to] of slots) {
+        for (let i = from; i < to && i < inSlot.length; i++) inSlot[i] = 1;
+    }
+    const offsets = [];
+    for (let i = 0; i < inSlot.length; i++) {
+        if (inSlot[i]) continue;
+        if (aHex.slice(i * 2, i * 2 + 2) !== bHex.slice(i * 2, i * 2 + 2)) offsets.push(i);
+    }
+    return offsets;
+}
 
 // Drop the CBOR metadata trailer, whose last two bytes give its own length. Returns a lowercase hex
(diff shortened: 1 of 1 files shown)


scripts/audit_chains/audit_contracts_setup.js
```diff
@@ -280,34 +304,47 @@ async function checkBytecode(provider, configContracts, contractName, log, idx =
...
-            // legitimate run is 20 B, while a wrong artifact shows runs of 282-695 B.
+            // So compare the executable bytecode with the CBOR trailer removed, and allow a difference
+            // only where the artifact has an unfilled immutable slot - a PUSH32 whose operand is all
+            // zero. Everything else must match byte for byte.
+            //
+            // An earlier revision of this check used a run-length threshold instead, which was not sound:
+            // a changed PUSH20 address, a flipped LT/GT or a changed PUSH4 selector all fit inside 32
+            // bytes and would have passed. The fleet's longest legitimate run is also 32 B, not 20 -
+            // mainnet's Dispenser 0x5650300f carries a full-word immutable - so the threshold sat exactly
+            // on a real value. The slot check has no threshold to tune: 59 of 59 genuine fleet contracts
```
dagacha, line 315: Actionable: this comment is right; `main()` still prints the old check.

`console.error` at the bottom of this file still says `Tier 1b: a differing run too long to be an immutable`. That is the #362 summary line. This PR replaced the check and did not replace the string. Operators grepping FAIL will look for a threshold that no longer exists.

Same class of fix #362 itself made to that line. Name length vs bytes-outside-unfilled-PUSH32-slots.

dagacha (approved):

Independent fleet run of the new slot check, not the description.

A byte may differ only inside an artifact `PUSH32` whose operand is all zero. Unequal executable length after CBOR strip is a mismatch, not a skip. `diffByteRuns` / `MAX_IMMUTABLE_RUN_BYTES` are gone.

64 `configuration.json` entries:

| | |
|---|---|
| pass 1b | **59** |
| known-stale | **4** — 7500 / 7496 / 7494 / 7496 B outside slots |
| Tier 1 length | **1** — Celo `UniswapPriceOracle` (audit block already skips it) |
| hard 1b FAIL | **0** |

Replacing the first opcode with `STOP` now fails on all 63 length-matched contracts. The DonatorBlacklist hole from #362 is closed.

Forge `immutableReferences`: **13/13** unique forge artifacts that declare refs match the scan exactly; every ref length is 32. Dispenser (the 32 B word) is 25 slots, 0 offenders.

`KNOWN_STALE_ARTIFACTS` still keyed on address **and** name.

One leftover, inlined. Not a merge blocker — the check is already right.

dagacha (approved):

dagacha (commented):
Reviewed with Codex (LLM: GPT-6; reasoning effort: Medium).

No actionable findings at 82f83f825501ccf1212a1aa4bd404f63c2f95836.

Validated all 41 configured artifacts: inferred immutable slots match compiler references for all 13 artifacts that provide them. Eleven targeted checks passed, covering identical runtime, full-word immutable replacement, opcode and non-slot operand mutations, metadata-only drift, unequal executable lengths, total-length mismatch, known-stale handling and contract-name scoping, and the real DonatorBlacklist STOP regression.

The live fleet audit was not rerun locally. The documented zero-PUSH32 constant caveat remains. As a non-blocking follow-up, consider committing regression tests for these cases so future changes preserve the behavior.

OjusWiZard (approved):
Approving. I ran this PR's own functions, extracted from the diff, over every configured contract's live code: **59 pass, the same 4 known-stale dispensers (7500 / 7496 / 7494 / 7496 B outside slots), 0 FAIL**.

Against a real 4663 deployment, a byte changed inside a `PUSH20` constant and `STOP` in place of the first opcode both FAIL, while a byte changed inside an immutable slot passes. Nothing still references the removed identifiers.

---

The final failure message still describes the check this PR removes (`audit_contracts_setup.js:1698`):

```js
console.error("AUDIT FAILED: at least one contract does not match its artifact (Tier 1: length, or Tier 1b: a differing run too long to be an immutable) — see FAIL lines above.");
```

Tier 1b no longer has a notion of runs, so every Tier 1b FAIL now ends the run with a reason the code doesn't test. Something like `Tier 1b: a byte differs outside the artifact's zeroed PUSH32 slots` would match what it does now.

*(In the body rather than inline: line 1698 is unchanged, so it sits outside the diff.)*

PR #364, opened 2026-09-10
audit: name the check that actually runs in the failure summary

DavidMinarsch:
#363 replaced Tier 1b's run-length threshold with the unfilled-`PUSH32`-slot comparison but left `main()`'s `console.error` describing the old one:

```
AUDIT FAILED: ... Tier 1b: a differing run too long to be an immutable
```

A FAIL therefore pointed operators at a threshold that no longer exists. Now:

```
AUDIT FAILED: ... Tier 1b: bytes differing outside the artifact's unfilled PUSH32 immutable slots
```

Caught by @dagacha on the merged PR. It is the same fix #362 made to this line — I made it once and then did not repeat it when the check changed underneath.

### DIFF
--- a/scripts/audit_chains/audit_contracts_setup.js
+++ b/scripts/audit_chains/audit_contracts_setup.js
@@ -1695,7 +1695,7 @@ async function main() {
 main()
     .then(() => {
         if (bytecodeMismatchFound) {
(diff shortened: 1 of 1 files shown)


dagacha (approved):

OjusWiZard (approved):
Approving. The summary line now describes the check `checkBytecode` actually runs, and nothing on `main` still references the removed run-length threshold.

PR #368, opened 2026-09-21
docs: correct the withheld reconciliation in the dispenser runbook

DavidMinarsch:
Docs only. The description is the **final state** after review — the first push carried a two-case step 14 and an L1-side alternative, both of which are gone.

## The error

§4 stated there is **"no L1-side admin setter"** for `mapChainIdWithheldAmounts`. There is: [`syncWithheldAmountMaintenance`](https://github.com/valory-xyz/autonolas-tokenomics/blob/main/contracts/Dispenser.sol) is `owner`-gated and its NatSpec says it exists to *"restore the data that was not delivered from L2"*.

That sentence had a consequence beyond itself. Because the map looked unrecoverable, §4 declared a non-zero L1 withheld state **out of scope** — *"stop and reconcile it first, the steps below do not cover a non-zero L1 withheld state"* — and §9 repeated it. So the runbook could not be executed at all once any chain had synced a withheld amount up to L1, which is now the case for six.

## What changed

**§4 — two readings, not one.** The L2 `withheldAmount` counter is zeroed once the amount syncs to L1, so a dispenser can hold the OLAS while reporting `withheldAmount == 0`. Phase 0 now records **both** the counter and the OLAS balance, and the precondition block covers the non-zero branch instead of stopping at it.

**Step 14 — one rule: pass the final OLAS balance of the new dispenser.** This is what `updateWithheldAmountMaintenance` already documents for this case:

> `[2] Withheld amount update after balance migration to a new contract` … `The amount here must correspond to` … `[2] Final OLAS balance of this contract address.`

Two independent reasons the previous "emitted withheld value" instruction could not work:

1. Where the chain has already synced to L1, the emitted value is `0` while the balance is not — restoring it strands the OLAS on both sides.
2. **The deployed dispensers emit the three-argument `Migrated(address,address,uint256)` and carry no withheld value at all.** The four-argument form is on the new build — the contract being migrated **to**, never the one `migrate()` runs on. The migrated balance is the only figure the event gives an operator at step 13, on every chain.

Where L1 read `0` the balance and the emitted value coincide, so the single rule covers both readings.

**The L1-side route is dropped, not caveated.** It would require the DAO to supply a `batchHash` for a message that never failed, and `updateHashMaintenance` marks that hash processed on the new L1 processor — poisoning a hash the new dispenser is going to produce makes that real sync revert `AlreadyDelivered` and lose its amount. No upside to weigh against that, since the L2 maintenance call does the job by design.

**Step 22 stops being a migration step.** With step 14 correct the amount reaches L1 through the ordinary `syncWithheldAmount` path. What remains is a check that it happened before the first full epoch distribution, plus a pointer that a shortfall is a step-14 error rather than something 22 can repair — the sync path rejects a zero amount. The old "skip if dust" framing invited exactly the omission the fix is about.

§9 updated to match.

## Verified on a Base fork

Against the live dispenser `0x9Ec97Be9…b241` (1,380,908.799963024002602865 OLAS at `withheldAmount == 0`), with the **real `OptimismTargetDispenserL2` from this repo's `main` deployed on the fork** and the actual cutover run — old `pause` → `migrate` → new `updateWithheldAmountMaintenance` → new `syncWithheldAmount`. 5/5:

- `migrate()` carries the full balance, reports only `amount`; the new contract deploys at withheld 0 / nonce 0
- restoring the emitted `0` → the OLAS is on the new dispenser and `syncWithheldAmount` reverts `ZeroValue`
- restoring the carried balance → the sync posts that exact amount under `keccak256(abi.encode(0, 8453, newDispenser))`
- batch hashes are bound to the dispenser address, so a fresh dispenser's cannot collide with the old one's by accident

The test currently lives in our analytics repo; happy to port it here so it travels with the runbook — it needs only a Base RPC and has no dependency on that repo.

## Scope

Docs only. **No procedural change for the all-zero case**, which behaves exactly as before — the additions are the non-zero branch, the Phase 0 balance read, and the corrected step 14.


docs/dispenser_migration_runbook_public.md
```diff
@@ -106,7 +110,10 @@ For every chain with an L2 dispenser (Arbitrum, Gnosis, Optimism, Base, Polygon,
...
+    - **L1 read `0` for this chain** → pass the **emitted `withheldAmount`** from `Migrated`. The accounting still lives on L2 and this restores it unchanged.
+    - **L1 read non-zero for this chain** → the emitted `withheldAmount` is **`0`** (the accounting moved to L1 when it was synced), while `migrate()` still carried a non-zero OLAS **balance**. Passing `0` here leaves that OLAS accounted for by neither side: the new L1 Dispenser also starts at `0`, nothing nets against it, and the new L2 dispenser quietly accumulates unusable OLAS. Pass the **migrated balance** from `Migrated` instead, so step 22 can carry it up.
+    - *Alternative for the non-zero case:* leave this at `0` and restore the L1 side directly with `Dispenser.syncWithheldAmountMaintenance(chainId, amount, batchHash)` **after step 19** (it resolves the processor via `mapChainIdDepositProcessors` and calls `updateHashMaintenance`). That consumes a batch hash on the new processor, so prefer the L2 route unless the interaction with the new L2 dispenser's first genuine sync has been checked.
```
DavidMinarsch: This is not valid afaik because the l2 side has no valid batchHash to emit because it restarts nonce at 0?

To me it seems the only valid approach is via `updateWithheldAmountMaintenance` and one has to make sure no sync is currently in place between chains (which is easy to do because DAO controls it)

docs/dispenser_migration_runbook_public.md
```diff
@@ -120,7 +127,7 @@ For every chain with an L2 dispenser (Arbitrum, Gnosis, Optimism, Base, Polygon,
...
-22. **Withheld re-sync (only if a carried L2 balance is material):** the new L1 Dispenser starts with `mapChainIdWithheldAmounts = 0` and does not know about the balance carried in step 14, so it will not *net* the next distribution against it (it bridges fresh OLAS; funds are not lost, the L2 stays "ahead"). To restore netting, trigger a withheld sync from the new L2 dispenser up to the new L1 Dispenser after wiring. If the carried balance is ~0/dust, skip.
+22. **Withheld re-sync (only if a carried L2 balance is material):** the new L1 Dispenser starts with `mapChainIdWithheldAmounts = 0` and does not know about the balance carried in step 14, so it will not *net* the next distribution against it (it bridges fresh OLAS; funds are not lost, the L2 stays "ahead"). To restore netting, trigger a withheld sync from the new L2 dispenser up to the new L1 Dispenser after wiring. **This only does anything if step 14 set a non-zero `withheldAmount`** — the sync path rejects a zero amount, so a chain left at `0` there stays unnetted no matter how often this is run. If the carried balance is ~0/dust, skip.
```
DavidMinarsch: This is not needed at all with updated step 14 afaik

docs/dispenser_migration_runbook_public.md
DavidMinarsch: Agreed, and removed rather than caveated in a4bd9b5.

Confirming the mechanism, because it is slightly worse than "no valid batchHash": `updateHashMaintenance` is `l1Dispenser`-only and sets `processedHashes[batchHash] = true` on the **new** L1 processor. The hash the L2 side produces is `keccak256(abi.encode(batchNonce, block.chainid, address(this)))` — the nonce restarts at 0 on the new dispenser, so if the DAO supplies a hash that collides with one the new L2 will genuinely produce, `_receiveMessage` later reverts `AlreadyDelivered(batchHash)` and that real sync's amount is lost. There is no upside to weigh against that, since the L2 maintenance call does the job by design.

Your "make sure no sync is currently in flight" condition is the right one, and it is cheap — the DAO owns both the pause and the sync trigger.

docs/dispenser_migration_runbook_public.md
DavidMinarsch: Agreed — step 22 is no longer a migration step in a4bd9b5.

And you pushed me to a better step 14. It is now **one rule, not a two-case branch: pass the final OLAS balance of the new dispenser.** That is what `updateWithheldAmountMaintenance` already documents for this case:

> `[2] Withheld amount update after balance migration to a new contract` … `The amount here must correspond to` … `[2] Final OLAS balance of this contract address.`

So the contract had the answer, and both the original runbook and my first fix were reading the emitted `withheldAmount` instead. Where L1 read `0` the balance and the emitted value coincide, so the single rule covers both readings.

What is left at step 22 is a check rather than an action: confirm the amount reached L1 through the ordinary `syncWithheldTokens` path before the first full epoch distribution, and note that a shortfall there points back to step 14 rather than being repairable at 22 (the sync path rejects a zero amount). The old "skip if dust" framing invited exactly the omission the fix is about.

Fork test on a Base fork to follow — migrate → restore → sync against the live 1,380,908.80 balance.

DavidMinarsch:
Both review points applied, plus a third thing that fell out of testing it.

**The deployed dispensers emit the three-argument `Migrated(address,address,uint256)`** — no `withheldAmount` at all. The four-argument form is on the new build, i.e. the contract being migrated **to**, never the one `migrate()` runs on. So "restore the emitted withheld value" was not merely wrong for chains that had already synced; the value does not exist on any chain at step 13. The migrated balance is the only figure the event gives you, which is the same answer the NatSpec gives.

**Verified on a Base fork** against the live dispenser `0x9Ec97Be9…b241`, holding 1,380,908.799963024002602865 OLAS at `withheldAmount == 0` — 4/4:

| test | result |
|---|---|
| `test_balanceAndWithheldHaveDiverged` | balance equals the L1 credit to the wei, counter is `0`, nonce is `1` |
| `test_migrateCarriesBalanceAndReportsOnlyAmount` | full balance moves, old dispenser emptied and bricked, event reports only `amount` |
| `test_zeroWithheldCannotBeSynced` | `syncWithheldAmount` reverts `ZeroValue` at `withheldAmount == 0` — step 22 cannot repair a step-14 omission |
| `test_restoreBalanceThenSync` | restore the balance → sync sends the full amount and consumes one nonce |

The test lives in our analytics repo rather than here since it asserts against live chain state; happy to port it into this repo's suite if you would rather it travelled with the runbook.

DavidMinarsch:
Redone properly: the test now **deploys the real `OptimismTargetDispenserL2` from this repo's `main`** (9197796) on a Base fork and runs the actual cutover, rather than asserting against the old side with a stand-in target. 5/5.

Path: old `pause` → old `migrate(new)` → new `updateWithheldAmountMaintenance(...)` → new `syncWithheldAmount("")`.

| test | what it establishes |
|---|---|
| `test_balanceAndWithheldHaveDiverged` | live dispenser holds 1,380,908.799963024002602865 at `withheldAmount == 0`, nonce 1 |
| `test_migrateIntoNewBuildCarriesBalanceOnly` | full balance lands on the new contract, old one emptied and bricked, event reports only `amount`; new deploys at withheld 0 / nonce 0 |
| `test_restoringEmittedZeroStrandsTheBalance` | **the bug:** restore the emitted `0` → the OLAS is on the new dispenser and `syncWithheldAmount` reverts `ZeroValue`, so L1 never learns about a balance it is holding |
| `test_restoringBalanceSyncsFullAmount` | **the fix:** restore the carried balance → the sync posts that exact amount under `keccak256(abi.encode(0, 8453, newDispenser))`, leaving withheld 0 and nonce 1 |
| `test_batchHashIsBoundToDispenserAddress` | a fresh dispenser's hashes cannot collide with the old one's by accident |

That last one sharpens the earlier point about the L1-side route: because the hash is bound to the dispenser **address**, an L1 maintenance sync does not collide by accident — it would have to supply a hash the new dispenser is going to produce, and poison it. Still a hazard, just a deliberate-footgun one rather than an inevitable one. The L2 route remains the right call.

The test and the three vendored contracts live in our analytics repo (`forge/test/ForkWithheldMigration.t.sol`, `forge/test/vendor/` with provenance recorded). Offer stands to port it here so it travels with the runbook — it needs only a Base RPC and has no dependency on our repo.

docs/dispenser_migration_runbook_public.md
```diff
@@ -106,7 +110,15 @@ For every chain with an L2 dispenser (Arbitrum, Gnosis, Optimism, Base, Polygon,
 
 12. `pause()` the **old** L2 dispenser (`migrate` requires paused).
 13. `migrate(newL2TargetDispenser)` on the old L2 dispenser — transfers its **full OLAS balance** (withheld + any residual) to the new one, zeroes the old owner and locks it permanently (one-way; the old dispenser is dead after this). The `Migrated` event surfaces both the migrated balance and the `withheldAmount` to restore.
```
OjusWiZard, line 112: Step 13 is unchanged and still asserts exactly what step 14 now spends three paragraphs refuting: *"The `Migrated` event surfaces both the migrated balance and the `withheldAmount` to restore."*

Step 13 is the step where the operator actually reads the event, so this is the sentence that will be acted on.

OjusWiZard (commented):
Post-merge review. The correction itself holds up — I replayed it on-chain: the deployed Base dispenser's runtime code carries topic0 `0x928fd553…` (`Migrated(address,address,uint256)`) and not `0xf2ea3ee6…`, `Dispenser.syncWithheldAmountMaintenance` is there at `Dispenser.sol:1396` and is `owner`-gated, and the six credited chains total 2,411,846.120060 OLAS against L2 balances that match to the wei.

Four things left behind, all cheap to fix.

docs/dispenser_migration_runbook_public.md
```diff
@@ -106,7 +110,15 @@ For every chain with an L2 dispenser (Arbitrum, Gnosis, Optimism, Base, Polygon,
...
-14. On the **new** L2 dispenser, `updateWithheldAmountMaintenance(withheldAmount)` to re-establish its `withheldAmount` = the **emitted** withheld value (not necessarily the migrated balance; it deploys at 0).
+14. On the **new** L2 dispenser, `updateWithheldAmountMaintenance(<final OLAS balance of the new dispenser>)` to re-establish its `withheldAmount` (it deploys at 0). **Pass the migrated balance, not the emitted `withheldAmount`** — which is what the function's own NatSpec requires for this case: *"[2] Withheld amount update after balance migration to a new contract … The amount here must correspond to … [2] Final OLAS balance of this contract address."*
```
OjusWiZard, line 113: This rule now contradicts `migrate()`'s own NatSpec in this repo, still on `main` after the merge — `contracts/staking/DefaultTargetDispenserL2.sol:529`:

> The value to restore is the emitted withheldAmount, **NOT** the migrated balance (they differ when residual OLAS sits on the contract).

The runbook says the opposite, and the contract is what an operator has open at step 13. The two NatSpec blocks disagree with each other too: `updateWithheldAmountMaintenance` says "Final OLAS balance of this contract address", which is the one quoted here. A one-line change to the `migrate()` block would settle it.

(The residual case the NatSpec names is real but narrow: queued amounts go to `queuedHashes`, not `localWithheldAmount` (`_processData`), so the balance includes them — though `queueHash` binds `address(this)`, so they are unredeemable after migration anyway. Worth a clause either way, since Phase 0's new read list doesn't cover queued state.)

docs/dispenser_migration_runbook_public.md
```diff
@@ -106,7 +110,15 @@ For every chain with an L2 dispenser (Arbitrum, Gnosis, Optimism, Base, Polygon,
...
+
+    The two values differ whenever the chain has already synced a withheld amount up to L1: the emitted `withheldAmount` is then `0` (the accounting moved to L1) while `migrate()` still carried a non-zero **balance**. Passing that `0` leaves the OLAS accounted for by neither side — the new L1 Dispenser also starts at `0`, so nothing nets against it and the new L2 dispenser accumulates unusable OLAS while L1 mints fresh. Where L1 read `0`, the balance and the emitted value coincide and the same rule gives the same answer.
+
+    There is a second reason not to reach for the emitted value: **the deployed L2 dispensers emit the three-argument `Migrated(address,address,uint256)` and do not report `withheldAmount` at all.** The four-argument form is on the *new* build — the contract being migrated **to**, never the one `migrate()` runs on. So at step 13 the migrated balance is the only figure the event gives you, on every chain.
```
OjusWiZard, line 117: "never the one `migrate()` runs on" and "on every chain" are not quite true: **Robinhood Chain 4663's deployed dispenser is already on the new build.**

`eth_getCode` on `0xc40C79C275F3fA1F3f4c723755C81ED2D53A8D81` (4663) contains only the 4-arg topic `0xf2ea3ee6…` — as a PUSH32 literal at offset 2180 — and no 3-arg one. Eight of the nine deployed dispensers are 3-arg; that one is the exception.

No impact on this cascade (its balance, `withheldAmount` and L1 credit are all 0, and it isn't in the step-12 chain list) — but see the separate point about whether it *should* be in that list.

docs/dispenser_migration_runbook_public.md
```diff
@@ -106,7 +110,15 @@ For every chain with an L2 dispenser (Arbitrum, Gnosis, Optimism, Base, Polygon,
...
+
+    Verified on a Base fork against the live dispenser `0x9Ec97Be9…b241` (holding 1,380,908.799963024002602865 OLAS at `withheldAmount == 0`): `migrate()` carries the full balance and emits only the amount; `syncWithheldAmount` reverts `ZeroValue` while `withheldAmount` is `0`; and restoring the balance then syncing sends the full amount and consumes one batch nonce.
+
+    Once restored, the amount reaches the new L1 Dispenser through the **ordinary** `syncWithheldTokens` path, which is what re-enables netting. There is no L1-side shortcut worth taking here: `Dispenser.syncWithheldAmountMaintenance` would require the DAO to supply a `batchHash` for a message that never failed, and `updateHashMaintenance` marks that hash `processed` on the new L1 processor — if it collides with one the new L2 dispenser will genuinely produce (its nonce restarts at 0), that real sync later reverts `AlreadyDelivered` and its amount is lost. Use the L2 maintenance call.
```
OjusWiZard, line 121: `syncWithheldTokens` doesn't exist. That name survives only in the audit-era flattened snapshots (`audits/internal4/analysis2/contracts/DefaultTargetDispenserL2-flatten.sol:415`); the shipping contract has `syncWithheldAmount(bytes)` at `contracts/staking/DefaultTargetDispenserL2.sol:382`, and the deployed Base dispenser carries selector `0x7424ddc8` but not `0x29f404cd`.

Same slip at line 135. Lines 50 and 119 use the right name.

docs/dispenser_migration_runbook_public.md
```diff
@@ -45,8 +45,12 @@ This migration is a whole-stack cascade only because `Dispenser` and `VoteWeight
...
+**The only cross-chain state — `mapChainIdWithheldAmounts`.** This tracks OLAS already bridged to an L2 that staking targets could not absorb, netted against the *next* distribution to that chain. The OLAS itself sits on the **L2 target dispenser**, not on L1 — but note the L2's own `withheldAmount` counter is **zeroed once that amount has been synced up to L1**, so an L2 dispenser can hold the OLAS while reporting `withheldAmount == 0`. The balance and the counter are not the same reading.
+
+The map is mutated by netting, by `syncWithheldAmount` from an L1 processor, and by **`syncWithheldAmountMaintenance`** — an `owner`-gated L1 admin setter in [`Dispenser.sol`](https://github.com/valory-xyz/autonolas-tokenomics/blob/main/contracts/Dispenser.sol) that exists precisely to restore state an L2 failed to deliver. (An earlier revision of this runbook said there was no L1-side setter for this map; that was wrong, and it made a non-zero L1 state look unrecoverable.)
+> **Read per chain before starting:** `mapChainIdWithheldAmounts(chainId)` on the live Dispenser, and on each L2 target dispenser **both** `withheldAmount` **and the OLAS balance**. Two cases:
```
dagacha, line 51: Phase 0 step 1 still encodes the discarded precondition.

> confirm all `0` (§4 precondition). Record `withheldAmount` on each **L2** target dispenser

§4/§9 now say non-zero L1 withheld is in scope and that the operator must record L1 map **and** L2 `withheldAmount` **and** the OLAS balance. This numbered step still stops at all-zero and never records the balance that step 14 restores. Operators follow numbered steps; the §4 callout does not save this.

docs/dispenser_migration_runbook_public.md
dagacha, line 121: This lands too early.

`DefaultDepositProcessorL1._receiveMessage` requires `l2Dispenser == l2TargetDispenser`. That binding is step 15 (and it zeros L1 processor `owner`). A `syncWithheldAmount` after step 14, before 15, reverts `WrongMessageSender` and never credits the new L1 map.

Step 22 is after 15, so the check belongs there. Drop the implication that restore at 14 is enough for the amount to reach L1.

dagacha (commented):
Review done with omp (xai-oauth/grok-4.6, effort Medium).

Mechanism of step 14 is correct. Two blocking leftovers that will still send an operator down the old path; not repeating the step 13 / `migrate()` NatSpec / `syncWithheldTokens` / 4663 event-shape points already on the thread.

